PR #1, opened 2018-09-18
add option to disable adding routes

mafintosh:


routes.js
```diff
@@ -13,6 +13,8 @@ const responseSchema = {
 }
 
 module.exports = async function (app, opts) {
+  if (opts.routes === false) return
```
mcollina: Is this working? the `routes: false`  option is not passed in https://github.com/mcollina/fastify-gql/blob/master/index.js#L49-L51.

In fact, it's probably better to add it there instead.

mafintosh: ah ok. was just quick and dirty hehehe

PR #22, opened 2018-12-23
fix(routes): graphiql redirection if a prefix have been set

PacoDu:
The redirection is incorrect if the plugin is registered with a prefix. 
This PR fixes the issue ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,7 @@ __fastify-gql__ supports the following options:
   executed before being jit'ed.
 * `routes`: boolean. Serves the Default: `true`. A graphql endpoint is
   exposed at `/graphql`.
-* `prefix`: String. Change the route prefix of graphql endpoint if enabled.
+* `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 
 ### HTTP endpoints
 
--- a/README.md
+++ b/README.md
@@ -103,6 +103,7 @@ __fastify-gql__ supports the following options:
   executed before being jit'ed.
 * `routes`: boolean. Serves the Default: `true`. A graphql endpoint is
   exposed at `/graphql`.
+* `prefix`: String. Change the route prefix of graphql endpoint if enabled.
 
 ### HTTP endpoints
 
--- a/test/routes.js
+++ b/test/routes.js
@@ -572,17 +572,33 @@ test('GET graphiql endpoint', async (t) => {
 
 test('GET graphiql endpoint with prefix', async (t) => {
   const app = Fastify()
+  app.register(GQL, {
+    graphiql: true,
+    prefix: '/test-prefix'
+  })
+
+  const res = await app.inject({
+    method: 'GET',
+    url: '/test-prefix/graphiql'
+  })
+
+  t.strictEqual(res.statusCode, 302)
+  t.strictEqual(res.headers.location, '/test-prefix/graphiql.html')
+})
+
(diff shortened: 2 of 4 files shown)


mcollina:
Can you add a unit test?

PacoDu:
```javascript
app.register(GQL, {
    prefix: '/test'
  })
```

This example is not managed, maybe we should handle it ?

mcollina:
> This example does not work, maybe we should handle it ?

What do you mean?

PacoDu:
```javascript
app.register(async function (app, opts) {
    app.register(GQL, {})
  }, { prefix: '/test' })
```

In order to add a prefix to fastify-gql you have to wrap the module like the example above. But maybe this is an intended behavior.

mcollina:
It depends if a plugin is encapsulated or not. Likely `fastify-swagger`  should support `prefix`, maybe you can open a bug there.

> Edit: We could handle prefix in GQL options by adding prefix: opts.prefix here

Definitely! Would you like to amend this PR or send a new one?

PacoDu:
It's WIP, I'm adding tests for prefixed `/graphql`

PacoDu:
We should be good now, prefix is handled both from a wrapper and from opts, tested and documented ;).

README.md
```diff
@@ -103,6 +103,7 @@ __fastify-gql__ supports the following options:
   executed before being jit'ed.
 * `routes`: boolean. Serves the Default: `true`. A graphql endpoint is
   exposed at `/graphql`.
+* `prefix`: String. Change the route prefix of graphql endpoint if enabled.
```
mcollina: of the graphql endpoint

test/routes.js
```diff
@@ -555,3 +555,78 @@ test('error if there are functions defined in the root object', async (t) => {
...
+
+test('GET graphql endpoint with prefix', async (t) => {
+  const app = Fastify()
+  app.register(GQL, {
+    prefix: '/test-prefix'
+  })
+
+  const res = await app.inject({
+    method: 'GET',
+    url: '/test-prefix/graphql'
```
mcollina: Can you please add a proper GraphQL query here?

test/routes.js
```diff
@@ -555,3 +555,78 @@ test('error if there are functions defined in the root object', async (t) => {
...
+  app.register(async function (app, opts) {
+    app.register(GQL, {})
+  }, { prefix: '/test-wrapper-prefix' })
+
+  const res = await app.inject({
+    method: 'GET',
+    url: '/test-wrapper-prefix/graphql'
+  })
+
+  t.strictEqual(res.statusCode, 400)
```
mcollina: Can you please add a proper GraphQL query?

test/routes.js
PacoDu: Ok I will change that.

But the test shouldn't be focus on only one feature ? Adding a query could make the test fail for another reason than its original purpose.  When I wrote this test, I was wondering if I should make a graphql query here, but I ended up only using a simple GET as a 400 response is enough to ensure that the endpoint was reached.

mcollina: You are right, as it’s better to test one thing only. I consider a 400 an error path, not a success path, and we should really validate the happy path in this situation.

PacoDu: Oh well I see your point, it's the 4xx as it's an error this could be returned from a hook or something like that and it doesn't ensure the validation of the desired path. 
Ok, so I will change the test with the add query !

PR #24, opened 2019-01-06
Plugin name definition

PacoDu:
There is no plugin definition which makes dependencies resolution not working if a plugin depends on `fastify-gql`

### DIFF
--- a/index.js
+++ b/index.js
@@ -204,4 +204,6 @@ module.exports = fp(async function (app, opts) {
 
     return execution
   }
+}, {
+  name: 'fastify-gql'
 })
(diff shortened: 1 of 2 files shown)


PacoDu:
I've added a test for the plugin definition, but I don't understand why it's taking ~700 ms to complete.

mcollina:
does it show up if you execute the file directly?

PacoDu:
Yes :
```
$ tap test/plugin-definition.js
test/plugin-definition.js ............................. 1/1 620ms
total ................................................. 1/1

  1 passing (641.724ms)

  ok
```

Is this due to `avvio` dependency resolution ?

mcollina:
Weird, it’s not related to fastify at all then.

Issue #25, opened 2019-01-06
type Query and Mutation are inconsistent


PacoDu:
When bootstrapping `fastify-gql` without a schema in order to add schemas with `extendSchema`, the declaration of type Query makes the usage of GQL inconsistent as you have to extend Query but not Mutation. I understand that declaring `type Mutation` in fastify-gql would force the user to declare a mutation to make a valid schema. But why chose query other mutation ? 

Either both shouldn't be declared, or maybe add an option that enable the definitions ?

By the way the example of `extendSchema` in the README is wrong the schema should be `extend Query` as query is already declared.

https://github.com/mcollina/fastify-gql/blob/d5d79d4a62a03a9ac5d742fe90354b1c6070aef6/index.js#L52-L59

Yesterday I've started a new plugin that adds the mutation, query, and date Scalar if needed [fastify-gql-base](https://github.com/PacoDu/fastify-gql-base). But I think that query and mutation should be handled in this plugin.

PacoDu:
Ok, after trying to implement a PR to make a proposition about this, I discovered that Query is required by GraphQL that's why you chose to define Query. So maybe just add a definieMutation option ? 

This would allow consistent usage of extend in plugins that add extend the definitions.

PR #26, opened 2019-01-06
Add defineMutation option

PacoDu:
This is a proposition to solve #25

### DIFF
--- a/index.js
+++ b/index.js
@@ -54,7 +54,11 @@ module.exports = fp(async function (app, opts) {
       query: new GraphQLObjectType({
         name: 'Query',
         fields: { }
-      })
+      }),
+      mutation: opts.defineMutation ? new GraphQLObjectType({
+        name: 'Mutation',
+        fields: { }
+      }) : undefined
     })
   }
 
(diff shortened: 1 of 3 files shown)


PacoDu:
Do not merge ⚠

It worst than I thought #25, as adding a `type Mutation` will not be considered like a mutation. Sending a mutation will return `Schema is not configured for mutations`

PacoDu:
This is fixed now and tested. This allows the usage of mutation which was not possible before if you register `fastify-gql`without a schema.

index.js
```diff
@@ -54,7 +54,11 @@ module.exports = fp(async function (app, opts) {
       query: new GraphQLObjectType({
         name: 'Query',
         fields: { }
-      })
+      }),
+      mutation: opts.defineMutation ? new GraphQLObjectType({
+        name: 'Mutation',
+        fields: { }
+      }) : undefined
```
mcollina: Why are you doing this only behind an option? Probably we should just do this all the time?

PacoDu: I chose the option way because otherwise it would force the user to define at least one mutation, which may not be required for every use case.

PR #27, opened 2019-01-14
When `jit` option equals 1, JIT is now properly computed.

bhamon:


bhamon:
I don't see a way to detect if JIT is used or not, I'm not sure on how to test that.
This PR corrects the `routes/JIT` unit test that originally silently failed.

PR #28, opened 2019-01-15
Handle errors according to graphql spec and allow custom error handling

PacoDu:
This PR adds the errorHandler option to fastify-gql in order to allow custom error handling.
With test and documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ __fastify-gql__ supports the following options:
   exposed at `/graphql`.
 * `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 * `defineMutation`: Boolean. Add the empty Mutation definition if schema is not defined (Default: `false`).
-* `errorHandler`: Function. Change the default error handler (Default: `false`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
+* `errorHandler`: Function. Change the default error handler (Default: `true`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
 
 ### HTTP endpoints
 
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ __fastify-gql__ supports the following options:
   exposed at `/graphql`.
 * `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 * `defineMutation`: Boolean. Add the empty Mutation definition if schema is not defined (Default: `false`).
-* `errorHandler`: Function. Change the default error handler.
+* `errorHandler`: Function. Change the default error handler (Default: `false`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
 
 ### HTTP endpoints
 
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ __fastify-gql__ supports the following options:
   exposed at `/graphql`.
 * `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 * `defineMutation`: Boolean. Add the empty Mutation definition if schema is not defined (Default: `false`).
-* `errorHandler`: Function. Change the default error handler (Default: `true`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
+* `errorHandler`: `Function`  or `boolean`. Change the default error handler (Default: `true`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
 
 ### HTTP endpoints
 
--- a/README.md
+++ b/README.md
@@ -105,6 +105,7 @@ __fastify-gql__ supports the following options:
   exposed at `/graphql`.
 * `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 * `defineMutation`: Boolean. Add the empty Mutation definition if schema is not defined (Default: `false`).
+* `errorHandler`: Function. Change the default error handler.
 
 ### HTTP endpoints
 
--- a/routes.js
+++ b/routes.js
@@ -67,7 +67,11 @@ function validationHandler (validationError) {
 }
 
 module.exports = async function (app, opts) {
-  app.setErrorHandler(opts.errorHandler || defaultErrorHandler)
+  if (typeof opts.errorHandler === 'function') {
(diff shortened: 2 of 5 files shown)


mcollina:
What is the problem with fastify-gql error handler?
If the fastify-gql error handler is broken, then we should likely fix it.
I had problems with it in the past, so it might very well be buggy.

express-graphql and similar do not offer  this functionality as far as I know. The output of a graphql endpoint in case of errors should be canonically defined.

PacoDu:
Maybe I didn't understand the correct way to handle errors in fastify but this is my use case:

I have resolvers that delegate the query to my business layer, which will : 
* check authorization
* validate user inputs
* etc.

Each of these steps can throw custom errors that I want to manage (ValidationError, AuthorizationError, ServiceError ...), I would like to be able to define the error handler to catch them and respond accordingly to avoid sensible information leaks. But because of the encapsulated error handler of fastify I've no way to catch them.

The idea is to throw an delegate error handling to an higher level of the application. That was the way I used to work with Koa. Because of the way graphql is implemented it's not possible to delegate this to fastify hooks (I use a hook for token parsing but checking authorization depends on the method called by the graphql query, same apply to inputs validation)

mcollina:
I think the format of the response of a GraphQL endpoint is standardized as: https://graphql.org/learn/serving-over-http/#response. I don't think that should be changed, and maybe the output would need to be adapted.

However, I see the point of this PR, let's do this.

README.md
```diff
@@ -105,6 +105,7 @@ __fastify-gql__ supports the following options:
   exposed at `/graphql`.
 * `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 * `defineMutation`: Boolean. Add the empty Mutation definition if schema is not defined (Default: `false`).
+* `errorHandler`: Function. Change the default error handler.
```
mcollina: I think this  should also be `false`, meaning that no errorHandler would be installed.

PacoDu: And maybe add details about the standardized response that the user should implement if a customized handler is defined ?

mcollina: Yes please.

mcollina:
> Based on the standardized response that graphql should return, we have to send the result of the second query, and errors related to the first query.
>
> For now the default error handler in fastify-gql will throw and return errors but data will be dropped. This standardized response makes the implementation other HTTP quite inconsistent, should we return statusCode 200 or 400 in this case ? Maybe 206 (Partial Content) ?

What is express-graphql returning? We might just return a 200 I presume, or a 206.

mcollina:
Note that for graphql error we should also call https://github.com/graphql/graphql-js/blob/master/src/error/formatError.js before sending them out.

PacoDu:
I tried to handle most of the errors according to Graphql spec.

I did not handle the partial result part, as I think it would require to attach data to the error which I think is quite strange ... I would like to have your point of view for this one.

test/routes.js
```diff
@@ -655,3 +655,67 @@ test('GET graphql endpoint with prefixed wrapper', async (t) => {
...
+    app.log.error(error)
+    reply.code(403)
+    reply.send()
+  }
+
+  app.register(GQL, {
+    schema,
+    resolvers,
+    errorHandler
+  })
```
mcollina: Can you add a test that passes `errorHandler: false` and sets a root error handler in fastify?

README.md
```diff
@@ -105,6 +105,7 @@ __fastify-gql__ supports the following options:
   exposed at `/graphql`.
 * `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 * `defineMutation`: Boolean. Add the empty Mutation definition if schema is not defined (Default: `false`).
+* `errorHandler`: Function. Change the default error handler (Default: `false`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
```
mcollina: I think should be `default: true`, as there is a default error handler set.

PacoDu: I think "`errorHandler`: Function. Change the default error handler (Default: `true`)" is quite ambiguous. Maybe rename the option to `customErrorHandler` and keep default `false` ?

test/routes.js
PacoDu: I think we do not see this option the same way, if I set `errorHandler` to false, based on the code https://github.com/mcollina/fastify-gql/pull/28/files#diff-d640e5835a6b0c342cf705a06018c2b9R70 it will enable the default error handler of `fastify-gql`.

I can change this if you prefer to have an option that enables/disable the fastify-gql errorHandler and having the user define it in his scope. The way I've implemented it here I was passing the handler function to fastify-gql.

README.md
PacoDu: Based on you other comment I think that, for you, this option is not a function, but only a boolean that enable or disable fastify-gql handler.

test/routes.js
PacoDu: Commit is ready to match the behavior you asked (errorHandler is only a Boolean that disable or enable the fastify-gql error handler). Let me know if it's really what you requested and I'll push it.

routes.js
```diff
@@ -11,27 +12,62 @@ const responseSchema = {
...
+function validationHandler (validationError) {
+  if (validationError) {
+    const err = new BadRequest()
+    err.errors = [validationError]
+    throw err
+  }
+}
+
+module.exports = async function (app, opts) {
+  app.setErrorHandler(opts.errorHandler || defaultErrorHandler)
```
mcollina:
```suggestion
  if (typeof opts.errorHandler === 'function') {
    app.setErrorHandler(opts.errorHandler)
  } else if (opts.errorHandler === true || opts.errorHandler === undefined) {
      app.setErrorHandler(defaultErrorHandler)
  }
```

This is the behavior I'm expecting. A test passing `errorHandler: false`  and `errorHandler: true` must be added.

README.md
mcollina: The default is `true`  at this point, as there is an error handler.

README.md
```diff
@@ -105,6 +105,7 @@ __fastify-gql__ supports the following options:
   exposed at `/graphql`.
 * `prefix`: String. Change the route prefix of the graphql endpoint if enabled.
 * `defineMutation`: Boolean. Add the empty Mutation definition if schema is not defined (Default: `false`).
+* `errorHandler`: Function. Change the default error handler (Default: `true`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
```
mcollina:
```suggestion
* `errorHandler`: `Function`  or `boolean`. Change the default error handler (Default: `true`). _Note: If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response)._
```

mcollina:
Good work! I'll release this now

PR #30, opened 2019-02-08
feat: add union and interface support

PacoDu:
This PR adds support for union and interface, and related tests !

### DIFF
--- a/index.js
+++ b/index.js
@@ -110,9 +110,9 @@ module.exports = fp(async function (app, opts) {
       } else if (type instanceof GraphQLObjectType) {
         const fields = type.getFields()
         const resolver = resolvers[name]
-        if (resolver.__isTypeOf) {
-          type.isTypeOf = resolver.__isTypeOf
-          delete resolver.__isTypeOf
+        if (resolver.isTypeOf) {
+          type.isTypeOf = resolver.isTypeOf
+          delete resolver.isTypeOf
         }
         for (const prop of Object.keys(resolver)) {
           fields[prop].resolve = resolver[prop]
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -108,6 +110,10 @@ module.exports = fp(async function (app, opts) {
       } else if (type instanceof GraphQLObjectType) {
         const fields = type.getFields()
         const resolver = resolvers[name]
+        if (resolver.__isTypeOf) {
+          type.isTypeOf = resolver.__isTypeOf
+          delete resolver.__isTypeOf
+        }
```
mcollina: What are these lines for? What is `__isTypeOf`?

test/app-decorator.js
```diff
@@ -464,3 +464,282 @@ test('enum should be supported', async (t) => {
...
+      },
+      getGeometryMultiPolygon: async () => {
+        return {
+          type: 'MultiPolygon',
+          coordinates: 1
+        }
+      }
+    },
+    Geometry: {
+      __resolveType (geometry) {
```
mcollina: What is `__resolveType`?

index.js
```diff
@@ -116,6 +122,9 @@ module.exports = fp(async function (app, opts) {
         for (const prop of Object.keys(resolver)) {
           type[prop] = resolver[prop]
         }
+      } else if (type instanceof GraphQLInterfaceType || type instanceof GraphQLUnionType) {
+        const resolver = resolvers[name]
+        type.resolveType = resolver.__resolveType
```
mcollina: Is this common syntax in the graphql world?

PacoDu: You are right this is not part of the official `graphql` lib, I confused with Apollo's makeExecutableSchema which use `__` syntax. 

I was looking for a way to add union and interface without using `graphql-tool` or raw schema definition. But I've mixed both libraries ... 👎 

I will remove the `__` to match the official `graphql` library, does it suits you ? 
Adding `isTypeOf` and `resolveType` will allow users to define resolvers for union and interface defined in the schema definition.

mcollina: Awesome!

PR #49, opened 2019-05-06
feat(types): initial typings support

### LINKED ISSUE #48: Add typescript typings
In order to use this package with typescript it requires proper typings.

I've made some progress, however I've find out that for now it's not possible. That's because now in fastify typings plugin `RegisterOptions` extend `RouteShorthandOptions` for some reason not known to me. `RouteShorthandOptions` however contains `schema` field that has type `RouteSchema` that's obviously incompatible with the same field `schema` in this module which could be either string or GraphQLSchema.

Typescript allows to redefine `RouteShorthandOptions` with additional types for `schema`, but it will affect all routes. So I think we should wait for https://github.com/fastify/fastify/pull/1569 to be merged. Or if you have any other suggestions, you're welcome.

SkeLLLa:
Initial types support. Currently it's not compatible with Fastify types. Draft for now until https://github.com/fastify/fastify/pull/1569 merged.

Closes #48.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -62,3 +62,4 @@ typings/
 
 # clinic
 *clinic*
+package-lock.json
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,123 @@
+/// <reference types="node" />
+
+import fastify, {
+  FastifyError,
+  FastifyReply,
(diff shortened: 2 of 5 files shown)


mcollina:
We have shipped types.

PR #53, opened 2019-05-15
Feat/add graphql playground

### LINKED ISSUE #52: Add graphql playground alongside with graphiql
Hi. There's graphql playground that does pretty much the same things that graphiql does. However it has more features. Here's demo page: https://www.graphqlbin.com/v2/6RQ6TM

It has standalone package https://www.npmjs.com/package/graphql-playground-html that contains several html files https://github.com/prisma/graphql-playground/tree/master/packages/graphql-playground-html that could be used just like as graphiql.html. We could copy one of them into static folder just like it's done in fastify-swagger plugin.

Also it doesn't use service worker which allows to explore it without having https (e.g. on some domain like example.loc that will point to 127.0.0.1). In such case graphiql doesn't work.

I could send PR with this changes, but I need a suggestion which route it should expose: `/gql-playground` or `graphiql-playground` or something else.

SkeLLLa:
Added `.npmignore` to remove unnecessary files from release.
Added `package-lock.json` to `.gitignore`.
Added GraphQL IDE aka GraphQL Playground. 

Closes #52.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,4 @@
+tap-snapshots/
+examples/
+test/
+bench.sh
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,0 @@
-tap-snapshots/
-examples/
-test/
-bench.sh
--- a/README.md
+++ b/README.md
@@ -100,9 +100,9 @@ __fastify-gql__ supports the following options:
 * `resolvers`: Object. The graphql resolvers.
 * `loaders`: Object. See [defineLoaders](#defineLoaders) for more
   details.
-* `graphiql`: boolean. Serve
-  [GraphiQL](https://www.npmjs.com/package/graphiql) on `/graphiql` and 
-  [GraphQL IDE](https://www.npmjs.com/package/graphql-playground-react) on `/playground`
+* `graphiql`: boolean | string. Serve
+  [GraphiQL](https://www.npmjs.com/package/graphiql) on `/graphiql` if `true` or `'graphiql'`, or 
+  [GraphQL IDE](https://www.npmjs.com/package/graphql-playground-react) on `/playground` if `'playground'`
   if `routes` is `true`.
 * `jit`: Intenger. The minimum number of execution a query needs to be
   executed before being jit'ed.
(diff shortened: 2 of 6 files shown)


.npmignore
```diff
@@ -0,0 +1,4 @@
+tap-snapshots/
+examples/
+test/
+bench.sh
```
mcollina: Can you remove this file? Thanks

SkeLLLa: ok

README.md
```diff
@@ -101,8 +101,9 @@ __fastify-gql__ supports the following options:
 * `loaders`: Object. See [defineLoaders](#defineLoaders) for more
   details.
 * `graphiql`: boolean. Serve
-  [GraphiQL](https://www.npmjs.com/package/graphiql) on `/graphiql` if
-  `routes` is `true`.
+  [GraphiQL](https://www.npmjs.com/package/graphiql) on `/graphiql` and 
+  [GraphQL IDE](https://www.npmjs.com/package/graphql-playground-react) on `/playground`
+  if `routes` is `true`.
```
mcollina: I think we should better expose `playground` with a `playground` option. WDYT?

SkeLLLa: This alongside as graphiql doing pretty much the same thing - providing an simple UI for current graphql server. So I don't see cases where playground could be enabled and graphiql - disabled and vice versa. So the main goal for server developers is that this flag provides UI. And other, for example, FE devs just will choose whatever they like more.

Also if I wrote such plugin just for myself, then I'd have slightly different structure with option called `ui` that will be a prefix for both solutions, e.g. 
```
{
 ui: '/graphql-ui
}
```
And then we have two options that are enabled with such prefix: `/graphql-ui/graphiql` and `graphql-ui/playground`. But `/graphiql` route is kind of de-facto standard for graphql servers and many devs used to it, so this approach would be a bit confusing for people.

One more option - is to have kind of switch between ui implementations: 
```
{
  graphiql: 'graphiql' // one of [undefined, 'graphiql', 'playground']
}
```

mcollina: I think we should support the option as you said, following this pattern:

* if true or ‘graphiql’, enable graphiql
* if ‘playground’, enable the playground

Does that sound reasonable?

SkeLLLa: Yes, sounds good.

SkeLLLa: Done. If you prefer to serve graphql playground on `/graphiql` route when `graphiql: 'playground'` is set, let me know and I'll update the code.

Issue #56, opened 2019-05-21
Making fastify-gql even faster


wtrocki:
# Introduction

Kudos for creating an amazing library.
I have played with it and noticed that there is another problem when dealing with GraphQL on the resolver side. GraphQL-js by default has this feature to execute query and pick what user wanted from the whole object returned from root resolver (and relationship resolvers as well)

https://github.com/graphql/graphql-js/blob/master/src/execution/execute.js#L457-L463

What it does is simply pick elements from the graph based on what the user requested in the client. In some cases, graphql queries can be misused on client. 
For example, someone can query half of the database with one resolver but just return a single key to client entire type. Querying username from `getUserProfile()` query. The problem can be seen for API's that are shared with 3rd parties etc.

I think that developers can improve performance dramatically if:

- They will use only root resolvers (yes - No nested/relationship resolvers)
- They will use `info` object with the client query to see what client realy wants (instead of blindly return everything)
- All nested relationships can be executed using SQL joins etc. to reduce the number of calls to the database.
- The final object can be shaped on the root resolver.

For simple `getProfile` example on the main page this can help backend to process around 80% more requests.

That approach will be an alternative to DataLoader which I think just adds to the problem for resolvers using direct database connections. 

See blog post: https://medium.freecodecamp.org/a-5-line-major-efficiency-hack-for-your-graphql-api-type-resolvers-b58438b62864

# Problem with info

Info object is not actually so nice to work with for someone who start with GraphQL. Adding helper to fastifyQL will radically improve the performance of average applications that connect to the database or REST endpoints that support the `filter` argument.

https://gist.githubusercontent.com/the-vampiire/3d4875ef51c18c20fa3a53fd9307aa63/raw/c1679be181be09683ab5db4e2afa623a92963204/mapAttributes.js

I have experimented with this approach in https://github.com/graphql-heroes/graphql-backend-gen and it really works.

mcollina:
This is an amazing technique! How would you integrate this in this repo?

wtrocki:
What will make sense is to have some flag in config same as `loaders` flag:

`enableRootOnlyQueries` - if true will provide helper method that 

Provide helper method mounted to context:

```javascript
context.buildRootQueryArguments()
```

This needs to be generic. I have done this for specific libraries before like knex so I think the key here will be to define an interface that will help people to query their own data. This method can return something like:

```json
['fieldName1', 
'fieldName2', 
'fieldName3', 
'relation':[
    'fieldName4'
]
[...]

PR #59, opened 2019-05-24
Add ability to filter data based on user query

wtrocki:
Implementation #56 
 
## Testing

1. Run `node ./examples/queryFilters.js`   
2. Use query in playground                                                  
```

 fragment Doggs on Dog{
   name   
   mother{
       name 
   }
 }

query { 
  dogs(limit:4) {
     ...Doggs
   }
}
```
3. We should see the following output
```
SQL ROOT QUERY: select "name","type" from Dog
SQL RELATION QUERY: select "name" from DoggyParents
```

# Tasks

- [x] Basic implementation
- [x] Experiment with format (relationships)
- [x] Unit tests
- [x] Documentation


PacoDu:
Nice feature, for now I'm using a similar method to filter requested fields to avoid over fetching inspired by:
https://github.com/bhamon/pluton-graphql/blob/4f062b2c1a8d1ee6bfaea5023588884fb829bfd4/lib/introspectFields.js#L64-L91

Maybe this can give you some idea for the implementation.

wtrocki:
@PacoDu Thank you so much. This is really useful. I have a really basic version done, but now experimenting with the relationships which are a little bit more complex to explain for the end user. By default, GraphQL will take care of the shape of the data for nested resolvers. If we do not want to use nested resolvers then probably we need an extra helper.

An additional idea I have is to provide a custom execute method that will calculate those fields beforehand and then pass them as part of the context so then users will get everything outside the box with a little bit customized graphql workflow. 

I will try to post things as soon as possible but want to make sure that I have unit tests and good documentation for features beforehand.

wtrocki:
Small update. I managed to get this implemented for simple cases but it still failing when fragments are used. Hoping to push less WIP version once that is sorted with unit tests and example app.

wtrocki:
Added more info to tickets. Only unit tests left to do.

queryFields.js
```diff
@@ -0,0 +1,139 @@
...
+      wrapper = '"'
+    }
+    if (!separator) {
+      separator = ','
+    }
+    return fields.fields.map((field) => {
+      return `${wrapper}${field}${wrapper}`
+    }).join(separator)
+  }
+  /**
```
wtrocki: Standard.js do not like it. 
I wish I could use typescript here.

wtrocki:
@mcollina I have this battle tested with:

- MongoDB/Couch
- SQL based queries against Postgres and MySQL

I did not provide any helper for structure nested data after as this can be done manually by developers.

Unit tests can be extended to include Fragment cases etc. but I did not want to overload repository with testing files. Generally, this should be ready for review.

What we missing is probably a short documentation chapter about this technique. 
Is it ok to add that?

wtrocki:
Documentation added.

queryFields.js
```diff
@@ -0,0 +1,139 @@
+// @ts-check
```
mcollina: What is this?

queryFields.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Imports:
+ * @typedef {import('graphql').FieldNode} FieldNode
+ * @typedef {import('graphql').GraphQLResolveInfo} GraphQLResolveInfo
+ * @typedef {import('graphql').SelectionSetNode} SelectionSetNode
+ * @typedef {Object<string, any>} QueryObject
+ */
```
mcollina: What is this?

mcollina:
I would like to see an example that is sending this data to a SQL database, as this is mentioned in some comment. We might add a devDependency to sqlite for example.

queryFields.js
wtrocki: I do write a lot of typescript and I'm used to work with the preprocessor for types. The line enables typescript preprocessor so I can add types and type checks. This is really useful for rapid dev on GraphQL types especially as they are very complex.

More info: https://code.visualstudio.com/docs/languages/javascript#_type-checking

In this case, types were used only for development time and they will be not exposed to end users - methods are mounted to the context. I was planning to remove this after review.

queryFields.js
wtrocki: Imports for type definitions for javascript. More info in the comment above.

wtrocki:
## Update:
1. [ ] Going to add a full suite of unit tests
2. [ ] Going to connect with SQLite
3. [ ] Going to remove TypeScript related noise.

wtrocki:
Still working on unit tests

wtrocki:
Finalized first pass of review comments.

examples/query-filters/queryFilters.js
```diff
@@ -0,0 +1,62 @@
...
+      const queryData = context.buildQueryObject(info)
+      const db = await dbPromise
+      let posts
+      let relation = 'category'
+
+      if (queryData.hasRelation(relation)) {
+        // Select with relation
+        posts = await db.all(
+          `SELECT  ${queryData.getRootFields()}, ${queryData.getRelationFields(relation)} 
+          FROM Post INNER JOIN Category ON Category.id = Post.categoryId`)
```
mcollina: Would you mind adding a comment that this can have a SQL injection?

wtrocki: Sorry, that took me long to get back on this. I was actually thinking about how to do that properly to avoid a couple of problems. Using low-level SQL will not only mean SQL injection but also compatibility issues ( for example select table.field __on__ table__field will require __AS__ in PostgresDB). I think to do this properly we can:

- [ ] Ignore those helpers that were written and made them part of example which will mean that we deliver fields only and users will need to apply their own pattern depending on their ORM. It makes sense as there will be so many dialects to support.

- [ ] Use more opinionated and high-level query mechanism in helpers:
nearform/sql or even knex. This will mean that things will work out of the box for some of the common databases that are supported by this libs. This will mean that helpers will be more solid and production ready but the general pattern will be not that simple to understand and map to non supported libraries.

- [ ] Provide helpers like now but mark them as just examples and provide variations for different dialects. I spent some time cracking GraphQL specification to see if it is possible to pass field value as invalid js type name and it does not seem to be passing validation.

mcollina:
What do you mean by “mark them as just examples”?

Would you mind adding a full test, including a complete Graphql endpoint? Some cut-down version of the example would be ok as well.

wtrocki:
Sorry for not being clear. By marking them I meant to make a comment that helpers provide SQL syntax and may not work for NoSQL db's etc

wtrocki:
This PR got outdated so I'm going to close it. I have extracted this code into a separate package that is being used in a couple of projects along with fastify-gql

https://github.com/aerogear/graphql-query-mapper

Thank you so much for collaboration on this PR.

PR #70, opened 2019-08-23
Spike/depth limit query

john-behan:


index.js
```diff
@@ -221,6 +222,16 @@ module.exports = fp(async function (app, opts) {
         throw err
       }
 
+      if (opts.queryDepths) {
```
mcollina: can you please put queryDepths in a variable at the top of the file?

PR #71, opened 2019-08-27
Added custom GraphQL context resolver for GraphQL routes.

### LINKED ISSUE #47: Enrich context
Is it possible to enrich the context thats being sent to the resolvers?

Perhaps through a function that you supply when creating the instance, something like this:

```
app.register(GQL, {
  schema,
  resolvers,
  context: (request, reply) => ({ foo: true }),
});
```

brettlaforge:
Added custom GraphQL context resolver for GraphQL routes.
Removed app and reply from the GraphQL context in decorators.

Closes #47

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,22 +88,40 @@ app.get('/', async function (req, reply) {
 app.listen(3000)
 ```
 
-### Access app context in resolver
+### Provide custom context
 
 ```js
-...
+'use strict'
+
+const Fastify = require('fastify')
+const GQL = require('fastify-gql')
+
+const app = Fastify()
+
+const schema = `
+  type Query {
+    add(x: Int, y: Int): Int
+  }
+`
 
 const resolvers = {
   Query: {
-    add: async (_, { x, y }, context) => {
-      // do you need the request object?
-      console.log(context.reply.request)
-      return x + y
+    add: async (_, { x, y }, ctx) => {
+        console.log(ctx.start)
+        return x + y
     }
   }
 }
 
-...
+app.register(GQL, {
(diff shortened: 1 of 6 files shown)


test/app-decorator.js
```diff
@@ -41,39 +41,6 @@ test('basic GQL', async (t) => {
...
-
-  const query = '{ ctx }'
-  const res = await app.graphql(query, { num: 42 })
-
-  t.deepEqual(res, {
-    data: {
-      ctx: 42
-    }
-  })
-})
```
mcollina: I would like to retain this as the default behavior.

README.md
```diff
@@ -88,22 +88,40 @@ app.get('/', async function (req, reply) {
...
   }
 }
 
-...
+app.register(GQL, {
+  schema,
+  resolvers,
+  context: async (request, reply) => {
+    return { start: Date.now() }
+  }
```
mcollina: Why this should be async? Would a normal function be enough?

test/app-decorator.js
mcollina: I would prefer to retain the default behavior.

test/loaders.js
```diff
@@ -286,7 +295,10 @@ test('support context in loader', async (t) => {
   app.register(GQL, {
     schema,
     resolvers,
-    loaders
+    loaders,
+    context: (request, reply) => {
+      return { reply }
+    }
```
mcollina: Are the changes in this file needed?

mcollina:
@brettlaforge any updates on this?

PR #74, opened 2019-09-05
Added custom GraphQL context resolver for GraphQL routes.

brettlaforge:
I've reduced the scope of this PR to only effect the routes so that the addition of the reply and app to the GraphQL context can be retained as the default behaviour.

Closes #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,33 @@ const resolvers = {
 ...
 ```
 
+### Build a custom GraphQL context object
+
+```js
+...
+
+const resolvers = {
+  Query: {
+    me: async (obj, args, ctx) => {
+      // access user_id in ctx
+      console.log(ctx.user_id)
+    }
+  }
+}
+
+app.register(GQL, {
+  schema: makeExecutableSchema({ typeDefs, resolvers }),
+  context: (request, reply) => {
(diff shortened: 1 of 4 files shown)


routes.js
```diff
@@ -111,7 +111,12 @@ module.exports = async function (app, opts) {
       }
     }
 
-    return reply.graphql(query, null, variables, operationName)
+    let context
+    if (opts.context) {
+      context = await opts.context(request, reply)
+    }
```
mcollina: Would you mind extracting `opts.context` into a `const` at the top of the plugin?

README.md
```diff
@@ -126,6 +153,7 @@ __fastify-gql__ supports the following options:
   executed before being jit'ed.
 * `routes`: boolean. Serves the Default: `true`. A graphql endpoint is
   exposed at `/graphql`.
+* `context`: `Function`. Result of function is passed to resolvers as a custom GraphQL context.
```
mcollina: would you mind specifying that the parameters will be `request` and `reply`? Also, we should note that this is called only with the pre-defined routes (`routes: true`).

mcollina:
@brettlaforge any updates?

PR #76, opened 2019-10-15
feat: add typings

frikille:
This PR adds Typescript typings.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,11 +6,12 @@ import fastify, {
 } from "fastify"
 import { IResolvers } from "graphql-tools";
 import { Server, IncomingMessage, ServerResponse } from "http";
+import { Http2Server, Http2ServerRequest, Http2ServerResponse } from 'http2';
 import graphql, { GraphQLSchema, GraphQLError, Source, DocumentNode, ExecutionResult } from 'graphql';
 
 declare namespace FastifyGQL {
 
-  export interface Plugin {
+  export interface Plugin<HttpResponse> {
     /**
      * Extend existing schema
      * @param schema graphql schema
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -17,13 +23,18 @@
   },
   "homepage": "https://github.com/mcollina/fastify-gql#readme",
   "devDependencies": {
+    "@types/graphql": "14.2.3",
```
mcollina:
```suggestion
    "@types/graphql": "^14.2.3",
```

index.d.ts
```diff
@@ -0,0 +1,121 @@
...
+     * Define data loaders
+     * @param loaders object with data loader functions
+     */
+    defineLoaders(loaders: {[key: string]: Function}): void;
+  }
+
+  export interface Options<
+    HttpServer,
+    HttpRequest,
+    HttpResponse
```
mcollina: In order to support http2, we might have to do: https://github.com/fastify/fastify-swagger/blob/2f7d7f68f8d87f16904b5da8bec38784b64ac5c8/index.d.ts#L70-L72.

test/types/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+const resolvers = {
+  Query: {
+    add: async (_: any, { x, y }: { x: number, y: number }) => x + y
+  }
+}
+
+app.register(GQL, {
+  schema,
+  resolvers
```
mcollina: would you mind including all the options?

test/types/index.ts
```diff
@@ -0,0 +1,39 @@
...
+
+app.register(async function (app) {
+  app.graphql.extendSchema(`extend type Query {
+    subtract(x: Int, y: Int): Int
+  }`)
+  app.graphql.defineResolvers({
+    Query: {
+      subtract: async (_: any, { x, y }: { x: number, y: number }) => x - y
+    }
+  })
```
mcollina: would you mind adding loaders as well?

package.json
frikille: This version must be fixed because this is the last version that contains type definitions. GraphQL v14.5.0 added Typescript definitions and with that addition the `@types/graphql` package became deprecated.

This dependency can be removed if the required `graphql` version is updated to `^14.5.0`

test/types/index.ts
```diff
@@ -0,0 +1,89 @@
...
+}
+
+app.register(GQL, {
+  schema: schema,
+  resolvers,
+  loaders: {},
+  graphiqls: true,
+  jit: 1,
+  routes: true,
+  prefix: 'prefix',
```
mcollina:
```suggestion
  prefix: '/prefix',
```

PR #77, opened 2019-10-22
Add custom GraphQL context option

frikille:
This PR is the same as stated in #74 and addresses the review comments.

Closes #74 
Closes #71

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,30 @@ const resolvers = {
 ...
 ```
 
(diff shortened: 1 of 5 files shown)


mcollina:
Would you mind adding a `co-authored-by` in the commit so that we keep attribution of @brettlaforge work in https://github.com/mcollina/fastify-gql/pull/74?

PR #78, opened 2019-10-22
Add subscription support

frikille:
This PR adds support of GraphQL subscriptions. It implements the `graphql-ws` websocket subprotocol as it is the most used protocol by client libraries.


subscriber.js
```diff
@@ -0,0 +1,74 @@
+const { Readable } = require('stream')
```
mcollina: Please use the readable-stream package.

subscription.js
```diff
@@ -0,0 +1,163 @@
...
+
+    const config = {
+      socket: conn.socket,
+      schema,
+      subscriber,
+      subscriptionContexts,
+      handleConnectionClose
+    }
+
+    conn.socket.on('message', handleMessage(config))
```
mcollina: The function that handleMessage will return is async, and it can throw. We need to add a catch in so that everything is all teared down properly if that happens.

subscription.js
```diff
@@ -0,0 +1,163 @@
...
+
+    async function handleConnectionClose () {
+      const a = Array.from(subscriptionContexts.values())
+      await Promise.all(a.map(subscriptionContext =>
+        subscriptionContext.close()))
+      conn.socket.close()
+    }
+
+    conn.socket.on('error', handleConnectionClose)
+    conn.socket.on('close', handleConnectionClose)
```
mcollina: Do not pass async functions to event emitter in general.

subscriber.js
```diff
@@ -0,0 +1,74 @@
...
+      }
+    }
+
+    this.emitter.on(event, listener)
+    this._listeners.set(event, listener)
+  }
+
+  /**
+   * @topic: string representing the subscription topic
+   */
```
mcollina: We do not use jsdoc.

routes.js
```diff
@@ -73,79 +73,81 @@ module.exports = async function (app, opts) {
...
+      } = request.body
 
-    return reply.graphql(query, null, variables, operationName)
-  })
+      return reply.graphql(query, null, variables, operationName)
+    }
+  }
+
+  if (opts.subscription) {
+    require('./subscription')(app, getOptions, {
```
mcollina: Can you move the require at the top?

subscription.js
```diff
@@ -0,0 +1,163 @@
...
+  fastify.register(Websocket, {
+    handle,
+    options: { maxPayload: 1048576 }
+  })
+
+  fastify.route({
+    ...getOptions,
+    wsHandler: createWSHandler(schema, subscriber)
+  })
+}
```
mcollina: This should be a fastify plugin, something that we call `.register(subscription)`.

routes.js
```diff
@@ -73,79 +73,83 @@ module.exports = async function (app, opts) {
     app.setErrorHandler(defaultErrorHandler)
   }
 
-  app.get('/graphql', {
+  app.post('/graphql', {
```
mcollina: why this was moved from get to post? It make this very hard to review

frikille: I thought it is better to have the `post` first as it is always the same. But you're right, it is really hard to review so changed the order back.

subscription.js
```diff
@@ -0,0 +1,181 @@
...
+
+      return
+    }
+
+    const subscriptionConnection = new SubscriptionConnection(socket, {
+      schema,
+      subscriber
+    })
+
+    connection.on('error', (e) => {})
```
mcollina: We should tear down the subscriptionConnection in case of an error, or are we doing it anyway?

subscription.js
```diff
@@ -0,0 +1,181 @@
...
+
+const GRAPHQL_WS = 'graphql-ws'
+const GQL_CONNECTION_INIT = 'connection_init' // Client -> Server
+const GQL_CONNECTION_ACK = 'connection_ack' // Server -> Client
+
+const GQL_CONNECTION_TERMINATE = 'connection_terminate' // Client -> Server
+const GQL_START = 'start' // Client -> Server
+const GQL_DATA = 'data' // Server -> Client
+const GQL_ERROR = 'error' // Server -> Client
+// const GQL_COMPLETE = 'complete' // Server -> Client
```
mcollina: Why is this commented out?

subscriber.js
```diff
@@ -0,0 +1,73 @@
...
+      },
+      iterator: queue
+    }
+  }
+
+  close () {
+    this._listeners.forEach((fn, event) => {
+      this.emitter.removeListener(event, fn)
+    })
+    this._listeners = new Map()
```
mcollina: If we are closing it, why are we creating a new map?

subscriber.js
```diff
@@ -64,8 +64,8 @@ module.exports = class Subscriber {
     }
   }
 
-  publish (event) {
-    this.emitter.emit(event)
+  publish (event, callback) {
+    this.emitter.emit(event, callback || (() => {}))
```
mcollina: I would rather return a promise here instead

subscription.js
```diff
@@ -45,7 +45,13 @@ class SubscriptionConnection {
     this.subscriber = subscriber
     this.subscriptionContexts = new Map()
 
-    this.socket.on('message', this.handleMessage.bind(this))
+    this.socket.on('message', (message) => {
+      try {
+        this.handleMessage(message)
+      } catch (e) {
+        this.handleConnectionClose()
+      }
```
mcollina: this should be a catch handler, because `handleMessage` returns a `Promise`. Can you add a test for this case? You might have to use https://www.npmjs.com/package/proxyquire to mock some things

subscription.js
```diff
@@ -102,17 +108,12 @@ class SubscriptionConnection {
...
-      })
-      .then(() => {
-        this.sendMessage(GQL_COMPLETE, data.id, null)
-      })
-      .catch(err => {
-        throw err
-      })
+
+    for await (const value of result) {
+      this.sendMessage(GQL_DATA, data.id, value)
```
mcollina: This does not implement any form of backpressure. Would you mind adding a TODO for this in the code? We might get back to it later in a follow-up PR.

subscriber.js
```diff
@@ -1,78 +1,62 @@
...
-    }
+    return this.queue
   }
 
   publish (event) {
     return new Promise((resolve, reject) => {
-      this.emitter.emit(event, resolve, reject)
+      this.pubsub.publish(event, (err) => {
+        if (err) {
+          return reject(err)
```
mcollina: It would be good to a test case for the error condition.

README.md
```diff
@@ -130,6 +130,70 @@ app.register(GQL, {
...
+        id,
+        message
+      }
+      notifications.push(notification)
+      pubsub.publish({
+        topic: 'NOTIFICATION_ADDED',
+        payload: {
+          notificationAdded: notification
+        }
+      })
```
mcollina: I would await this

README.md
```diff
@@ -130,6 +130,70 @@ app.register(GQL, {
...
+
+      return notification
+    }
+  },
+  Subscription: {
+    notificationAdded: {
+      subscribe: (root, args, { pubsub }) => pubsub.subscribe('NOTIFICATION_ADDED')
```
mcollina: Can this be awaited?

test/subscription.js
```diff
@@ -0,0 +1,185 @@
...
+          query: `
+            mutation {
+              addNotification(message: "Hello World") {
+                id
+              }
+            }
+          `
+        }
+      })
+    }, 1000)
```
mcollina: Can we rewrite the test to avoid the `setTimeout`? Using them makes the tests more brittle and prone to failures.

lib/subscription-connection.js
```diff
@@ -0,0 +1,132 @@
...
+    fastify
+  }) {
+    this.fastify = fastify
+    this.socket = socket
+    this.schema = schema
+    this.subscriber = subscriber
+    this.subscriptionContexts = new Map()
+
+    this.socket.on('message', (message) => {
+      this.handleMessage(message).catch(e => {
```
mcollina: We should likely log this error here.

PR #86, opened 2019-11-15
Increase test coverage

proti:


test/subscriber.js
```diff
@@ -55,3 +55,21 @@ test('subscription context publish event errs, error is catched', t => {
...
+  emitter.on = (topic, listener, done) => {
+    done(new Error('Dummy error'))
+  }
+
+  const pubsub = new PubSub(emitter)
+  const sc = new SubscriptionContext({ pubsub })
+
+  try {
+    await sc.subscribe('TOPIC')
+    sc.publish({ topic: 'TOPIC' }, () => {})
```
mcollina: publish is not needed

test/subscriber.js
```diff
@@ -55,3 +55,21 @@ test('subscription context publish event errs, error is catched', t => {
...
+  } catch (e) {
+    t.pass()
```
mcollina: please validate that the error is the same that is generated inside on()

lib/subscription.js
```diff
@@ -3,7 +3,7 @@ const Websocket = require('fastify-websocket')
 const SubscriptionConnection = require('./subscription-connection')
 const GRAPHQL_WS = 'graphql-ws'
 
-function handle (conn) {
+export function handle (conn) {
```
mcollina: This will not work outside of tap, you should use `module.export`. However I would prefer this to be tested by opening a websocket connection to a different route instead.

lib/subscription.js
```diff
@@ -3,7 +3,7 @@ const Websocket = require('fastify-websocket')
 const SubscriptionConnection = require('./subscription-connection')
 const GRAPHQL_WS = 'graphql-ws'
 
-export function handle (conn) {
+function handle (conn) {
```
mcollina: Good work on the creating a test for `handle`. Can you set change `conn.pipe(conn)` to `conn.end(JSON.stringify({ error: "unknown route" }))`, and update the test accordingly?

mcollina:
Is code coverage at 100 after this change?

proti:
sadly no.
before I start working on this, was:
Screenshot 2019-11-13 at 12 21 49
and now it is:
Screenshot 2019-11-19 at 17 16 55

mcollina:
Good work so far! Would you mind bringing `index.js` up as well?

proti:
@mcollina `index.js` doesn't have any tests. I will work on this.

test/loaders.js
```diff
@@ -327,3 +327,63 @@ test('support context in loader', async (t) => {
     }
   })
 })
+test('rersolver unknown type', async t => {
```
mcollina: can you add a space before this line?

test/loaders.js
```diff
@@ -327,3 +327,63 @@ test('support context in loader', async (t) => {
...
+  })
+
+  const typeError = new Error('Cannot find type test')
+
+  try {
+    // needed so that graphql is defined
+    await app.ready()
+    app.graphql('query { test }')
+  } catch (error) {
+    t.deepEqual(error, typeError)
```
mcollina: Can you avoid creating a new error and just check the message on the generate one?

test/loaders.js
```diff
@@ -327,3 +327,63 @@ test('support context in loader', async (t) => {
...
+  })
+
+  // needed so that graphql is defined
+  await app.ready()
+
+  // TODO: How to test if cache is sets properly ?
+})
+
+test('app is not ready, throw error', async t => {
+  const app = Fastify()
```
mcollina: Add

```js
app.register(() => throw new Error('kaboom')
```

test/loaders.js
```diff
@@ -327,3 +327,95 @@ test('support context in loader', async (t) => {
...
+
+  app.register(GQL, {
+    schema,
+    resolvers,
+    loaders
+  })
+  // needed so that graphql is defined
+  await app.ready()
+
+  app.graphql()
```
mcollina: You need to call the loader here and await the result.

test/loaders.js
```diff
@@ -327,3 +327,95 @@ test('support context in loader', async (t) => {
...
+
+  app.register(GQL, {
+    cache: 256
+  })
+
+  // needed so that graphql is defined
+  await app.ready()
+
+  // TODO: How to test if cache is sets properly ?
+})
```
mcollina: Do we need this to cover some lines?

proti: unfortunately yes.

mcollina: Then just remove the comment.

test/app-decorator.js
```diff
@@ -743,3 +743,20 @@ test('union should be supported with resolveType', async (t) => {
...
+  const app = Fastify()
+  const schema = 666
+
+  app.register(GQL)
+  app.register(async function (app) {
+    app.graphql.extendSchema(schema)
+  })
+
+  try {
+    // needed so that graphql is defined
```
mcollina: Can you remove this comment?

test/loaders.js
```diff
@@ -327,3 +328,115 @@ test('support context in loader', async (t) => {
...
+test('minJit is not a number, throw error', async t => {
+  const app = Fastify()
+
+  app.register(GQL, {
+    jit: '0'
+  })
+  const typeError = new Error('the jit option must be a number')
+
+  try {
+    // needed so that graphql is defined
```
mcollina: Can you remove this comment?

test/loaders.js
```diff
@@ -327,3 +328,115 @@ test('support context in loader', async (t) => {
...
+})
+
+test('options cache is type != number', async t => {
+  const app = Fastify()
+
+  app.register(GQL, {
+    cache: 'cache'
+  })
+
+  // needed so that graphql is defined
```
mcollina: Can you remove this comment?

test/loaders.js
```diff
@@ -327,3 +328,115 @@ test('support context in loader', async (t) => {
...
+  await app.ready()
+})
+
+test('options cache is type != number', async t => {
+  const app = Fastify()
+
+  app.register(GQL, {
+    cache: 'cache'
```
mcollina: Can you make the validation of the cache option a bit tighter?

proti: what do you have in mind?

proti:
Test coverage:
Screenshot 2019-11-21 at 17 30 04

index.js
```diff
@@ -106,6 +102,8 @@ module.exports = fp(async function (app, opts) {
   fastifyGraphQl.extendSchema = function (s) {
     if (typeof s === 'string') {
       s = parse(s)
+    } else {
```
mcollina:
```suggestion
    } else if (!s || typeof s !== 'object') {
```

This will also need an updated test. Essentially if the user has already passed in a parsed AST, we should support it.

PR #91, opened 2019-12-08
Add subscription endpoint to playground

SkeLLLa:
In order to work with subscriptions playground needs additional endpoint.


mcollina:
Thanks!

PR #93, opened 2019-12-15
Fix TypeError: this.queue.close is not a function

SkeLLLa:
In some rare cases there's a possibility when `close` will be called before `subscribe` will happen. In this case that error could cause even app server crash. Also it prevents `this.queue.destroy()` to be called.

I don't know how to properly test this case, cause it's caused by some abnormal behaviour of client side. Tried some options with closing ws client right after establishing connection, sending some crappy staff into it - nothing caused this bug to appear.
Anyway it's present and needed to be fixed.

### DIFF
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -61,7 +61,9 @@ class SubscriptionContext {
   }
 
   close () {
-    this.queue.close()
+    if (typeof this.queue.close === 'function') {
+      this.queue.close()
+    }
     this.queue.destroy()
   }
 }
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -61,6 +61,8 @@ class SubscriptionContext {
   }
 
   close () {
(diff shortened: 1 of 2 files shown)


lib/subscriber.js
```diff
@@ -61,7 +61,10 @@ class SubscriptionContext {
   }
 
   close () {
-    this.queue.close()
+    /* istanbul ignore else */
+    if (typeof this.queue.close === 'function') {
+      this.queue.close()
```
mcollina: can you add a comment explaining this?

SkeLLLa: Yes. BTW I've got another idea, we can set `close` to noop function in constructor to avoid that `if`.

SkeLLLa: @mcollina added a comment.

mcollina:
Can you add a test that directly call the `Subscriber`? I think that should be possible.

SkeLLLa:
@mcollina test added.

PR #96, opened 2020-01-01
feat(deps): update fastify-websocket to 1.x.x, fix tests

SkeLLLa:
Update fastify-websockets to 1.x.x and fix tests.

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -123,7 +123,6 @@ module.exports = class SubscriptionConnection {
         payload
       }))
     } catch (e) {
-      /* istanbul ignore next */
       this.handleConnectionClose()
     }
   }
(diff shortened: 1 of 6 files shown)


lib/subscription-connection.js
```diff
@@ -123,6 +123,7 @@ module.exports = class SubscriptionConnection {
         payload
       }))
     } catch (e) {
+      /* istanbul ignore next */
```
mcollina: This was covered before, why it is not anymore? Maybe we should add a new test.

SkeLLLa: It was covered because of server tried to send data to already closed socket. And as far as I understood it was triggered just occasionally.

I've added one test, but I'd rather refactor send message method slightly:
1) Check if socket is open before trying to send anything to it. Don't know how it would impact on performance, but it seems better approach for me.
2) Since it called in a loop calling `JSON.stringify` in it may result in overhead. I think it would be better to call it once when subscription message arrives and send method should accept only strings as messages.
But I think this would be in other PR.

mcollina: Agreed!

PR #101, opened 2020-01-09
feat: add .replaceSchema method

### LINKED ISSUE #100: Proposal: Graphql Schema replacement
Hey, everyone
Does `fastify-gql` have method for replacing existing graphql schema with another one?

I'm using a microservice architecture with graphql and in some cases, I need to setup schema's autorefresh. (I'm using simple recursive setTimeout)

Example of code for Apollo-Server

```js
const refreshSchema = async () => {
	try {
		const schemaDerivedData = await graphqlServer.schemaDerivedData;

		graphqlServer.schemaDerivedData = Promise.resolve({
			...schemaDerivedData,
			schema: await generateSchema(),
		});
	} catch (err) {
		logger.warn(err, {
			method: 'server.initializeApplication::refreshSchema',
		});
	}

	setTimeout(refreshSchema, config.refreshTimeout);
};

setTimeout(refreshSchema, config.refreshTimeout);
```

Could I somehow replace the existing graphql schema with a new one?

# Implementation

```js
  fastifyGraphQl.replaceSchema = function (s) {
    if (!s || typeof s !== 'object') {
      throw new Error('Must provide valid Document AST')
    }

    schema = s
  }

```

iviaks:
This PR closes #100 

`replaceSchema` works great with `makeSchemaExecutable`

### DIFF
--- a/README.md
+++ b/README.md
@@ -495,6 +495,10 @@ run()
 It is possible to replace schema and resolvers using `makeSchemaExecutable` function in separate fastify plugins, like so:
 
 ```js
+const Fastify = require('fastify')
+const GQL = require('fastify-gql')
+const { makeExecutableSchema } = require('graphql-tools')
+
 const app = Fastify()
 
 app.register(GQL, {
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ __fastify-gql__ supports the following options:
 * `loaders`: Object. See [defineLoaders](#defineLoaders) for more
   details.
 * `graphiql`: boolean | string. Serve
-  [GraphiQL](https://www.npmjs.com/package/graphiql) on `/graphiql` if `true` or `'graphiql'`, or 
+  [GraphiQL](https://www.npmjs.com/package/graphiql) on `/graphiql` if `true` or `'graphiql'`, or
   [GraphQL IDE](https://www.npmjs.com/package/graphql-playground-react) on `/playground` if `'playground'`
   if `routes` is `true`.
 * `jit`: Integer. The minimum number of execution a query needs to be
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -104,6 +104,14 @@ module.exports = fp(async function (app, opts) {
 
   app.decorate('graphql', fastifyGraphQl)
 
+  fastifyGraphQl.replaceSchema = function (s) {
+    if (!s || typeof s !== 'object') {
+      throw new Error('Must provide valid Document AST')
+    }
+
+    schema = s
```
mcollina: You need to flush the AST and jit cache.

iviaks: I've updated PR and added

```js
lru.clear()
lruErrors.clear()
```

I hope, it should be correct.

iviaks:
@mcollina 

I've updated `README.md` file and described how to use `.replaceSchema` method.

According to resolvers.
`.replaceSchema` allows only GraphQLSchema type (we can get this type after execution `makeExecutableSchema` for example)

README.md
```diff
@@ -490,6 +490,66 @@ async function run () {
...
 
+#### app.graphql.replaceSchema(schema)
+
+It is possible to replace schema and resolvers using `makeSchemaExecutable` function in separate fastify plugins, like so:
+
+```js
+const app = Fastify()
+
+app.register(GQL, {
+  schema: makeExecutableSchema({
```
mcollina: This is imported from somewhere, can you add the relevant requires at the top?

README.md
```diff
@@ -490,6 +490,66 @@ async function run () {
 run()
 ```
 
+#### app.graphql.replaceSchema(schema)
+
+It is possible to replace schema and resolvers using `makeSchemaExecutable` function in separate fastify plugins, like so:
+
+```js
+const app = Fastify()
```
mcollina: this needs a require.

test/app-decorator.js
```diff
@@ -148,6 +148,83 @@ test('operationName', async (t) => {
...
+
+  const query = '{ add(x: 2, y: 2, z: 2) }'
+  const res = await app.graphql(query)
+
+  t.deepEqual(res, {
+    data: {
+      add: 6
+    }
+  })
+})
```
mcollina: Can you add a test where you verify that the caches are cleared? What you need is to issue the query once (it gets cached), then _remove_ that part of the schema, and then issue the same query again. The second call should give an error.

README.md
iviaks: I've added import for `makeExecutableSchema`

README.md
iviaks: I've added `Fastify` and `GQL` imports

test/app-decorator.js
iviaks: I've added a test for checking cache.
At the first call, we have two queries (`add` and `subtract`), I execute `subtract` and checking output.
At the second call, we have one query (`add`), I execute `subtract` and checking error.

PR #102, opened 2020-01-09
Refactoring: improve ide usage clarity

dougg0k:
I wanted to improve the clarity on the ide feature.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -76,7 +76,8 @@ declare namespace FastifyGQL {
     /**
      * Serve GraphiQL on /graphiql if 'graphiql', or GraphQL IDE on /playground if 'playground' and if routes is true
      */
-    ide?: boolean | string,
+		graphql?: boolean | string,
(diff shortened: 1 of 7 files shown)


dougg0k:
We could also have graphiql as the default option, without having to set anything in the ide.

index.d.ts
```diff
@@ -74,9 +74,9 @@ declare namespace FastifyGQL {
       }
     },
     /**
-     * Serve GraphiQL on /graphiql if true or 'graphiql', or GraphQL IDE on /playground if 'playground' if routes is true
+     * Serve GraphiQL on /graphiql if 'graphiql', or GraphQL IDE on /playground if 'playground' and if routes is true
      */
-    graphiql?: boolean | string,
```
mcollina: do not remove this from the types

dougg0k: Leave both then?

mcollina: yes, the option is still there

PR #104, opened 2020-01-10
Feature: Add custom path into GQL constructor

### LINKED ISSUE #103: Graphql URI
Hey, everyone
Could I somehow provide graphql URI?

For example, I need to set up service on `/` endpoint instead of `/graphql`.

I had next code for `apollo-server-hapi` library
```js
await graphqlServer.applyMiddleware({ app: server, path: '/' });
```

I guess it can be a little bit simpler (for example, parameter `path` in the constructor).

iviaks:
This PR tries to close #103 

I have not found a way of providing custom graphql endpoint to playground/graphiql.

@mcollina Could you help me with it?


mcollina:
Good work, thanks!

PR #107, opened 2020-01-14
Fix: Incorrect Graphql Errors processing

### LINKED ISSUE #106: Incorrect Graphql Error processing
The current implementation of processing Graphql Errors is not compatible with [Graphql Specification](http://spec.graphql.org/June2018/#sec-Errors).

From https://github.com/rmosolgo/graphql-ruby/issues/1130#issuecomment-347373937

> It's common practice to use 200 for all GraphQL requests. When errors happen, they're put in the "errors" key. Non-200 response codes indicate that some issue occurred at the HTTP transport layer, not the GraphQL layer.

Current implementation includes next validation for 200 status code

```js
const responseSchema = {
  '2xx': {
    type: 'object',
    properties: {
      data: {
        type: 'object',
        additionalProperties: true
      },
      errors: {
        type: 'object', // <------- Problem is here. Errors should have type array
        properties: {
          message: { type: 'string' },
          locations: {
            type: 'array',
            items: {
              type: 'object',
              properties: {
                line: { type: 'integer' },
                column: { type: 'integer' }
              }
            }
          },
          path: {
            type: 'array',
            items: { type: 'string' }
          }
        }
      }
    }
  }
}
```

Also, I noticed, that `data` can't be null.

iviaks:
PR closes #106

### DIFF
--- a/test/routes.js
+++ b/test/routes.js
@@ -312,7 +312,7 @@ test('disable routes', async (t) => {
   t.deepEqual(res.statusCode, 404)
 })
 
-test('GET return 500 on resolver error', async (t) => {
+test('GET return 200 on resolver error', async (t) => {
   const app = Fastify()
   const schema = `
     type Query {
@@ -334,11 +334,11 @@ test('GET return 500 on resolver error', async (t) => {
     url: '/graphql?query={add(x:2,y:2)}'
   })
 
-  t.equal(res.statusCode, 500) // Internal Server Error
+  t.equal(res.statusCode, 200)
   t.matchSnapshot(JSON.stringify(JSON.parse(res.body), null, 2))
 })
 
-test('POST return 500 on resolver error', async (t) => {
+test('POST return 200 on resolver error', async (t) => {
   const app = Fastify()
   const schema = `
     type Query {
@@ -364,7 +364,7 @@ test('POST return 500 on resolver error', async (t) => {
     }
   })
 
-  t.equal(res.statusCode, 500) // Internal Server Error
+  t.equal(res.statusCode, 200)
   t.matchSnapshot(JSON.stringify(JSON.parse(res.body), null, 2))
 })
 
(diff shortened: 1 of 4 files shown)


test/routes.js
```diff
@@ -849,6 +850,103 @@ test('Custom error handler', async (t) => {
...
+  `
+
+  const resolvers = {
+    add: async ({ x, y }) => { throw new GraphQLError('Graphql Error in resolver') }
+  }
+
+  function errorHandler (error, req, reply) {
+    reply.code(200)
+    reply.send({ errors: error.errors })
+  }
```
mcollina: Is this needed to test this feature?

iviaks: for now, `errors` is an object

firstly, we need to return status 200
secondary, we need to return errors as an array

mcollina: Can you change the code in routes so that this block is not needed?

iviaks:
@mcollina 
I've extended defaultErrorHandler, but I'm not sure according to `MethodNotAllowed` exception

Apollo-Server raises 404 even for DELETE method on GraphQL endpoint

Status codes from Apollo Server:
400 - Validation Error (User sent invalid request query)
404 - URL not found (Even for incorrect method, etc.)
200 - in other cases

Also, I've updated previous tests and regenerated snapshots.

bmullan91:
@iviaks could you link to the Apollo-Server source where this is - it's a bit of a labyrinth 🙈

iviaks:
@bmullan91 it's hard because I've tested on my local service :(

bmullan91:
@iviaks Take a look at express-graphql and it's http tests as a reference: https://github.com/graphql/express-graphql/blob/master/src/__tests__/http-test.js  - there are a few cases where it does return 500 for internal server error.

iviaks:
@bmullan91 Yes, you're right. But, Apollo-Server-Hapi returns 400 with the same request.

mcollina:
I would try to match the express-graphql implementation as it's the reference one.

iviaks:
@mcollina 
I will try to re-implement it in express-graphql way

iviaks:
@mcollina 
Could I move response tests into another file? (for example, `responses.js`)

iviaks:
@mcollina @bmullan91 

Okay, I have these two errors. What do you think, which parameter should we use for getting the correct status code? For the first case, it should be 500, for the second case 200

```
{ InternalServerError: Internal Server Error
    at Object.fastifyGraphQl [as graphql] (/home/iviaks/projects/asteria/testing/fastify-gql/index.js:299:19)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  message: 'Internal Server Error',
  errors:
   [ { GraphQLError: Must provide operation name if query contains multiple operations.
         at buildExecutionContext (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:181:13)
         at executeImpl (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:89:20)
         at execute (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:64:63)
         at Object.graphql (/home/iviaks/projects/asteria/testing/fastify-gql/index.js:290:29)
         at _Reply.graphql (/home/iviaks/projects/asteria/testing/fastify-gql/index.js:104:16)
         at Object.<anonymous> (/home/iviaks/projects/asteria/testing/fastify-gql/lib/routes.js:203:18)
         at preHandlerCallback (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/fastify/lib/handleRequest.js:111:30)
         at preValidationCallback (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/fastify/lib/handleRequest.js:100:5)
         at handler (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/fastify/lib/handleRequest.js:69:5)
         at done (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/fastify/lib/contentTypeParser.js:115:7)
       message:
        'Must provide operation name if query contains multiple operations.' } ] }
------------------------------------------------------------
{ InternalServerError: Internal Server Error
    at Object.fastifyGraphQl [as graphql] (/home/iviaks/projects/asteria/testing/fastify-gql/index.js:299:19)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  message: 'Internal Server Error',
  errors:
   [ { GraphQLError: Simple error
         at hello (/home/iviaks/projects/asteria/testing/fastify-gql/test/routes2.js:101:30)
         at resolveFieldValueOrError (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:467:18)
         at resolveField (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:434:16)
         at executeFields (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:275:18)
         at executeOperation (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:219:122)
         at executeImpl (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:104:14)
         at execute (/home/iviaks/projects/asteria/testing/fastify-gql/node_modules/graphql/execution/execute.js:64:63)
         at Object.graphql (/home/iviaks/projects/asteria/testing/fastify-gql/index.js:290:29)
         at _Reply.graphql (/home/iviaks/projects/asteria/testing/fastify-gql/index.js:104:16)
         at Object.<anonymous> (/home/iviaks/projects/asteria/testing/fastify-gql/lib/routes.js:203:18) message: 'Simple error', locations: [Array], path: [Array] } ] }
```

iviaks:
@mcollina @bmullan91 

Could you check the new implementation and correct me if I'm wrong?

test/routes.js
```diff
@@ -106,6 +107,118 @@ test('GET route', async (t) => {
...
+    method: 'POST',
+    url: '/graphql',
+    body: { query }
+  })
+
+  t.equal(response.statusCode, 500)
+  t.matchSnapshot(JSON.stringify(JSON.parse(response.body), null, 2))
+})
+
+test('server should return 200 on graphql errors', async (t) => {
```
bmullan91: According to the `express-graphql`, it depends on the field's type:

- If the field can be null -> 200: [source](https://github.com/graphql/express-graphql/blob/master/src/__tests__/http-test.js#L1131-L1159)
- If it's set to be NonNull, and the field throws -> 500 [source](https://github.com/graphql/express-graphql/blob/master/src/__tests__/http-test.js#L1161-L1196)

iviaks: I've found an issue with missing `data` field inside Error response and after fixing, it works as expected.

Same implementation inside `exporess-graphql`
https://github.com/graphql/express-graphql/blob/master/src/index.js#L366-L368

iviaks:
@bmullan91 @mcollina 
I've pushed with for `data` field inside Error response and it works as expected.
Could you check it?

iviaks:
@bmullan91, woho! thanks :)

PR #109, opened 2020-01-14
Add application/graphql support

### LINKED ISSUE #108: Support for Content-Type: application/graphql
Add support for [application/graphql](https://graphql.org/learn/serving-over-http/#post-request) Content-Type. 

express-graphql already supports it, while appolo-server does not.

It is true, that GQL variables can not be passed as part of the query, but it becomes much easier to use in command line, automated tests and [benchmarks](https://github.com/the-benchmarker/graphql-benchmarks).

mihaur:
PR closes #108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -391,8 +391,10 @@ payload must conform to the following JSON schema:
 
 ```js
 {
-  type: 'string',
-  description: 'the GraphQL query'
+  query: {
+    type: 'string',
+    description: 'the GraphQL query'
+  }
 }
 ```
 
--- a/README.md
+++ b/README.md
@@ -379,6 +379,28 @@ payload must conform to the following JSON schema:
 }
 ```
 
+For code from [example](#example) use:
+```sh
+curl -H "Content-Type:application/json" -XPOST -d '{"query": "{ add(x: 2, y: 2) }"}' http://localhost:3000/graphql
+```
+
+#### POST /graphql with Content-type: application/graphql
+
+Executes the GraphQL query or mutation described in the body. `operationName` and `variables` can not be passed using this method. The
+payload must conform to the following JSON schema:
+
+```js
+{
+  type: 'string',
+  description: 'the GraphQL query'
+}
+```
+
+For code from [example](#example) use:
+```sh
+curl -H "Content-Type:application/graphql" -XPOST -d "{ add(x: 2, y: 2) }" http://localhost:3000/graphql
+```
+
 #### GET /graphiql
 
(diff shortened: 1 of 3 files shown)


lib/routes.js
```diff
@@ -151,6 +151,14 @@ module.exports = async function (app, opts) {
     app.route(getOptions)
   }
 
+  app.addContentTypeParser('application/graphql', function (req, done) {
+    var data = ''
+    req.on('data', chunk => { data += chunk })
```
mcollina: Would you mind to add `req.setEncoding('utf8')`?

lib/routes.js
```diff
@@ -151,6 +151,14 @@ module.exports = async function (app, opts) {
...
+    req.on('end', _ => {
+      done(null, { query: data })
```
mcollina: How are variables being sent?

mihaur: Variables are not supported with application/graphql content-type, just query. See https://graphql.org/learn/serving-over-http/#post-request second special case.

curl -H "Content-Type:application/graphql" -XPOST -d "{ add(x: 2, y: 1) }" http://localhost:3000/graphql

mcollina: Can you add a bit of docs in https://github.com/mcollina/fastify-gql#post-graphql describing this?

mihaur: I have updated documentation and added new test with mutation via POST and application/graphql.

README.md
```diff
@@ -379,6 +379,28 @@ payload must conform to the following JSON schema:
...
+```js
+{
+  type: 'string',
+  description: 'the GraphQL query'
+}
+```
+
+For code from [example](#example) use:
+```sh
+curl -H "Content-Type:application/graphql" -XPOST -d "{ add(x: 2, y: 2) }" http://localhost:3000/graphql
```
mcollina: This body is not conforming to the description above. It does not respect the JSON schema

mihaur: I think JSON shema was OK. It accepts only a string that is a GraphQL query or mutation.
"{ add(x: 2, y: 1)}" equal to "query { add(x: 2, y: 1)}". Curly braces are part of the query. 

Internally it is transformed to { "query": "query { add(x: 2, y: 1)}" } before execution.

mcollina: A graphql query is not a JSON, so the content of the body cannot follow the JSON schema from a user perspective.

mihaur: Then I will remove the reference to JSON schema validation.

mihaur: I have removed JSON schema validation from POST application/graphql section.

mcollina:
Thanks!

PR #113, opened 2020-02-04
fix: error handler setting undefined status code

feugy:
Only set error response status code when it is defined. Otherwise [response will be truncated.](https://github.com/fastify/fastify/issues/2078).

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -54,8 +54,8 @@ const responseSchema = {
 async function defaultErrorHandler (err, request, reply) {
   if (err.data) {
     reply.code(200)
-  } else if (err.statusCode) {
-    reply.code(err.statusCode)
+  } else {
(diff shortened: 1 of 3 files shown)


lib/routes.js
```diff
@@ -54,7 +54,7 @@ const responseSchema = {
 async function defaultErrorHandler (err, request, reply) {
   if (err.data) {
     reply.code(200)
-  } else {
+  } else if (err.statusCode) {
```
mcollina: Could you add a unit test? I would also make it explicit and set the status code either to `200` or `500`. I'm not sure which one is correct.

feugy: I opted in for 500, and added a test that represent exactly the error we had on our project.
It is a schema parsing error, but strangely, it is only thrown when running request, as opposed to schema syntax errors which are raised at startup.

Issue #105, opened 2020-01-14
queryDepth being too strict


feugy:
Hello!
It seems the calculation for query depth is not considering only the depth of a query, but also hoy many fields it returns.
Here is a real example:
```graphQL
fragment metric on WeatherDatum {
  actual
  units
  delta
  style
  normal
}
fragment metricArray on WeatherData {
  actual
  normal
  units
  style
}
fragment summedMetricArray on WeatherSummedData {
  units
  array
  sum
}

query FieldMetrics($farmId: String) {
  fieldForecast(id: $farmId, unit: Metric) {
    temperature {
      today {
        ...metric
      }
      maximaOverYear {
        ...metricArray
      }
    }
    rainfall {
      accumulation {
        ...metricArray
      }
      last30DaysSummed {
        ...summedMetricArray
      }
      today {
        ...metric
      }
      yesterday {
        ...metric
      }
      nextWeek {
        ...metric
      }
    }
    country
    state
    county
    updateTime
    createdTime
  }
}
```

The maximum depth seems to be 3: fieldForecast > temperature > today
But when running the query in playground, I am receiving:
![image](https://user-images.githubusercontent.com/186268/72345638-90e33f00-36d4-11ea-9fd9-0d6ab5dd95a2.png)

PR #116, opened 2020-02-10
updates the way query depth is computed

leorossi:
@mcollina Not really sure if this new way of computing query depth covers all cases.
In the proposed test I used the following query
```
flavors {
      id
      name 
      nutrition {
        calories,
        description {
          body
          other {
            body
          }
        }
      }
      recipes {
        name
        ingredients {
          name
        }
      }
      seasons {
        name
      }
    }
```
which have a depth of 4 (`flavors -> nutrition -> description -> other`)
Fixes: #105

### DIFF
--- a/lib/queryDepth.js
+++ b/lib/queryDepth.js
@@ -14,16 +14,14 @@ const { Kind } = require('graphql')
 function queryDepth (definitions, queryDepthLimit) {
   const queries = getQueriesAndMutations(definitions)
   const queryDepth = {}
+
   for (const name in queries) {
     queryDepth[name] = determineDepth(queries[name])
   }
   const errors = []
-
   if (typeof queryDepthLimit === 'number') {
     for (const query of Object.keys(queryDepth)) {
-      const totalDepth = Object.values(queryDepth[query]).reduce((acc, curr) => {
-        return acc + curr
-      })
+      const totalDepth = queryDepth[query]
       if (totalDepth > queryDepthLimit) {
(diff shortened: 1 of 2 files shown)


lib/queryDepth.js
```diff
@@ -33,7 +31,15 @@ function queryDepth (definitions, queryDepthLimit) {
 
   return errors
 }
-
+function determineDepth (node, current = 0) {
+  let result = current
+  if (node.selectionSet) {
+    node.selectionSet.selections.forEach(selection => {
+      result = determineDepth(selection, current++)
+    })
```
mcollina: Please use a `for...of`.

lib/queryDepth.js
```diff
@@ -43,19 +49,20 @@ function getQueriesAndMutations (definitions) {
...
+//       totals[name] += 1
+//     } else {
+//       totals[name] = 1
+//     }
+//     node.selectionSet.selections.forEach(selection => {
+//       determineDepth(selection, currentDepth++, totals)
+//     })
+//   }
+//   return totals
+// }
```
mcollina: please delete

test/query-depth.js
```diff
@@ -232,3 +233,204 @@ test('queryDepth - definition.kind and definition.name change', async (t) => {
...
+              { name: 'Spring' },
+              { name: 'Summer' }
+            ]
+          }
+        ]
+      }
+    })
+  } catch (error) {
+    t.fail(error.message)
+  }
```
mcollina: There is no need for this try-catch

Issue #121, opened 2020-02-12
Unable to find 'main.js' on '/graphiql' page 


wale-A:
After setting up fastify-gql, i get a 404 error relating to the main.js file whenever i go to the '/graphiql' endpoint.

here's my setup 

``` js
export const gqlSchema = `
  type Query {
    adhesive: [String]
  }
`

export const gqlResolvers = {
  Query: {
    adhesive: () => ['1', '2', '3']
  }
}
`
```

and here's how i registered it on fastify
```js
const server: fastify.FastifyInstance<
    Server,
    IncomingMessage,
    ServerResponse
  > = fastify({
    logger
  })

  // eslint-disable-next-line
  server.register(require('fastify-gql'), {
    gqlSchema,
    gqlResolvers,
    graphiql: true
  })

```

PR #122, opened 2020-02-12
Serve .js files when ide is enabled

leorossi:
Fixes: #121

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -221,6 +221,15 @@ module.exports = async function (app, opts) {
       wildcard: false,
       serve: false
     })
+
+    app.get('/main.js', (req, reply) => {
+      reply.sendFile('main.js')
+    })
+
+    app.get('/sw.js', (req, reply) => {
+      reply.sendFile('sw.js')
+    })
+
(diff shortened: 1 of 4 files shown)


lib/routes.js
```diff
@@ -221,6 +221,15 @@ module.exports = async function (app, opts) {
...
       serve: false
     })
+
+    app.get('/main.js', (req, reply) => {
+      reply.sendFile('main.js')
+    })
+
+    app.get('/sw.js', (req, reply) => {
+      reply.sendFile('sw.js')
+    })
```
mcollina: You'll need to serve those only for graphiql.

test/routes.js
```diff
@@ -1342,3 +1342,22 @@ test('render graphiql if graphiql: true', async (t) => {
...
+test('if ide is set, always serve main.js and sw.js', async (t) => {
+  const app = Fastify()
+  app.register(GQL, {
+    ide: true
+  })
+
+  const res = await app.inject({
+    method: 'GET',
+    url: '/main.js'
+  })
```
mcollina: I would wrap them in a prefix, i.e. `/graphiql/main.js`.

PR #127, opened 2020-03-08
Improve Typescript definitions

ShogunPanda:
This PR:

1. Improved TypeScripts types by naming the namespace and the plugin function using one symbol. This make sure that both the plugin its inner definitions (especially `Options`) are exported.
2 Fixed the return type of the graphql decorator. It was mistakenly reported as a fastify response, but it is actually the graphql execution result, and also it is always returned a promise.

Hope this helps!

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -145,7 +145,7 @@ declare module "fastify" {
       context?: any,
       variables?: { [key: string]: any },
       operationName?: string
-    ): FastifyReply<HttpResponse>;
+    ): Promise<ExecutionResult>;
   }
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,15 +1,10 @@
-import fastify, {
-  FastifyError,
-  FastifyReply,
-  FastifyRequest,
-  RegisterOptions
-} from "fastify"
+import fastify, { FastifyError, FastifyReply, FastifyRequest, RegisterOptions } from "fastify";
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,15 +1,10 @@
...
+import fastify, { FastifyError, FastifyReply, FastifyRequest, RegisterOptions } from "fastify";
+import { DocumentNode, ExecutionResult, GraphQLSchema, Source } from 'graphql';
 import { IResolvers } from "graphql-tools";
-import { Server, IncomingMessage, ServerResponse } from "http";
+import { IncomingMessage, Server, ServerResponse } from "http";
 import { Http2Server, Http2ServerRequest, Http2ServerResponse } from 'http2';
-import graphql, { GraphQLSchema, GraphQLError, Source, DocumentNode, ExecutionResult } from 'graphql';
 
-declare namespace FastifyGQL {
+declare namespace fastifyGQL {
```
mcollina: Is this change of capitalization important?

ShogunPanda: Yup. It the main function matches the namespace, the two declarations are merged.
This ensures that the plugin can be used as it was but it makes the Options and Plugin declaration outside. While the latter might be useless, the former is important if you want to create factory functions that internally register the plugin.

ShogunPanda:
> Could you please add a test for this change?

Sure, will do soon.

ShogunPanda:
Test added!

PR #129, opened 2020-03-19
Support federation protocol

frikille:
This PR adds enables federation support so the service can be deployed behind an Apollo Gateway.

Based on the [specification](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/) there are 4 requirements
- [x] Implementation of the federation schema specification
- [x] Support for fetching service capabilities
- [x] Implementation of stub type generation for references
- [x] Implementation of request resolving for entities

A fork of the original [federation-demo](https://github.com/apollographql/federation-demo) updated to use `fastify-gql` can be found [here](https://github.com/frikille/federation-demo/tree/fastify-example)


lib/federation.js
```diff
@@ -0,0 +1,95 @@
+const {
```
mcollina: Use strict is missing

index.d.ts
```diff
@@ -121,7 +121,11 @@ declare namespace fastifyGQL {
         info: object,
         next: (result: boolean) => void
       ) => void
-    }
+    },
+    /**
+     * Enable federation support so the service can be deployed behind an Apollo Gateway
+     */
+    enableFederation?: boolean
```
mcollina: How about we call this `federationMetadata`? I think it reflects more the fact that it will serve the federation metadata itself.

lib/federation.js
```diff
@@ -0,0 +1,210 @@
...
+
+  const parsedOriginalSchema = parse(originalSchema)
+
+  const missingTypeDefinitions = []
+  const definitionsMap = {}
+  const extensionsMap = {}
+  const directiveDefinitions = []
+
+  for (const definition of parsedOriginalSchema.definitions) {
+    /* istanbul ignore else */
```
mcollina: Can you remove these `else`?

lib/federation.js
```diff
@@ -0,0 +1,210 @@
...
+# directive @extends on OBJECT | INTERFACE
+`
+
+function hasDirectives (node) {
+  return Boolean('directives' in node && node.directives)
+}
+
+function gatherDirectives (type) {
+  let directives = []
+  if ('extensionASTNodes' in type && type.extensionASTNodes) {
```
mcollina: Please use hasOwnProperty instead

frikille: If I use that the lint fails (`Do not access Object.prototype method 'hasOwnProperty' from target object`)

mcollina: Yes, you need to get the function from the prototype and use `call()`.

lib/federation.js
```diff
@@ -0,0 +1,227 @@
...
+    `), {
+      assumeValid: true
+    })
+
+    const query = schema.getType('Query')
+    const queryFields = query.getFields()
+    queryFields._entities = {
+      ...queryFields._entities,
+      resolve: (_source, { representations }, context, info) => {
+        return representations.map((reference) => {
```
mcollina: This implementation is going to cause massive performance issues. We might want to add support for the loaders instead.

frikille: It needed a small modification in the `__resolveReference` signature but it is possible to define that method as a resolver which solves the performance issue.

Added a test for this case along with the change.

test/federation.js
```diff
@@ -688,3 +688,112 @@ test('subscription server sends update to subscriptions', t => {
...
+            params: {}
+          }, {
+            obj: {
+              __typename: 'User',
+              id: '2'
+            },
+            params: {}
+          }])
+          return queries.map(({ obj }) => users[obj.id])
+        },
```
mcollina: This needs to be documented / we should add an example somewhere. Note that without loader this is going to create _1+N_ performance issues.

frikille: Added a new section to the readme

README.md
```diff
@@ -292,6 +292,140 @@ app.register(GQL, {
...
+      return users['1']
+    }
+  }
+}
+
+const loaders = {
+  User: {
+    __resolveReference: {
+      async loader(queries, context) {
+        return queries.map(({ obj }) => users[obj.id])
```
mcollina: Can you add a comment saying that this should be a _bulk_ query to a database?

README.md
```diff
@@ -292,6 +292,140 @@ app.register(GQL, {
...
+        cache: true
+      }
+    }
+  }
+}
+
+app.register(GQL, {
+  schema,
+  resolvers,
+  federationMetadata: true
```
mcollina: You are not passing the loaders option.

README.md
```diff
@@ -292,6 +292,142 @@ app.register(GQL, {
...
+const loaders = {
+  User: {
+    __resolveReference: {
+      async loader(queries, context) {
+        // This should be a bulk query to the database
+        return queries.map(({ obj }) => users[obj.id])
+      },
+      opts: {
+        cache: true
+      }
```
mcollina: I don't think we support `cache: true` at all here.

frikille: It is passed to the factory https://github.com/mcollina/fastify-gql/blob/ca23bb6c22d88b021a4b4503cfae321895e5dc8d/index.js#L186 but it's the same as not setting the `opts` property at all and passing only the loader function.

PR #130, opened 2020-03-26
GraphQL Gateway

frikille:
This PR will add the option to the plugin that enables running the GraphQL server in gateway mode.

Goals of current PR:
- [x] Support queries 
- [x] Support mutations
- [x] Support using fragments in queries
- [x] Support using variables in queries
- [x] Support JIT compiler
- [x] Always add` @key` field to an entity resolver query
- [x] Always add `@required` field to an entity resolver query

Supporting subscriptions are not part of this PR but is recognised as an essential feature and will be added later.

An example can be found here: [federation-demo](https://github.com/frikille/federation-demo/tree/fastify-example)


lib/gateway.js
```diff
@@ -0,0 +1,141 @@
...
+          }, (err, response) => {
+            if (err) {
+              return reject(err)
+            }
+
+            if (response.statusCode === 200) {
+              response.stream.on('data', data => resolve({
+                statusCode: response.statusCode,
+                json: JSON.parse(data.toString())
+              }))
```
mcollina: This is not going to work. You'll need to accumulate all the chunks.

lib/gateway.js
```diff
@@ -0,0 +1,141 @@
...
+    body: JSON.stringify({
+      query: `
+      query ServiceInfo {
+        _service {
+          sdl
+        }
+      }
+      `
+    })
+  }).then(response => response.json.data._service.sdl)))
```
mcollina: It seems that request is callback-based and not promise based.

index.js
```diff
@@ -59,14 +60,19 @@ module.exports = fp(async function (app, opts) {
 
   const root = {}
   let schema = opts.schema
+  let gateway = opts.gateway
+
+  if (gateway && (schema || opts.resolvers || opts.loaders || opts.subscription)) {
+    throw new Error('Adding "schema", "resolvers", "loaders" or "subscription" to plugin options when plugin is running in gateway mode is not allowed')
+  }
```
mcollina: There are a few methods to change the schema/resolvers/loaders etc at aruntime, you'll need to throw in them as well.

lib/gateway.js
```diff
@@ -0,0 +1,141 @@
...
+    'https:': https
+  }
+
+  function close () {
+    agents['http:'].destroy()
+    agents['https:'].destroy()
+  }
+
+  function request (opts, done) {
+    console.log('sending request', opts.method, opts.url.port, opts.url.pathname, opts.url.hostname)
```
mcollina: I think you might want to log using fastify logger instead.

lib/gateway.js
```diff
@@ -0,0 +1,250 @@
...
+          schemaDefinition,
+          typeMap
+        }
+      }
+    }
+
+    serviceMap[service.name] = serviceConfig
+  }
+
+  await Promise.all(services.map(service => {
```
mcollina: This is not going to work well if there are many services. Use http://npm.im/p-map with a max concurrency of 8.

lib/gateway.js
```diff
@@ -0,0 +1,250 @@
...
+  ].includes(type)
+}
+
+function makeResolveQuery (service) {
+  return function (root, args, context, info) {
+    const { fieldName, returnType, operation } = info
+
+    // Must deep-copy the operation because if we modify it (remove an item from the selection set)
+    // then the resolvers for those fields won't be executed
+    const currentOperation = JSON.parse(JSON.stringify(operation))
```
mcollina: This is terribly slow. Please check if you can use http://npm.im/rdfc

lib/gateway.js
```diff
@@ -0,0 +1,250 @@
...
+    currentOperation.selectionSet.selections[0].selectionSet.selections = currentOperation.selectionSet.selections[0].selectionSet.selections.filter(selection => service.typeMap[returnType].has(selection.name.value))
+
+    return service.request({
+      method: 'POST',
```
mcollina: We might be missing the content-type and content-length.

lib/gateway.js
```diff
@@ -0,0 +1,250 @@
...
+              field.resolve = makeResolveQuery(serviceMap[serviceForFieldType])
+            } else {
+              field.resolve = makeFieldResolver(serviceMap[serviceForFieldType])
+            }
+          }
+        }
+      }
+    }
+  }
+}
```
mcollina: Would you mind to add a bit of comments to this function, it's very complex.

lib/gateway.js
```diff
@@ -0,0 +1,476 @@
...
+              response.stream.on('data', chunk => {
+                data += chunk
+              })
+
+              response.stream.on('end', () => {
+                resolve({
+                  statusCode: response.statusCode,
+                  json: JSON.parse(data.toString())
+                })
+              })
```
mcollina: please use end-of-stream or stream.finished.

lib/gateway.js
```diff
@@ -0,0 +1,476 @@
...
+    }
+  }])
+}
+
+function makeResolveQuery (service) {
+  return function (root, args, context, info) {
+    const { fieldName, returnType, fieldNodes } = info
+    const type = getNamedType(returnType)
+
+    let selections = JSON.parse(JSON.stringify(fieldNodes[0].selectionSet.selections))
```
mcollina: please use http://npm.im/rdfc if possible (if not put a comment that it is not really possible)

frikille: I tried and it is not possible. The fields array is not a simple object and it throws a max call stack size exceeded error.

lib/gateway.js
frikille: I tried and it is not possible. The fields array is not a simple object and it throws a max call stack size exceeded error.

lib/gateway.js
mcollina: If it's not a simple object this should not really work.

lib/gateway.js
```diff
@@ -0,0 +1,476 @@
...
+            if (response.statusCode === 200) {
+              let data = ''
+              response.stream.on('data', chunk => {
+                data += chunk
+              })
+
+              response.stream.on('end', () => {
+                resolve({
+                  statusCode: response.statusCode,
+                  json: JSON.parse(data.toString())
```
mcollina: This can throw, we need to wrap it in a try/catch.

lib/gateway.js
frikille: It was possible to remove this in the lastest change

lib/gateway.js
frikille: It was possible to remove this in the lastest change

lib/gateway/service-map.js
```diff
@@ -0,0 +1,109 @@
...
+            if (response.statusCode === 200) {
+              let data = ''
+              response.stream.on('data', chunk => {
+                data += chunk
+              })
+
+              response.stream.on('end', () => {
+                resolve({
+                  statusCode: response.statusCode,
+                  json: JSON.parse(data.toString())
```
mcollina: This needs a try-catch as it can throw on invalid json

lib/gateway/service-map.js
```diff
@@ -0,0 +1,109 @@
...
+                })
+              })
+            }
+          })
+        })
+      },
+      close,
+      init: async () => {
```
mcollina:
```suggestion
      async init () {
```

lib/gateway/service-map.js
```diff
@@ -0,0 +1,109 @@
...
+function createTypeMap (schemaDefinition) {
+  const parsedSchema = parse(schemaDefinition)
+  const typeMap = {}
+
+  for (const definition of parsedSchema.definitions) {
+    if (
+      isTypeDefinitionNode(definition) ||
+      (isTypeExtensionNode(definition) && definition.name.value === 'Query')
+    ) {
+      typeMap[definition.name.value] = new Set(definition.fields.map(field => field.name.value))
```
mcollina: I would avoid `map` and I'll iterate over `definition.fields` using a for loop that adds everything to the `Set` or you can use a `reduce` operation.

lib/gateway.js
frikille: Added comment, will do the same for the `makeResolver` method as that one is complex as well.

index.js
```diff
@@ -59,14 +60,19 @@ module.exports = fp(async function (app, opts) {
 
   const root = {}
   let schema = opts.schema
+  let gateway = opts.gateway
+
+  if (gateway && (schema || opts.resolvers || opts.loaders || opts.subscription)) {
```
PacoDu, line 65: Thanks for this PR and the federation implementation 😃👍 

Why subscriptions aren't available in a federated environment ? Wouldn't it be possible to forward the websocket connection to the related services ?

frikille: Subscription support is coming but only after this first version is working as expected. This PR already adds a lot of complexity, and I would like to keep the PR as small as possible.

PacoDu: Nice ! I was wondering if there was any limitation, thanks !

frikille: To be honest I haven't looked into it yet. But I don't think it is as simple as forwarding WebSocket connections to the related services because you can fetch data in the subscription service that lives in some other service.

PacoDu: I don't know much about federation, but for what I know about subscriptions in apollo-server:

A subscription is defined with two functions a `subscribe` and `resolve`, subscribe could be forwarded to the service which declared the subscription and I think we could resolve the subscription the same way the gateway retrieves the fragments for a federated query or mutation. But I don't know if this would work or cover every use cases 🤔.

lib/gateway/make-resolver.js
```diff
@@ -0,0 +1,347 @@
...
+      parentType,
+      fieldName,
+      selections,
+      isQuery,
+      isReference,
+      variableDefinitions: variablesToDefine,
+      args: fieldNodes[0].arguments
+    })
+
+    let query = print(operation)
```
mcollina, line 306:
```suggestion
    // TODO: implement caching strategy as createOperation and print are expensive.
    let query = print(operation)
```

test/gateway/request.js
```diff
@@ -0,0 +1,48 @@
...
+})
+
+test('sendRequest method rejects when request errs', t => {
+  const app = fastify()
+  app.post('/', async (request, reply) => {
+    return 'response'
+  })
+  t.tearDown(() => app.close())
+
+  app.listen(3002).then(() => {
```
mcollina, line 30: please use `await`

test/gateway/schema.js
```diff
@@ -0,0 +1,265 @@
...
+    },
+    p4: {
+      pid: 'p4',
+      title: 'Post 4',
+      content: 'Content 4',
+      authorId: 'u2'
+    }
+  }
+
+  await createService(t, 4001, `
```
mcollina, line 61: Please avoid using defined ports in the tests. Those should not be set (or set to `0`), and you should get the port or the URL as the result of `createService`.

lib/gateway.js
```diff
@@ -140,6 +148,18 @@ async function buildGateway (gatewayOpts) {
         typeFieldsToService[`${type}-${field}`] = service
       }
     }
+
+    factory.add(`${service}Entity`, async (queries, context) => {
+      return queries.map(({ query, variables }) => {
```
mcollina, line 153: This should be one single http request with the same entity, or possibly coalescing all the queries into one.

frikille: Done in the latest commit

lib/gateway.js
```diff
@@ -0,0 +1,218 @@
...
+      const result = []
+
+      await Promise.all(Object.entries(mergedQueries).map(async ([query, variables], queryIndex, entries) => {
+        const response = await serviceDefinition.sendRequest({
+          method: 'POST',
+          body: JSON.stringify({
+            query,
+            variables
+          })
+        })
```
mcollina, line 183: I think you can also combine the queries. It's possible using GraphQL.

frikille: Added a detailed comment here: https://github.com/mcollina/fastify-gql/pull/130/files#diff-8341b2b42feef9f7f0e004614072f8bdR153

How it is possible, but will be part of a future release

README.md
```diff
@@ -453,6 +498,10 @@ __fastify-gql__ supports the following options:
   * `subscription.emitter`: Custom emitter
   * `subscription.verifyClient`: `Function` A function which can be used to validate incoming connections.
 * `federationMetadata`: Boolean. Enable federation metadata support so the service can be deployed behind an Apollo Gateway
+* `gateway`: Object. Run the GraphQL server in gateway mode.
+  * `gateway.services`: Service[] An array of GraphQL services that are part of the gateway
+    * `service.name`: A unique name for the service. Required.
+    * `service.url` The url of the service endpoint. Required
```
mcollina, line 504: `rewriteHeaders` needs to be documented.

PR #132, opened 2020-04-03
Update GraphQL to version 15.0.0

frikille:
This PR fixes the tests that are broken in #131.

The tests were broken because of changes in the error messages.

The Typescript check also failed because in the `node_modules/graphql-tools` package has two TS errors related to the schema stitching which is not used by this project at all. This can be avoided if the `IResolvers` type is not imported from the `graphql-tools` lib but included in the `types.d.ts` file.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,6 +1,5 @@
 import fastify, { FastifyError, FastifyReply, FastifyRequest, RegisterOptions } from "fastify";
-import { DocumentNode, ExecutionResult, GraphQLSchema, Source } from 'graphql';
-import { IResolvers } from "graphql-tools";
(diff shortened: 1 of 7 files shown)


numToStr:
Any ETA when this going to be released?

mcollina:
I’m waiting for a new version of graphql-jit. I’m also uncertain to ship this together with the support for fastify v3 or not.

wdyt?

numToStr:
I guess, you know better than me ✌🏻. But, If fastify v3 support is not a breaking change than you can ship it in a minor version. Bcz most of the gql stuffs are already migrated to v15.

BTW thanks for giving us some of the most awesome libs for nodejs ♥️.

mcollina:
Are there any new features you plan to use or some incompatibilities you know about this change?

numToStr:
Nothing major. I am just using type-graphql w/ fastify-gql. And, type-graphql already released v1.rc w/ v15 support. So, I can't use that until fastify-gql gets v15 support.

PR #137, opened 2020-04-16
Subscription support in gateway

frikille:
This PR enables subscriptions when the server runs in gateway mode.

Todo:
- [x] Enable WebSocket connection to the gateway
- [x] Forward subscription to correct service
- [x] Forward received data from service to all clients
- [x] Generate subscription query
- [x] Support full query execution after data received from service and pushed to clients
- [x] Websocket client updates in gateway


lib/gateway/service-map.js
```diff
@@ -84,6 +85,37 @@ async function buildServiceMap (services) {
       }
     }
 
+    if (service.wsUrl) {
+      const client = createClient(service.wsUrl, subscriber, service.name)
+
+      const subscriptionQueryMap = {}
```
mcollina, line 91: how are we cleaning data from this map? Looks like a memory leak.

lib/gateway/socket-connection.js
```diff
@@ -0,0 +1,26 @@
...
+
+function createClient (url, subscriber, serviceName) {
+  const ws = new WebSocket(url, 'graphql-ws')
+  const client = WebSocket.createWebSocketStream(ws, { encoding: 'utf8', objectMode: true })
+  client.setEncoding('utf8')
+  client.write(JSON.stringify({
+    type: 'connection_init'
+  }))
+
+  client.on('data', async (chunk) => {
```
mcollina, line 13: Don't do this. This is going to leak memory in case of errors.

lib/gateway/make-resolver.js
```diff
@@ -382,13 +382,20 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
...
         ...removeNonIdProperties(parent, parentType),
         ...getRequiredFields(parent, schema.getType(parentType).getFields()[fieldName])
       }]
     }
 
+    if (isSubscription) {
+      const subscriptionId = service.createSubscription(query, variables)
+      const subscription = await pubsub.subscribe(`${service.name}_${subscriptionId}`)
+
+      return subscription
```
mcollina, line 396:
```suggestion
      return pubsub.subscribe(`${service.name}_${subscriptionId}`)
```

We can save a few promises in this way. Also remove `async` at the top.

test/gateway/subscription.js
```diff
@@ -0,0 +1,272 @@
...
+                  }
+                `
+              }
+            })
+          }
+        })
+      })
+    })
+  })
+})
```
mcollina, line 272: This is way too much nesting for my taste, please linearize it a bit by using an async function.

lib/gateway/socket-connection.js
frikille: This was replaced by the `subscription-client.js`

mcollina:
Good work!

test/gateway/subscription.js
frikille: I have a version with awaits but I can't make it work somehow.

lib/subscription-connection.js
```diff
@@ -1,25 +1,30 @@
 const { subscribe, parse } = require('graphql')
```
mcollina, line 1: add 'use strict'

test/subscription-client.js
```diff
@@ -0,0 +1,200 @@
...
+
+  setTimeout(() => {
+    server = new WS.Server({ port: 8891 })
+    server.on('connection', function connection (ws) {
+      throw new Error('Client connected when it should not!')
+    })
+    client.close()
+    server.close()
+    t.end()
+  }, 2000)
```
mcollina, line 134: Could we try to avoid `setTimeout` and use lolex or similar to fake the timers? This makes tests brittle

test/subscription-client.js
```diff
@@ -0,0 +1,200 @@
...
+      payload: null
+    })
+    client.close()
+    server.close()
+    t.end()
+  })
+})
+
+test('subscription client tries to reconnect when server closes', (t) => {
+  let server = new WS.Server({ port: 8890 })
```
mcollina, line 69: Plese use `0` as a port.

test/subscription-client.js
```diff
@@ -0,0 +1,200 @@
...
+      ws.on('message', function incoming (message) {
+        const data = JSON.parse(message)
+        if (data.type === 'start') {
+          ws.send(JSON.stringify({ id: '1', type: 'complete' }))
+        }
+      })
+
+      ws.send(JSON.stringify({ id: undefined, type: 'connection_ack' }))
+    })
+  }, 2000)
```
mcollina, line 103: please use lolex instead of the timer.

frikille: refactored the test and removed all setTimeout usage

PR #139, opened 2020-04-21
feat: support for persisted queries

salmanm:
First draft for https://github.com/mcollina/fastify-gql/issues/110

- Added support to accept `persistedQueries` as the plugin options. It's a JSON object containing hash -> query mapping which is used by GET queries handler to resolve to full query text.


examples/persisted-queries/queries.json
```diff
@@ -0,0 +1,5 @@
+{
+  "1111": "{ add(x: 1, y: 1) }",
+  "2222": "query Add($x: Int!, $y: Int!) { add(x: $x, y: $y) }",
+  "3333": "{ add(x: 3, y: 3) }"
```
mcollina, line 4: Can you use proper hashes?

lib/routes.js
```diff
@@ -118,6 +119,9 @@ module.exports = async function (app, opts) {
           query: {
             type: 'string'
           },
+          isPersisted: {
```
mcollina, line 122:
```suggestion
          persisted: {
```

salmanm:
Added option `onlyPersisted`. Also added some tests.

frikille:
Shouldn't persisted queries be supported via `POST` as well? It would increase security by rejecting unknown queries.

salmanm:
Good point. The initial plan wasn't to make the server reject unknown queries, but now it is. So it'd make sense to do so on post too. Updating.

README.md
```diff
@@ -91,6 +91,39 @@ app.get('/', async function (req, reply) {
 app.listen(3000)
 ```
 
+### persistedQueries support
+
```
thomheymann, line 95: I would add a quick explanation about what persisted queries are and why they're useful

salmanm: Yep, that'd be useful. Done.

Issue #144, opened 2020-05-02
Support for extended errors in results


yndc:
Hello, it's me again 😁

In the [GraphQL's spec in errors result format](https://spec.graphql.org/June2018/#sec-Errors), GraphQL services may provide an additional entry to errors with key `extensions`.

Example:
```
{
  "errors": [
    {
      "message": "Name for character with ID 1002 could not be fetched.",
      "locations": [ { "line": 6, "column": 7 } ],
      "path": [ "hero", "heroFriends", 1, "name" ],
      "extensions": {
        "code": "CAN_NOT_FETCH_BY_ID",
        "timestamp": "Fri Feb 9 14:33:09 UTC 2018",
        "id": "1002"
      }
    }
  ]
}
```
Apollo supports this through their custom error, `ApolloError`. I think it'll be really useful for this library to have the same kind of method to provide additional information on errors. 

Thanks.

PR #145, opened 2020-05-02
Extended errors

yndc:
This is a PR for adding feature described in #144 

I'd like for some input on my implementation and naming. 
- Is `FastifyGraphQLError` class for the extended error suitable?
- How should we structure the error parameters? At the moment I'm using Apollo's `ApolloError` format; i.e. `(message: string, code: string, additionalProperties: object)`

### DIFF
--- a/index.js
+++ b/index.js
@@ -10,6 +10,7 @@ const {
   parse,
   buildSchema,
   getOperationAST,
+  GraphQLError,
   GraphQLObjectType,
   GraphQLScalarType,
   GraphQLEnumType,
@@ -26,6 +27,7 @@ const buildFederationSchema = require('./lib/federation')
 const buildGateway = require('./lib/gateway')
 const mq = require('mqemitter')
 const { PubSub } = require('./lib/subscriber')
+const { FastifyGraphQLError } = require('./lib/errors')
 
 const kLoaders = Symbol('fastify-gql.loaders')
 
@@ -361,6 +363,15 @@ module.exports = fp(async function (app, opts) {
     )
 
     if (execution.errors) {
+      execution.errors = execution.errors.map(e => {
+        if (e.originalError instanceof FastifyGraphQLError) {
+          return new GraphQLError(e.originalError.message, e.nodes, e.source, e.positions, e.path, e.originalError, {
+            code: e.originalError.code,
+            ...e.originalError.additionalProperties
+          })
+        }
+        return e
+      })
       const err = new InternalServerError()
       err.errors = execution.errors
       err.data = execution.data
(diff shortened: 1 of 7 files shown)


index.js
```diff
@@ -361,6 +363,15 @@ module.exports = fp(async function (app, opts) {
     )
 
     if (execution.errors) {
+      execution.errors = execution.errors.map(e => {
+        if (e.originalError instanceof FastifyGraphQLError) {
```
mcollina, line 367: I would prefer if we duck-typed this instead of checking for inheritance, so any error with `code` and `additionalProperties`   would work.

lib/routes.js
```diff
@@ -37,10 +37,7 @@ const responseSchema = {
             },
             extensions: {
               type: 'object',
-              properties: {
-                code: { type: 'string' },
-                timestamp: { type: 'string' }
-              }
+              additionalProperties: true
```
yndc, line 40: Would this cause security issues?

lib/errors.js
```diff
@@ -0,0 +1,13 @@
...
+  constructor (message, code, additionalProperties) {
+    super(message)
+    this.code = code
+    this.additionalProperties = additionalProperties
+  }
+}
+
+module.exports = {
+  FastifyGraphQLError: FastifyGraphQLError
+}
```
mcollina, line 13: I would call this `ErrorWithProps` or similar and not have the name of the framework in it.

lib/routes.js
```diff
@@ -37,10 +37,7 @@ const responseSchema = {
             },
             extensions: {
               type: 'object',
-              properties: {
-                code: { type: 'string' },
-                timestamp: { type: 'string' }
-              }
```
mcollina, line 43: why were this removed?

yndc: Hmm, yes I'll re-add `code`, but should `timestamp` stays? I think it shouldn't be hard-coded as a string. There might be an use-case where a developer want's `extensions.timestamp` to be a `number` type.

yndc:
Should we use the current error constructor (with `code`) or just let the developer shape everything inside `extensions` themselves to make it less opinionated?

Then `ErrorWithProps` constructor will only be `new ErrorWithProps(message: string, additionalProperties: object)`

lib/routes.js
mcollina: no this is ok.

lib/routes.js
```diff
@@ -37,10 +37,8 @@ const responseSchema = {
...
+              code: { type: 'string' },
```
mcollina, line 40: this needs to be in properties

yndc: whoops! missed that

PR #148, opened 2020-05-04
fix: typo in reply

bvanneerven:
I found this typo by accident and thought it would be nice to create a PR to fix it :)


mcollina:
Nice one! Would you mind to add a test for it?

PR #152, opened 2020-05-05
Variable validation before query execution

yndc:
I found an inconsistency in returned error codes when handling user input errors. Syntax errors in queries will cause `Bad Request` (400) error, which is correct. However missing variables such as this example:

```js
const app = Fastify();
const schema = `
    type Query {
      add(x: Int, y: Int): Int
    }
  `;

const resolvers = {
  add: async ({ x, y }) => x + y,
};

app.register(GQL, {
  schema,
  resolvers,
});

const query = "query ($x: Int!, $y: Int!) { add(x: $x, y: $y) }";

const res = await app.inject({
  method: "GET",
  url: `/graphql?query=${query}&variables=${JSON.stringify({ x: 5 })}`, // notice there's no 'y' 
});

```

Along with other errors related to query variables like mis-typed variables would cause `Internal Server Error` (500) errors which isn't correct since this problem is coming from the client-side providing wrong variables. 

In this PR I added variable validation just before the execution of the query to produce `Bad Request` (400) error on invalid variables without affecting execution errors.

I also added a test case for requests with missing variables scenario I described above.

### DIFF
--- a/index.js
+++ b/index.js
@@ -22,6 +22,7 @@ const {
   validateSchema,
   execute
 } = require('graphql')
+const { buildExecutionContext } = require('graphql/execution/execute')
 const queryDepth = require('./lib/queryDepth')
 const buildFederationSchema = require('./lib/federation')
 const buildGateway = require('./lib/gateway')
@@ -325,6 +326,16 @@ const plugin = fp(async function (app, opts) {
         throw err
       }
 
+      // Validate variables
+      if (variables !== undefined) {
+        const executionContext = buildExecutionContext(schema, document, root, context, variables, operationName)
+        if (Array.isArray(executionContext)) {
+          const err = new BadRequest()
+          err.errors = executionContext
+          throw err
+        }
+      }
+
       if (queryDepthLimit) {
         const queryDepthErrors = queryDepth(document.definitions, queryDepthLimit)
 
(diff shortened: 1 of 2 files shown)


yndc:
I benchmarked my changes and I'm not seeing any loss in performance in queries with variables despite effectively validating the variables twice (once in `buildExecutionContext` and `execute`).

However I see major performance loss in queries with wrong variables (about 75% request-per-second loss) on this `throw` line. 
```js
if (variables !== undefined) {
  const executionContext = buildExecutionContext(schema, document, root, context, variables, operationName)
  if (Array.isArray(executionContext)) {
    const err = new BadRequest()
    err.errors = executionContext
    throw err // line causing performance loss
  }
}
```
I don't know why `throw`-ing would be so expensive.

index.js
```diff
@@ -373,7 +384,7 @@ const plugin = fp(async function (app, opts) {
 
     if (execution.errors) {
       execution.errors = execution.errors.map(e => {
-        if (Object.prototype.hasOwnProperty.call(e.originalError, 'code') || Object.prototype.hasOwnProperty.call(e.originalError, 'additionalProperties')) {
+        if (e.originalError && (Object.prototype.hasOwnProperty.call(e.originalError, 'code') || Object.prototype.hasOwnProperty.call(e.originalError, 'additionalProperties'))) {
```
yndc, line 387: Sometimes errors from `execute` doesn't have `originalError`. This fixes some problems that arise when handling execution errors.

yndc:
Alright, I forgot that you could have different variables with the same (cached) query.

frikille:
lgtm!

Issue #151, opened 2020-05-04
Enable building context for subscriptions


bvanneerven:
This issue looks like a duplicate of #47 (which is already implemented :tada: ), but differs because it requests adding a similar API to allow building the `SubscriptionContext` instead of the context for queries/mutations.

A possible way of doing this would be like:

```js
app.register(GQL, {
  schema,
  resolvers,
  subscription: {
      // This would add { foo: "bar" } to the SubscriptionContext, such that it is 
      // available in subscription resolvers
      context: () => ({ foo: "bar"})
  }
})
```

The main use case for this, in my case, is creating the possibility to subscribe to user-specific content. That is, when a client sends a subscription query and includes an `Authorization` header, we populate the `SubscriptionContext` with the user information extracted from the Bearer token and use that in the subscription resolver to send updates only to that user.

```js
// server.js
app.register(GQL, {
  schema,
  resolvers,
  subscription: {
      context: (req) => ({ user: jwt.verify(req.headers["Authorization"].slice(7))})
  }
})


// resolvers.js
const resolvers = {
  Mutation: {
    sendMessage: async (_, { message, user }, { pubsub }) => {
      await pubsub.publish({
        topic: user.id,
        payload: message
      })

      return "OK"
    }
  },
  Subscription: {
    receivedMessage: {
      subscribe: (root, args, { pubsub, user }) => pubsub.subscribe(user.id)
    }
  }
}
```

I have not yet found an intuitive way to support this use case, which is why I am submitting this issue. However, since fastify is still new to me, I could be the problem too :sweat_smile:. If that is the case, I'd be happy to add the solution to the docs if someone can point me in the right direction!

mcollina:
It would be great to add support for this!

@frikille what do you think?

frikille:
I think it's a good idea and should be implemented (I looked into it earlier today if the `options.context` function is passed to the subscription, but no, it's not).

I think the main question is whether we want to add a separate context builder function or re-use the one that can be already set in the plugin options.

mcollina:
Would the API be different?

bvanneerven:
Very cool that you guys are willing to add this functionality! :smiley:. I went ahead and created a first prototype for this in #153.

asciidisco:
Closing as it has been merged in https://github.com/mcollina/fastify-gql/pull/233

PR #153, opened 2020-05-05
Add subscription context

bvanneerven:
This PR tries to implement the functionality described in #151 .

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -81,8 +81,7 @@ module.exports = class SubscriptionConnection {
 
     const sc = new SubscriptionContext({
       fastify: this.fastify,
-      pubsub: this.subscriber,
-      context: this.context
+      pubsub: this.subscriber
     })
     this.subscriptionContexts.set(id, sc)
 
@@ -93,6 +92,7 @@ module.exports = class SubscriptionConnection {
       document,
       {}, // rootValue
       {
+        ...this.context,
         pubsub: sc,
         lruGatewayResolvers: this.lruGatewayResolvers,
         reply: {
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -20,7 +20,8 @@ module.exports = class SubscriptionConnection {
     subscriber,
     fastify,
     lruGatewayResolvers,
-    entityResolvers
+    entityResolvers,
+    context
   }) {
     this.fastify = fastify
     this.socket = socket
@@ -29,6 +30,7 @@ module.exports = class SubscriptionConnection {
     this.entityResolvers = entityResolvers
     this.subscriber = subscriber
     this.subscriptionContexts = new Map()
+    this.context = context
 
     this.socket.on('message', (message) => {
       this.handleMessage(message).catch(e => {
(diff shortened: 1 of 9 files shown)


lib/subscription-connection.js
```diff
@@ -79,7 +81,8 @@ module.exports = class SubscriptionConnection {
 
     const sc = new SubscriptionContext({
       fastify: this.fastify,
-      pubsub: this.subscriber
+      pubsub: this.subscriber,
+      context: this.context
```
frikille, line 85: I think the context shouldn't be added here, but to the actual GraphQL context object (line 95). If we add it here the context object can be accessed in the subscriber function via the `pubsub` object (`context.pubsub.context`). If it's spread on the GraphQL context object it seems better I think (also in line with your original example in #151 )

bvanneerven: I agree! Implemented this in https://github.com/mcollina/fastify-gql/pull/153/commits/c7d4cc846181ed5711b412267926ec4bcb162091

frikille:
This looks good to me!

Could you update the Readme (and the type definition) with this new option?

lib/subscription.js
```diff
@@ -21,12 +21,18 @@ function createConnectionHandler ({ schema, subscriber, fastify, lruGatewayResol
       return
     }
 
+    let context = {}
+    if (subscriptionContextFn) {
+      context = await subscriptionContextFn(connection, request)
```
mcollina, line 26: I would prefer if these would properties a single object instead.

Also, could you add a test where this function throws an error? Thanks

katsanva:
@mcollina Hey. Is there something that prevents this PR to be merged?

mcollina:
1. it's still flagged as "draft", so I thought @bvanneerven  were still working on it
2. this comment must be addressed https://github.com/mcollina/fastify-gql/pull/153#pullrequestreview-406968633.

bvanneerven:
@mcollina Yes, I am still interested in pursuing this. I asked some questions a while ago based on your review, because I didn't fully understand it. Could you take a look at those?

mcollina:
Can you link the comment? I don't seem to be able to find it.

bvanneerven:
@mcollina I cannot really link the comment itself, but the file is here: https://github.com/mcollina/fastify-gql/pull/153/files/a712d11fdae7e81464e0635ff85ca8fc4eae489d#diff-49234199fc0c7c90d47d4a9524c21f0b

mcollina:
I'm lost. What are the questions?

bvanneerven:
@mcollina Sorry it has been a while, haven't been able to create time for this recently because of uni...
Anyways, here is a screenshot of the comment I see on my screen:

Screenshot_2020-07-08 Add subscription context by bvanneerven · Pull Request #153 · mcollina fastify-gql


Also, could you guide me in creating the failing test? I'm not so sure where and how is the best place to implement it.

mcollina:
I'm sorry that tests are failing on master. I'll ask to rebase as soon as master is fixed, sorry for the incovenience.

mcollina:
I have reverted the commit on master.

PR #156, opened 2020-05-07
fix: add @extends federation directive

katsanva:
Hello. Adding the missing directive that will allow to use `type U @extends { ... }` alongside with `extend U { ... }`

Please see the [docs](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/#create-stub-types):

> Some libraries such as `graphql-java` don't have native support for type extensions in their printer. Apollo Federation supports using an `@extends` directive in place of extend type to annotate type references:

```
  id: ID! @external
  reviews: [Review]
}
```

Without this, `fastify-gql` will fail with something like
```
"msg":"Unknown directive \"extends\".",
"stack":"Error: Unknown directive \"extends\".\n
    at assertValidSDLExtension (/node_modules/fastify-gql/node_modules/graphql/validation/validate.js:125:11)
    at extendSchema (/node_modules/fastify-gql/node_modules/graphql/utilities/extendSchema.js:71:43)
    at buildFederationSchema (/node_modules/fastify-gql/lib/federation.js:231:22)
    at buildGateway (/node_modules/fastify-gql/lib/gateway.js:141:18)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async fp.name (/node_modules/fastify-gql/index.js:118:15)"
```

### DIFF
--- a/lib/federation.js
+++ b/lib/federation.js
@@ -43,6 +43,7 @@ const BASE_FEDERATION_TYPES = `
   directive @requires(fields: _FieldSet!) on FIELD_DEFINITION
   directive @provides(fields: _FieldSet!) on FIELD_DEFINITION
   directive @key(fields: _FieldSet!) on OBJECT | INTERFACE
+  directive @extends on OBJECT | INTERFACE
 `
 
 const FEDERATION_SCHEMA = `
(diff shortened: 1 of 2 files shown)


mcollina:
Would you mind to add a test that demonstrate how this works?

katsanva:
That may take some time, because I haven’t dig deep into how it may be implemented in place...

frikille:
I had a look at this issue: adding the directive will only resolve the issue of throwing an error because of an unknown directive, but the gateway will create an incorrect schema because it would handle a type with the `@extends` directive as a type definition node instead of a type extension node.

PR #165, opened 2020-05-13
Fixes missing error extensions when query is jit'ed

### LINKED ISSUE #164: Error extensions not available when query is jit'ed
While using `ErrorWithProps`, the extensions is not available when query is not jit'ed.

This happens because `graphql-jit` [expects extensions property inside the originalError](https://github.com/zalando-incubator/graphql-jit/blob/d2048c8c37e43f04a1589f6eba5fa0c8935f7d09/src/error.ts#L18).

A simple fix would be to not have explicit `code` field in `ErrorWithProps` and rename `additionalProperties` field to `extensions` and expect users to pass the same inside the `extensions` field itself.

tHBp:
Fixes #164 

Removes explicit `code` field in `ErrorWithProps` and renames `additionalProperties` field to `extensions`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -555,7 +555,7 @@ const resolvers = {
     findUser: (_, { id }) => {
       const user = users[id]
       if (user) return users[id]
-      else throw new ErrorWithProps('Invalid User ID', "USER_ID_INVALID", { id, timestamp: Math.round(new Date().getTime()/1000) })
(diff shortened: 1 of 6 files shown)


mcollina:
@frikille can you take a look?

tHBp:
Sure, will do.

PR #167, opened 2020-05-15
feat: add subscription.onConnect option to authorize incoming connection

### LINKED ISSUE #166: Subscription client authentication
What is the recommended authentication flow for subscription ? 
The documentation suggest:
```js
verifyClient: (info, next) => {
      if (info.req.headers['x-fastify-header'] !== 'fastify is awesome !') {
        return next(false) // the connection is not allowed
      }
      next(true) // the connection is allowed
    }
```

But it's not possible to add custom headers to the websocket http request from a browser (see: https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api/41521871 and https://github.com/apollographql/apollo-client/issues/3967). So basically it's not possible to authenticate an incoming connection from a browser with verifyClient except with a cookie header.

Apollo uses a onConnect method to validate the `connection_init` payload send by client that contains the authentication infos.

PacoDu:
Fixes #166


PacoDu:
> Great work! I would make the `onConnect` function async/return a promise. Authentication might require go to a DB/external service after all.

I'm not sure how we should handle a promise rejection in this case.

lib/subscription-connection.js
```diff
@@ -52,7 +55,7 @@ module.exports = class SubscriptionConnection {
 
     switch (type) {
       case GQL_CONNECTION_INIT:
-        this.sendMessage(GQL_CONNECTION_ACK)
+        this.handleConnectionInit(data)
```
mcollina, line 58:
```suggestion
        await this.handleConnectionInit(data)
```

lib/subscription-connection.js
```diff
@@ -72,6 +75,23 @@ module.exports = class SubscriptionConnection {
...
+    if (typeof this.onConnect === 'function') {
+      let authorize = false
+      try {
+        authorize = await this.onConnect(data)
+      } catch (e) {
+        this.fastify.log.error(e)
+      }
+      if (!authorize) {
+        this.sendMessage(GQL_CONNECTION_ERROR, undefined, { message: 'Forbidden' })
+        this.handleConnectionClose()
```
mcollina, line 88:
```suggestion
        return this.handleConnectionClose()
```

lib/subscription-connection.js
mcollina: This might be harder than this to fix. Essentially it's possible to bypass the connectionInit hook by sending two messages without waiting for an ack. While we are doing connectionInit, we cannot process any more messages.

This feature is often called pipelining, and I'm not sure GraphQL subscriptions even supports that. A better fix would be to let the logic process only once message at a time.

PacoDu: I tried to fix this and add some kind of pipelining inspired from on [Apollo implementation](https://github.com/apollographql/subscriptions-transport-ws/blob/52758bfba6190169a28078ecbafd2e457a2ff7a8/src/server.ts). 
See: https://github.com/PacoDu/fastify-gql/commit/df59562a02b7d8033b267a9d8dc31e07ef8b59e0 

But tests are inconsistent (sometime pass, sometime fail), I guess this is related to promises that are not properly handled, it would be nice to have your inputs on this one ! Also I think that the current version will allow `start` payload even if `connection_init` hasn't been sent.

mcollina: Yes, exactly. We need to stregthen up a bit the current implementation otherwise it would be fairly easy to overcome any auth check in `onConnect`.

I would recommend to turn the `socket` into a stream, and the use async iteration to process all messages one by one.

PacoDu: Hi, sorry I didn't had much time to work on this last week. I'm not sure how to address this, could you please give me some details/leads on the changes it will imply for the message handler to turn the socket into a stream ? 

I guess I've to use this https://www.npmjs.com/package/ws#use-the-nodejs-streams-api but what about the message handler ? (I'm not confortable with stream yet - but I'm ready to try and learn)

mcollina: You can do something like:

```js
const stream = WebSocket.createWebSocketStream(ws, { encoding: 'utf8' });
for await (let msg of stream) {
  await handleMesasge(mg)
}
```

PacoDu: 😅ok pretty simple, thanks !

PacoDu: I'm unable to catch any message from this loop. I'm passing the connection from the wsHandler to the SubscriptionConnection class and then create the stream and the async iteration loop in a `consume` method called from the wsHandler after the `new SubscriptionConnection`. I removed the `on('data')` handlers but the loop is never triggered. I'll try to find some times to investigate further next weekend but I think I'm missing something here, and I can't find what it is for now ...

mcollina: Can you push the broken code somewhere?

PacoDu: https://github.com/PacoDu/fastify-gql/commit/fc8b823eb2b93dffb4d7a182685e99e6060c959d

PacoDu: Do you have any idea of what I'm doing wrong here ? PacoDu@fc8b823

The issue here is that the `consume` method (https://github.com/PacoDu/fastify-gql/commit/fc8b823eb2b93dffb4d7a182685e99e6060c959d#diff-d0a65557fba21a77676fdc6c8ab721c7R47) is not receiving any message, it's called from the wsHandler here: https://github.com/PacoDu/fastify-gql/commit/fc8b823eb2b93dffb4d7a182685e99e6060c959d#diff-49234199fc0c7c90d47d4a9524c21f0bR41

ALITHEALI:
Hey everyone! Is there any update on this and can we help in any shape or form? Many thanks for all your great works.

mcollina:
I think this has stalled. Would you like to pick it up? Essentially this PR is still receiving messages while it's processing `onConnect`.

PacoDu:
Hey, yes sorry I don't have time right now to work on this PR

ALITHEALI:
Hey! Sure. I'll take a look at the PR very soon. Are there any other problems besides the messages?

mcollina:
that is the main one.

lib/subscription-connection.js
PacoDu: Resolved by #228

PacoDu:
I think this PR is ready once #228 is merged

mcollina:
#228 landed.

PR #174, opened 2020-05-27
fix: prevent `cached.jit.query is not a function`

katsanva:
`compileQuery` from `graphql-jit` returns `CompiledQuery | ExecutionResult` by signature. 

Unfortunately, in case of some error, it will return `ExecutionResult` instead of `CompiledQuery`.

That will simply throw an error to the client about  `cached.jit.query is not a function` instead of skipping jit.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,7 @@ const fp = require('fastify-plugin')
 const LRU = require('tiny-lru')
 const routes = require('./lib/routes')
 const { BadRequest, MethodNotAllowed, InternalServerError } = require('http-errors')
-const { compileQuery } = require('graphql-jit')
+const { compileQuery, isCompiledQuery } = require('graphql-jit')
 const { Factory } = require('single-user-cache')
 const {
   parse,
@@ -358,7 +358,7 @@ const plugin = fp(async function (app, opts) {
       cached.jit = compileQuery(schema, document, operationName)
     }
 
-    if (cached && cached.jit !== null) {
+    if (cached && cached.jit !== null && isCompiledQuery(cached.jit)) {
       const res = await cached.jit.query(root, context, variables || {})
 
       // as `graphql-jit` takes the original response that's dispatched


katsanva:
@mcollina the `isCompiledQuery` is already well-tested in `graphql-jit` repo. Also, it's a type guard, so its semantically right to use it here. 

Unfortunately, I'm out of the scope of the _fastify way_ of testing, so if you could point me how I can make the test for `fastify-gql` quickly, I would appreciate it.

mcollina:
I would recommend to duplicate https://github.com/mcollina/fastify-gql/blob/master/test/errors.js#L9-L66 and send a query that would cause this problem.

asciidisco:
@katsanva I'd also be happy to contribute a testcase, but I'm not sure how I can provoke this behaviour. Do you have an example I can use to build the test upon?

katsanva:
@asciidisco I've encountered this behavior while using different versions of graphql package. By browsing the graphql-git package compileQuery will return errors in a [few different conditions](https://github.com/zalando-incubator/graphql-jit/blob/50855a2540502ee78dd034bc55853c255a5500b5/src/execution.ts#L1440). Also, it may be possible to get an error  by looking through both [the code](https://github.com/zalando-incubator/graphql-jit/blob/50855a2540502ee78dd034bc55853c255a5500b5/src/execution.ts#L184) and [the tests](https://github.com/zalando-incubator/graphql-jit/blob/50855a2540502ee78dd034bc55853c255a5500b5/src/__tests__/execution.test.ts)

mcollina:
@asciidisco @katsanva could any of you send a fresh PR with tests?

katsanva:
Sorry, I don't know how to reproduce the case within the current test implementation in the project. This was caused by different graphql versions used in the dependency tree, and as far that was resolved I haven't encountered this issue. 

AFAIK there is a pretty same PR implemented and which is newer then this one.

mohammed-gehad:
i have faced this issue 
`{
  "data": null,
  "errors": [
    {
      "message": "cached.jit.query is not a function"
    }
  ]
}`

also isCompiledQuery(cached.jit) fixed it.

it only happens when jit is enabled and i query with input. queries without input pass.

this query throws cached.jit.query is not a function.
`query findManySpecialty($where: SpecialtyWhereInput) {
  findManySpecialty(where: $where) {
    id
  }
}
`

this query work fine.
`query findManySpecialty {
  findManySpecialty {
    id
  }
}
`

PR #186, opened 2020-06-16
Add Support For Batched Queries

### LINKED ISSUE #185: Support Batched Queries
I'd like to submit a PR to implement this. First I wanted to raise the issue and see if you'd be willing to accept a PR or if you had any feedback.

---

We're currently migrating from Apollo Server to Fastify GQL because of the performance benefits.

Our mobile app makes use of [apollo-link-batch-http](https://www.apollographql.com/docs/link/links/batch-http/) for performance. This batches GraphQL queries which are run within a configurable interval. The request issued is an array of query objects, rather than a singular query object.

In order to support this, the graphql server must be able to merge the query objects and variables prior to the executing the query as per usual.

In order to support this in Fastify GQL:
- [x] Update schema to allow array or object.
- [x] Update handler to merge GraphQL queries if an array is provided.
- [x] Test with batched queries
- [x] Test with normal queries

pscarey:
Fix #185

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -224,9 +224,9 @@ module.exports = async function (app, opts) {
   }, async function (request, reply) {
     validationHandler(request.validationError)
 
-    if (Array.isArray(request.body) && allowBatchedQueries) {
+    if (allowBatchedQueries && Array.isArray(request.body)) {
       // Batched query
-      return Promise.all(request.body.map(async r =>
+      return Promise.all(request.body.map(r =>
         execute(r, request, reply)
           .catch(e => ({ errors: [{ message: e.message }] }))
       ))
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -7,48 +7,92 @@ const { formatError, GraphQLError } = require('graphql')
 const subscription = require('./subscription')
 const { FEDERATED_ERROR, toGraphQLError } = require('./errors')
 
-const responseSchema = {
-  '2xx': {
-    type: 'object',
-    properties: {
-      data: {
-        type: ['object', 'null'],
-        additionalProperties: true
-      },
-      errors: {
-        type: 'array',
-        items: {
-          type: 'object',
-          required: ['message'],
-          properties: {
-            message: { type: 'string' },
-            locations: {
-              type: 'array',
-              items: {
-                type: 'object',
-                properties: {
-                  line: { type: 'integer' },
-                  column: { type: 'integer' }
-                }
-              }
-            },
-            path: {
-              type: 'array',
-              items: { type: 'string' }
-            },
-            extensions: {
-              type: 'object',
-              properties: {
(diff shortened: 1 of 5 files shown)


mcollina:
Would you mind to add some documentation for batched queries?

mcollina:
code looks great.

pscarey:
Sure will do

pscarey:
Readme has been updated - hope that's OK.

Thanks.

lib/routes.js
```diff
@@ -190,55 +219,21 @@ module.exports = async function (app, opts) {
...
-    let { query } = request.body
-    const {
-      persisted,
-      variables,
-      operationName
-    } = request.body
-
-    if (persisted || onlyPersisted) {
-      query = persistedQueries[query]
+    if (Array.isArray(request.body) && allowBatchedQueries) {
```
mcollina, line 227:
```suggestion
    if (allowBatchedQueries && Array.isArray(request.body)) {
```

lib/routes.js
```diff
@@ -7,49 +7,93 @@ const { formatError, GraphQLError } = require('graphql')
...
+  body: {
+    type: allowBatchedQueries ? ['object', 'array'] : 'object',
+    properties: {
+      ...requestProperties,
+      variables: {
+        type: ['object', 'null']
+      }
+    }
+  },
+  // JSON schema isn't allowing ['object', 'array'] on response.
```
mcollina, line 88: Have you tried using an `anyOf`? I would prefer to keep the speedup done by the schema.

lib/routes.js
```diff
@@ -190,55 +219,21 @@ module.exports = async function (app, opts) {
...
-      persisted,
-      variables,
-      operationName
-    } = request.body
-
-    if (persisted || onlyPersisted) {
-      query = persistedQueries[query]
+    if (Array.isArray(request.body) && allowBatchedQueries) {
+      // Batched query
+      return Promise.all(request.body.map(async r =>
```
mcollina, line 229:
```suggestion
      return Promise.all(request.body.map(r =>
```

This does not need to be async.

lib/routes.js
pscarey: I've had a decent look into this after your suggestion, and I can't seem to make it work. Any combination of `anyOf` I tried I got errors compiling the schema.

If you have further suggestions as to where to look for a possible example or e.g. where in the `fast-json-stringify` codebase I might get a lead on what wall I'm hitting, I'm happy to look into it more - am not particularly familiar with JSON Schema though.

pscarey:
I've added the test & made the suggested changes. Thank you. Still stuck as mentioned above with the schema.

lib/routes.js
mcollina: Let's land this as is then, we'll investigate in the future.

Issue #110, opened 2020-01-15
Support for Automatic Persisted Queries


guigaoliveira:
Some support for this feature would be interesting, which helps to improve network performance. But I don't know if there is any standardization to implement this.

About persisted queries:  "A persisted query is an ID or hash that can be sent to the server instead of the entire GraphQL query string. This smaller signature reduces bandwidth utilization and speeds up client loading times. Persisted queries are especially nice paired with GET requests, enabling the browser cache and integration with a CDN."

Interesting links: 
- [Automatic persisted queries](https://www.apollographql.com/docs/apollo-server/performance/apq/)
- [GraphQL Dynamic Persisted Queries](https://medium.com/open-graphql/graphql-dynamic-persisted-queries-eb259700f1d3)
- [Persisted Query support with Apollo Link](https://github.com/apollographql/apollo-link-persisted-queries)
- [Persisted queries in GraphQL: Slim down Apollo requests to your Ruby application](https://evilmartians.com/chronicles/persisted-queries-in-graphql-slim-down-apollo-requests-to-your-ruby-application) 
- [Relay Persisted Queries](https://relay.dev/docs/en/persisted-queries)

mcollina:
This would be handy! What the developer workflow look like?

ericlewis:
@mcollina from my knowledge there isn't really a set-in-stone workflow around this- it differs wildly depending on server implementation and client implementation. 

The primary idea behind it is that: at build time of the client application you "compile" or "hash" the given graphql query and then upload or store that hash somewhere else. Then, instead of sending up an entire query, the client will send an ID and if a mutation, also some data. the graphql server knows enough to match this ID with the appropriate query then runs it as usual.

The primary reasoning for this is when you have many routes, and sending up those query blobs becomes expensive. It is also good for fairly consistent latency on mobile.

mcollina:
What is the best way to set those from your point of view? A JSON file with the query/sha combination?

salmanm:
> at build time of the client application you "compile" or "hash" the given graphql query and then upload or store that hash somewhere else.

@ericlewis Generating a query+hash map at build time comes at the cost of maintaining a static file (or even a shares space like redis) and keeping it in sync between client and server. While it saves a little bit of processing at run-time, I think the overhead of keeping the client/server in sync isn't a good trade-off.

@mcollina @frikille 
[...]

PR #147, opened 2020-05-04
Update fastify to v3

frikille:
Update dependencies
- [x] fastify to v3
- [x] fastify-plugin to v2
- [x] fastify-static to v2
- [x] fastify-websocket waiting to merge and release fastify/fastify-websocket#54

This PR also adds dependabot config and moves CI from Travis to Github Actions

PR #187, opened 2020-06-16
Add Support For Automatic Persisted Queries

pscarey:
Adds customisable support for Automatic Persisted Queries, with default settings compatible with `apollo-client`.

#110 already has discussion on this. We required support internally, so it's up to maintainers if support in this way is appropriate. Happy to make tweaks based on feedback if required.

Note this builds on #186.


mcollina:
I'm not convinced by this implementation. How do you plan to scale this horizontally? The queries will be stored only on one peer.

pscarey:
Interesting point - I think this is how Apollo Server implemented it, which is why I went down this road, then generalised the way the query is delivered.

The argument in favour would be that the cost of hashing a query the first time an instance sees it will still massively reduce the number of times the full query is sent over the wire.

---

Really, this entire interface or setting might be better off merged into the existing persisted queries solution, in a customisable manner, since there is no firm 'Best Practice', such that there is:

Persisted Query Settings:
-  `isPersistedQuery` check if the given GraphQL request/query is persisted or regular.
- `getHash` get the hash from a GraphQL persisted query request, or falsy if not found.
- `getQueryFromHash` return the query for a given hash, or falsy if not found.
- `getHashForQuery` return the hash for a given query, or falsy if not to be stored.
- `saveQuery` store the query for the given hash.
- `onlyPersisted` whether to allow persisted queries or not.

Example implementations would be:

Current Persisted Queries (Ahead of time):
```
{
  isPersistedQuery: (r) => r.persisted,
  getHash: (r) => r.query,
  getQueryFromHash: (hash) => persistedQueries[hash], // Load from memory
  // getHashForQuery and saveQuery are not required as we do not save new queries
}
```

Apollo Client Compatible Automatic Persisted Queries (Runtime):
```
{
  isPersistedQuery: (r) => r.extensions.persistedQuery.sha256Hash && r.extensions.persistedQuery.version === 1,
  getHash: (r) => r.extensions.persistedQuery.sha256Hash ,
  getQueryFromHash: (hash) => persistedQueries[hash], // Load from memory
  getHashForQuery: (query) => sha256(query),
  saveQuery: (hash, query) => persistedQueries[hash] = query // Store in memory
}
```

Cached Apollo Client Compatible Automatic Persisted Queries (Runtime + Shared Cache):
```
{
  isPersistedQuery: (r) => r.extensions.persistedQuery.sha256Hash && r.extensions.persistedQuery.version === 1,
  getHash: (r) => r.extensions.persistedQuery.sha256Hash ,
  getQueryFromHash: (hash) => redis.get(hash), // Load from redis
  getHashForQuery: (query) => sha256(query),
  saveQuery: (hash, query) => redis.set(hash, query) // Store in redis
}
```

mcollina:
My point is that the current interface (ahead of time) cannot be made dynamic without adding support for multiple instances of the server that store the query. Otherwise subsequent requests that hit a different server would not have the cached query.

In the code in this PR the “get” operation is synchronous and it should be made async (and likely moved in a common interface). In your list, the Persisted Query Settings:
-  `isPersistedQuery` check if the given GraphQL request/query is persisted or regular.
- `getHash` get the hash from a GraphQL persisted query request, or falsy if not found.
- `getQueryFromHash` return the query for a given hash, or falsy if not found. __must be async__
- `getHashForQuery` return the hash for a given query, or falsy if not to be stored.
- `saveQuery` store the query for the given hash. __must be async__
- `onlyPersisted` whether to allow persisted queries or not.

---

I’m also skeptical that adding a delay of a round trip to Redis is going to improve latency much in practice. We are trading the cost of sending a few hundred bytes from the client to fetching those from Redis. Given how HTTP/1.1 keepalive and TCP work, sending a few hundreds bytes more should have no impact.
The risk is to have an ahead-of-line blocking on the redis connection which will make thus caching moot.

Notably this makes a lot of sense if our query is massive.

pscarey:
I agree with your reasoning that it'd need to be async.

As a general rule, any persisted query requires that the payload reduction savings are greater than the lookup delay. How big the lookup delay is depends on the implementation.

My understanding is that:

Ahead of Time Persisted Queries (AoTPQ):
- Best performance in every use case.
- Hardest to set up, as all queries need to be compiled into hashes ahead of time.

Automated Persisted Queries (APQ):
- Using local memory will be slow for each of their first query, but then give equal performance compared to AoTPQ afterwards.
- The performance of this is basically equivalent to AoTPQ for long lived servers (i.e. not cloud functions).
- This is also the easiest to set up.

Cached APQ:
- Remove that first overhead at the cost of a roundtrip to the cache on every request.
- The performance for this would be worse than APQ for long lived servers, but better than APQ for short lived servers (cloud functions).
- The performance of this is heavily caveated on the payload reduction being sufficiently large to counteract the cost of the lookup time. However, it can reasonably be assumed than an in-datacenter Redis connection is much faster than e.g. a 3G mobile connection. Really this means it boils down to the use case, and the performance of the client.

---

Appreciate the responsiveness here, but I also want to flag that in order for us to migrate to fastify-gql, we need to support APQ as otherwise we'll lose backwards compatibility - obviously that doesn't mean that it's something fastify-gql should support - so if it's a no-go, we can fork.

mcollina:
Could you confirm that you are using "sticky sessions" for APQ? How can you guarantee that the same client goes to the same server otherwise? What i do not understand about APQ is if it's actually feasible in production. It seems very likely that a client talking to a different server might not get the cached one.

> I agree with your reasoning that it'd need to be async.

Then we are good, can you update the PR?

---

> Appreciate the responsiveness here, but I also want to flag that in order for us to migrate to fastify-gql, we need to support APQ as otherwise we'll lose backwards compatibility - obviously that doesn't mean that it's something fastify-gql should support - so if it's a no-go, we can fork.

I'm cautious about adding feature, I think this is legit and we should add it in one form or another.

pscarey:
I will do updates to both PR's tomorrow.

> Could you confirm that you are using "sticky sessions" for APQ? How can you guarantee that the same client goes to the same server otherwise? What i do not understand about APQ is if it's actually feasible in production. It seems very likely that a client talking to a different server might not get the cached one.

Our use case is as follows:

**Client:**
- Mobile App running `apollo-client` with approx 20 unique queries.
- Prior versions of the mobile app are supported, but we haven't tracked legacy queries (i.e. they are no longer in our master branch, but are still valid queries due to no breaking changes).
- `apollo-client` attempts to issue a query using the persisted queries hash. Expected API responses are as per [apollo-link-persisted-query](https://github.com/apollographql/apollo-link-persisted-queries#protocol).
- If the client receives the appropriate `PersistedQueryNotFound` response, then it resends the whole query. This causes the APQ cache to be populated for future requests.

**Server:**
- 3-4 replicas of our GraphQL API running in K8s.
- Queries are persisted when they are received in full, in local memory.
- After a short amount of usage, the most common queries will all have been persisted.
- After a little while longer, less frequently called queries are persisted.
- The machine continues its lifetime for a long period after, directly replying to persisted queries.

I hope this (very rough) diagram can covey the idea slightly better:
Screen Shot 2020-06-18 at 10 01 22 pm


Sticky sessions would ensure that the cache of the machine returning `PersistedQueryNotFound` is the one populated by the query, but are not strictly necessary, as the queries will still be persisted in a short period of time compared to the lifetime of the machine.

mcollina:
Thanks, it makes sense now. Please add a short version of that in the README.

pscarey:
I've reworked the API & tidied things up a lot based on the conversation above.

- The proposed API gives defaults which align with the existing options, support for Apollo's APQ, and room for additional customisation.
- Documentation and types are updated.
- All persisted query tests have been moved into a seperate file.

mcollina:
Can you rebase this on top of `master`?

pscarey:
Have merged the other changes in, diff should be clean.

Thanks.

README.md
```diff
@@ -92,51 +93,74 @@ app.get('/', async function (req, reply) {
...
 ```
 
-### Disallowing unknown queries
+#### PreparedOnly
+
+This offers similar performance and considerations to the `Prepared` queries, but only allows persisted queries. This provides additional secuirity benefits, but means that the server **must** know all queries ahead of time or will reject the request.
+
+The API is the same as the `Prepared` default.
+
+#### ApolloAutomatic
```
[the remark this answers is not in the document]
pscarey: Sure

README.md
```diff
@@ -648,8 +672,14 @@ __fastify-gql__ supports the following options:
     * `service.rewriteHeaders`: `Function` A function that gets the original headers as a parameter and returns an object containing values that should be added to the headers
     * `service.wsUrl`: The url of the websocket endpoint
     * `service.wsConnectionParams`: `Function` or `Object`
-* `persistedQueries`: A hash/query map to resolve the full query text using it's unique hash.
-* `onlyPersisted`: Boolean. Flag to control whether to allow graphql queries other than persisted. When `true`, it'll make the server reject any queries that are not present in the `persistedQueries` option above. It will also disable any ide available (playground/graphiql).
```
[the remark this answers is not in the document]
pscarey: OK

README.md
```diff
@@ -92,51 +93,74 @@ app.get('/', async function (req, reply) {
...
+
+It is also possible to extend or modify these persisted query implementations for custom cases, such as Automatic Persisted Queries, but with a shared cache between servers.
+
+This would enable all persisted queries to be shared between all server instances in a cache which is dynamically populated. The lookup time from the cache is an additional overhead for each request, but a higher rate of persisted query matches would be achieved. This may be beneficial, for example, in a public facing API which supports persisted queries and uses cloud functions (short lived server instances). *Note the performance impacts of this need to be considered thoroughly: the latency added to each request must be less than the savings from smaller requests.*
+
+A example of using this with Redis would be:
+
+```js
+const persistedQuerySettings = {
+  ...GQL.PersistedQueryDefaults.ApolloAutomatic(),
```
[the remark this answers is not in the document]
pscarey: Will add

index.js
```diff
@@ -32,6 +33,35 @@ const { ErrorWithProps, FEDERATED_ERROR } = require('./lib/errors')
...
+        return version === 1 ? sha256Hash : false
+      },
+      getQueryFromHash: async (hash) => AUTOMATIC_PERSISTED_QUERIES[hash],
+      getHashForQuery: (query) => crypto.createHash('sha256').update(query, 'utf8').digest('hex'),
+      saveQuery: async (hash, query) => { AUTOMATIC_PERSISTED_QUERIES[hash] = query },
+      notFoundError: 'PersistedQueryNotFound',
+      notSupportedError: 'PersistedQueryNotSupported'
+    })
+  }
+}
```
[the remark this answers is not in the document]
pscarey: Sure

lib/routes.js
```diff
@@ -145,29 +148,85 @@ module.exports = async function (app, opts) {
...
+          // as we do not recognise them
+          throw new BadRequest(notSupportedError)
+        }
+      }
     }
 
+    // Validate a query is present
     if (!query) {
-      return new BadRequest('Unknown query')
+      return new BadRequest('QueryNotFound')
```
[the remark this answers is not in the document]
pscarey: No particular reason, have reverted.

pscarey:
I've:
- Updated the naming to Automatic, and clarified the imports in the README.
- Moved the `PersistedQueryDefaults` into a new file.
- Reverted the change of the naming of the query not found error back to `Unknown query`.
- Added support for `persistedQueries` and `onlyPersisted`, to automatically map to the `persistedQuerySettings` internally.
- Added tests for both using `persistedQuerySettings` and `persistedQueries`/`onlyPersisted`.

README.md
```diff
@@ -92,51 +93,101 @@ app.get('/', async function (req, reply) {
...
+    '<hash>': '{ add(x: 1, y: 1) }'
+  },
+  onlyPersisted: true
+})
 ```
 
-### Disallowing unknown queries
+#### Automatic
+
+This default is compatible with `apollo-client`, and requires no additional tooling to set up at the cost of some performance. In order for this mode to be effective, you must have long lived server instances (i.e *not* cloud functions). This mode is also appropriate for public APIs where queries are not known ahead of time.
```
mcollina, line 163: Adding some links to the relevant portion of apollo-client docs would be really handy, thanks.

README.md
```diff
@@ -92,51 +93,101 @@ app.get('/', async function (req, reply) {
...
+This would enable all persisted queries to be shared between all server instances in a cache which is dynamically populated. The lookup time from the cache is an additional overhead for each request, but a higher rate of persisted query matches would be achieved. This may be beneficial, for example, in a public facing API which supports persisted queries and uses cloud functions (short lived server instances). *Note the performance impacts of this need to be considered thoroughly: the latency added to each request must be less than the savings from smaller requests.*
 
-Note: `onlyPersisted` disables all IDEs (graphiql/playground) so typically you'd want to use it in production.
+A example of using this with Redis would be:
 
+```js
+const persistedQuerySettings = {
+  ...GQL.PersistedQueryDefaults.Automatic(),
```
mcollina, line 181: Can you add a require for GQL? Thanks

lib/persistedQueryDefaults.js
```diff
@@ -0,0 +1,32 @@
+const crypto = require('crypto')
+
+const PersistedQueryDefaults = {
+  Prepared: (persistedQueries) => ({
```
mcollina, line 4: I prefer to keep naming of those lowercase

lib/persistedQueryDefaults.js
```diff
@@ -0,0 +1,32 @@
+const crypto = require('crypto')
```
mcollina, line 1: `'use strict'` is missing at the top

README.md
```diff
@@ -649,8 +700,16 @@ __fastify-gql__ supports the following options:
     * `service.rewriteHeaders`: `Function` A function that gets the original headers as a parameter and returns an object containing values that should be added to the headers
     * `service.wsUrl`: The url of the websocket endpoint
     * `service.wsConnectionParams`: `Function` or `Object`
-* `persistedQueries`: A hash/query map to resolve the full query text using it's unique hash.
-* `onlyPersisted`: Boolean. Flag to control whether to allow graphql queries other than persisted. When `true`, it'll make the server reject any queries that are not present in the `persistedQueries` option above. It will also disable any ide available (playground/graphiql).
+* `persistedQueries`: A hash/query map to resolve the full query text using it's unique hash. Overrides `persistedQuerySettings`.
+* `onlyPersisted`: Boolean. Flag to control whether to allow graphql queries other than persisted. When `true`, it'll make the server reject any queries that are not present in the `persistedQueries` option above. It will also disable any ide available (playground/graphiql). Requires `persistedQueries` to be set, and overrides `persistedQuerySettings`.
+* `persistedQuerySettings`
```
mcollina, line 705: Maybe `persistedQueryStorage` is a better name? those are functions.

lib/persistedQueryDefaults.js
```diff
@@ -0,0 +1,32 @@
...
+    getQueryFromHash: async (hash) => persistedQueries[hash]
+  }),
+  PreparedOnly: (persistedQueries) => ({
+    isPersistedQuery: (r) => true,
+    getHash: (r) => r.persisted ? r.query : false, // Only support persisted queries
+    getQueryFromHash: async (hash) => persistedQueries[hash]
+  }),
+  Automatic: () => {
+    // Initialize only in the scope of this server instance
+    const AUTOMATIC_PERSISTED_QUERIES = {}
```
mcollina, line 16: This must be a very large LRU cache instead, otherwise we are exposing ourselves to a DoS attack by memory exhaustion. check out https://www.npmjs.com/package/mnemonist.

README.md
pscarey: Sure

README.md
pscarey: Sure

lib/persistedQueryDefaults.js
pscarey: No worries

lib/persistedQueryDefaults.js
pscarey: Thanks - too used to TS

README.md
pscarey: Whilst most members are functions:
- `isPersistedQuery` and `getHash` functions extract data from the request payload. 
- `getQueryFromHash`, `saveQuery` are related to storage.
- `getHashForQuery` option is about hashing a query.
- `notFoundError` and `notSupportedError` are strings.

I think based on this breakdown, the options/settings are broader than "storage". Happy to change if you still would like, but maybe another name is better?

lib/persistedQueryDefaults.js
pscarey: This is a good point, I'll look into implementing it.

pscarey:
Minor fixes have been completed. Don't mind the long process - happy to get the prompt feedback, and I appreciate it's a non trivial set of changes.

Should other casing also be updated?
- `Prepared` -> `prepared`
- `PreparedOnly` -> `preparedOnly`
- `Automatic` -> `automatic`

lib/persistedQueryDefaults.js
pscarey: Have added

README.md
```diff
@@ -92,51 +93,127 @@ app.get('/', async function (req, reply) {
...
+  })
+})
+```
 
-const schema = `
-  type Query {
-    add(x: Int, y: Int): Int
-  }
-`
+Alternatively the `peristedQueries` option may be used directly, which will be internally mapped to the `Prepared` default:
```
mcollina, line 130: Prepared -> prepared.

README.md
```diff
@@ -649,8 +726,16 @@ __fastify-gql__ supports the following options:
...
+* `persistedQuerySettings`
+  * `isPersistedQuery: (request: object) => boolean`: Return true if a given request matches the desired persisted query format.
+  * `getHash: (request: object) => string`: Return the hash from a given request, or falsy if this request format is not supported.
+  * `getQueryFromHash: (hash: string) => string`: Return the query for a given hash.
+  * `getHashForQuery?: (query: string) => string`: Return the hash for a given query string. Do not provide if you want to skip saving new queries.
+  * `saveQuery?: (hash: string, query: string) => Promise<void>`: Save a query, given its hash.
+  * `notFoundError?: string`: An error message to return when `getQueryFromHash` returns no result. Defaults to `Bad Request`.
+  * `notSupportedError?: string`: An error message to return when a query matches `isPersistedQuery`, but returns no valid hash from `getHash`. Defaults to `Bad Request`.
```
mcollina, line 738: Can you add `async` for the functions that should be async?

README.md
mcollina: `persistedQueryProvider` maybe?

index.d.ts
```diff
@@ -157,6 +208,15 @@ declare namespace fastifyGQL {
...
     extensions?: object
   }
+
+  /**
+   * Default options for persisted queries.
+   */
+  export const persistedQueryDefaults: { 
+    Prepared: (persistedQueries: object) => PeristedQuerySettings
+    PreparedOnly: (persistedQueries: object) => PeristedQuerySettings
+    Automatic: (maxSize?: number) => PeristedQuerySettings
```
mcollina, line 218: can you please use lower case here as well?

lib/persistedQueryDefaults.js
```diff
@@ -0,0 +1,36 @@
...
+        const { version, sha256Hash } = r.extensions.persistedQuery
+        return version === 1 ? sha256Hash : false
+      },
+      getQueryFromHash: async (hash) => cache.get(hash),
+      getHashForQuery: (query) => crypto.createHash('sha256').update(query, 'utf8').digest('hex'),
+      saveQuery: async (hash, query) => cache.set(hash, query),
+      notFoundError: 'PersistedQueryNotFound',
+      notSupportedError: 'PersistedQueryNotSupported'
+    })
+  }
```
mcollina, line 33: What is `r`? Can you please use descriptive variables names?

lib/routes.js
```diff
@@ -145,29 +148,85 @@ module.exports = async function (app, opts) {
...
+      persisted = isPersistedQuery(body)
+      if (persisted) {
+        // This is a peristed query, so we use the hash in the request
+        // to load the full query string.
+
+        // Extract the hash from the request
+        const hash = getHash && getHash(body)
+        if (hash) {
+          // Load the query for the provided hash
+          query = getQueryFromHash && await getQueryFromHash(hash)
```
mcollina, line 190: I think we should validate during plugin setup that if somebody is enabling persisted queries they should pass in a `getQueryFromHash` function.

lib/routes.js
```diff
@@ -145,29 +148,85 @@ module.exports = async function (app, opts) {
...
-    let context = {}
-    if (contextFn) {
-      context = await contextFn(request, reply)
+    // Handle the query, throwing an error if required
+    const result = await reply.graphql(query, { pubsub: subscriber, ...context, __currentQuery: query }, variables, operationName)
+
+    // Check for not yet persisted queries
+    if (enablePersistedQueries && !persisted) {
+      // If provided the getHashForQuery, saveQuery settings we save this query
+      const hash = getHashForQuery && getHashForQuery(query)
```
mcollina, line 217: We have already computed the hash, so we should reuse what we computed before.

mcollina:
@pscarey any updates on this? It would be great to land!

pscarey:
@mcollina Sorry about the delay - was away for a few days. Will be updating shortly.

lib/routes.js
pscarey: The process is broken down into two steps:
1. Retrieve the hash from the graphql request (i.e. accessing the appropriate property of the request object), and loading the appropriate query from the given hash.
2. If the query is not (yet) persisted (i.e. 1 has failed), compute the hash from the query string, and save it.

The part you've commented on is step 2, where we compute the hash. We don't have a computed hash at this stage, and we have been unable to find a valid hash in the query (as per the `!persisted` check).

pscarey:
All updated. FYI, we've been running this in production for over a week, and have had no issues.

lib/routes.js
```diff
@@ -145,29 +148,84 @@ module.exports = async function (app, opts) {
...
+    getQueryFromHash,
+    getHashForQuery,
+    saveQuery,
+    notFoundError,
+    notSupportedError
+  } = persistedQueryProvider || {}
+
+  const enablePersistedQueries = !!persistedQueryProvider
+
   async function execute (body, request, reply) {
```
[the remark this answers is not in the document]
pscarey: Have tidied this up - is substantially more readable, thanks.

pscarey:
Have implemented the feedback, but it looks like some changes from `master` are breaking CI?

mcollina:
Yes, we are fixing them today, unfortunately something broke in #147

mcollina:
I have the reverted the offending commit on master.

lib/routes.js
```diff
@@ -145,31 +148,95 @@ module.exports = async function (app, opts) {
...
+  }
+
+  async function execute (body, request, reply) {
+    if (persistedQueryProvider) {
+      return executePersistedQuery(body, request, reply)
+    } else {
+      const { query, operationName, variables } = body
+      return executeQuery(query, variables, operationName, request, reply)
+    }
+  }
```
mcollina, line 238: I would do something like:

```js

function executeQueryWrap (body, request, reply) {
      const { query, operationName, variables } = body
      return executeQuery(query, variables, operationName, request, reply)
}

const execute = persistedQueryProvider ? executePersistedQuery : executeQueryWrap
```

pscarey: Thanks, have updated

PR #197, opened 2020-07-08
Error Formatter Option

### LINKED ISSUE #196: Error Formatting Improvements
The `errorHandler` option controls the Fastify App Error handler directly, meaning:

- Users must fully rewrite the formatting of GraphQL errors in order to do error formatting.
- Batched query errors can't be formatted, as the overall response is a success. Errors should be formatted for each individual query, and the current implementation just sends the message of the error.

I propose a new option, `errorFormatter`. The Fastify GQL plugin will then:

- Set the app `errorHandler` using the `errorHandler` option, if provided, or if not false, set the default error handler.
- The default error handler will use either the `errorFormatter` if provided, or the default error formatter.

The `errorFormatter` option will take the form:

```js
(err: Error | GraphQLError | FastifyError) => {
    statusCode: number
    response: {
      data: object
      errors: object[]
    }
 }
```

For regular queries, the returned `statusCode` will be set, and the response will be set accordingly.
For batched queries, the `statusCode` will be ignored (set to 200) and the individual responses will be set accordingly.

An example usage of this option is:

```js
const errorParser = (e: GraphQLError) => {
    const { originalError, locations, path } = e
    const error = originalError as PlatformError

    let message
    if (error.name === 'PlatformError') {
        // Custom error formatter for internal usage
        message = getMessageForError(error.code, error.metadata)
    }

    if (!message) {
        // Error monitoring and reporting for unrecognised errors
        captureError(error)
    }

    // Default to internal server error
    return {
        message: message || 'Internal Server Error',
        locations,
        path
    }
}

const errorFormatter = (error: GraphQLServerError) => {
    if (!error.data) {
        switch (error.message) {
            // Send persisted query errors back in a formatted fashion
            case 'PersistedQueryNotFound':
                return {
                    statusCode: 200,
                    response: { data: null, errors: [{ message: 'PersistedQueryNotFound' }] }
                }
            case 'PersistedQueryNotSupported':
                return {
                    statusCode: 500,
                    response: { data: null, errors: [{ message: 'PersistedQueryNotSupported' }] }
                }
            // Send middleware errors directly
            default:
                return {
                    statusCode: error.statusCode || 500,
                    response: error.message
                }
        }
    }

    // Format all other errors as a valid GraphQL response
    const errors = (error.errors || [error]).map(errorParser)
    const data = error.data || null

    return {
        statusCode: 200,
        response: { errors, data }
    }
}
```

pscarey:
Fix #196 

PR based on top of #187 (was required for internal usage and testing in production).


mcollina:
I shipped v5 updated to fastify v3. Can you rebase?

lib/routes.js
```diff
@@ -143,10 +130,25 @@ function validationHandler (validationError) {
...
 
 module.exports = async function (app, opts) {
+  const errorFormatter = typeof opts.errorFormatter === 'function' ? opts.errorFormatter : defaultErrorFormatter
+
   if (typeof opts.errorHandler === 'function') {
     app.setErrorHandler(opts.errorHandler)
   } else if (opts.errorHandler === true || opts.errorHandler === undefined) {
-    app.setErrorHandler(defaultErrorHandler)
+    app.setErrorHandler((error, _, reply) => {
+      /* istanbul ignore if */
```
pscarey, line 139: I'm not sure about the inclusion of this here.

I see why it was added to the error handler (and this is where I've added it back here), but the `errorFormatter` can now override the status code of an error.

mcollina: Let's remove it.

PR #206, opened 2020-07-09
Feature :: Refresh / Polling Graphql Schema inside Gateway mode

### LINKED ISSUE #205: [question] Refresh / Reload schema inside Gateway mode
Hi,

We want to use Federation + Gateway on our project, but we need to use polling or refresh schema on change.
Could you tell me, is it possible to do polling / refreshing schema inside Gateway mode?

Thank you.

iviaks:
This PR closes #205


iviaks:
@mcollina 
Let's make this feature awesome! :)

index.js
```diff
@@ -122,10 +122,22 @@ const plugin = fp(async function (app, opts) {
...
     entityResolversFactory = gateway.entityResolversFactory
 
+    if (gateway.pollingInterval !== undefined) {
+      setInterval(async () => {
+        try {
+          fastifyGraphQl.schema = await gateway.refresh()
+        } catch (err) {
+          console.error(err)
+        }
+      }, gateway.pollingInterval)
```
mcollina, line 132: this interval must be cancelled during the `app.onClose()` block down below.

index.js
```diff
@@ -122,10 +122,22 @@ const plugin = fp(async function (app, opts) {
     schema = gateway.schema
     entityResolversFactory = gateway.entityResolversFactory
 
+    if (gateway.pollingInterval !== undefined) {
+      setInterval(async () => {
+        try {
+          fastifyGraphQl.schema = await gateway.refresh()
```
mcollina, line 128: This needs to be a bit more precise and tell us _if_ there are new schemas. We have a few internal caches for the parsed queries that must be invalidated if the schema change. In case of a change, I would also drop all outstanding websocket connections.

index.js
```diff
@@ -122,10 +122,22 @@ const plugin = fp(async function (app, opts) {
...
+        } catch (err) {
+          console.error(err)
```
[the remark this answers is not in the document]
iviaks: I replaced this line with

```js
        } catch (error) {
          app.log.error(error)
        }
```

index.js
iviaks: I have added a condition into app.onClose

```js
if (gatewayInterval !== null) {
    clearInterval(gatewayInterval)
}
```

index.js
iviaks: I have replaced this line with

```js
fastifyGraphQl.replaceSchema(await gateway.refresh())
```

mcollina: This still need tests and logic to verify that when changing the schema the JIT cache is dropped and the websocket connections are closed, but only if the schema change and not every interval.

mcollina:
This needs tests.

iviaks:
I understand and I will do it asap

iviaks:
@mcollina 
I have added tests for this feature
I needed to install https://github.com/sinonjs/fake-timers for working with timers (We don't need to use real timers for it)
Could you check the latest changes?

Thank you.

iviaks:
Hi, @mcollina 

I wrote down an initial implementation of subscription test
That's weird, but subscription can't handle updates for a schema.

Gateway has an updated schema, user-service has an updated schema, even cache does not exist, but subscription can't find a new field.

Could you take a look at it? I hope, you will find my mistake

Thanks.

test/gateway/pollingInterval.js
```diff
@@ -488,3 +489,302 @@ test('Polling schemas (cache should be cleared)', async (t) => {
...
+              name
+            }
+          }
+        `
+      }
+    })
+  )
+
+  const promise = () =>
+    new Promise((resolve) => {
```
mcollina, line 624: please avoid the promise constructor. You can use `require('events').once()` to achieve the same result.

test/gateway/pollingInterval.js
```diff
@@ -488,3 +489,302 @@ test('Polling schemas (cache should be cleared)', async (t) => {
...
+            name: 'John'
+          })
+
+          resolve()
+        }
+      })
+    })
+
+  return promise()
+    .then(async () => {
```
mcollina, line 654: use await, remove `then`

test/gateway/pollingInterval.js
```diff
@@ -488,3 +489,302 @@ test('Polling schemas (cache should be cleared)', async (t) => {
...
+            triggerUser: Boolean
+          }
+
+          type User @key(fields: "id") {
+            id: ID!
+            name: String!
+            lastName: String
+          }
+        `)
+      )
```
mcollina, line 675: I would expect the websocket connection to error/disconnect after this happens, this is not tested.

test/gateway/pollingInterval.js
iviaks: Thank you. I replaced all promise constructors with `.once()`

test/gateway/pollingInterval.js
iviaks: I have added a few new checking for `ws.readyState`.
Why do you think that WebSocket should be disconnected?

It will be bad for the end-user (because he needs to refresh the page, reload the application, etc.)

test/gateway/pollingInterval.js
```diff
@@ -0,0 +1,797 @@
...
+      t.deepEqual(updatedUser, {
+        id: 'u1',
+        name: 'John'
+      })
+
+      eventEmitter.emit('client.on.data')
+    }
+  })
+
+  await once(eventEmitter, 'client.on.data')
```
mcollina, line 656: There is no need for a custom eventEmitter. Just do `once(client, 'data')`

iviaks: I have replaced custom eventEmitter with client, but in some cases, I needed to trigger it twice (connection and action)

lib/gateway.js
```diff
@@ -281,6 +281,22 @@ async function buildGateway (gatewayOpts, app) {
...
     entityResolversFactory: factory,
+    pollingInterval: gatewayOpts.pollingInterval,
+    async refresh () {
+      for (const service of Object.values(serviceMap)) {
+        try {
+          await service.refresh()
+        } catch (err) {
+          continue
+        }
+      }
```
mcollina, line 292: Can we refresh those in parallel with a `Promise.allSettled()`?

test/gateway/pollingInterval.js
```diff
@@ -0,0 +1,790 @@
...
+
+      t.deepEqual(updatedUser, {
+        id: 'u1',
+        name: 'John'
+      })
+    }
+  })
+
+  await once(client, 'data')
+  await once(client, 'data')
```
mcollina, line 653: You should be able to do `const [chunk] = await once(stream, 'data')`

test/gateway/pollingInterval.js
```diff
@@ -0,0 +1,790 @@
...
+          mutation {
+            triggerUser
+          }
+        `
+    }
+  })
+
+  await once(client, 'data')
+
+  t.equal(ws.readyState, 1)
```
mcollina, line 715: I think the first client should be closed now.

lib/gateway.js
```diff
@@ -281,6 +281,22 @@ async function buildGateway (gatewayOpts, app) {
...
+
+      const serviceSDLs = Object.values(serviceMap).map(service => service.schemaDefinition)
+      const schema = buildFederatedSchema(serviceSDLs.join(''), true)
```
mcollina, line 295: Can we cache this output and avoid replacing the schema if the `serviceSDLs.join('')` did not change?

test/gateway/pollingInterval.js
iviaks: I created a new client for getting a new subscription.
I can manually close ws.client inside tests, but we have it inside `t.tearDown()`.

If we compare it with an end-user perspective.

For example,
You're on the page with notifications. You're using an application and getting new notifications in real-time. Some developer updated GraphQL Type for notifications (he added one field).

We can close a WebSocket connection. In this case, end-user has to refresh the page after these changes.
If we don't close WebSocket connection, end-user can use an application as before and in the future, he will refresh the page and get new data.

Do you agree with me?

mcollina: If you change the schema, there is a very good chance that all the connections to that peer will crash/return bad data :/. I think it's better if the client is closed.

iviaks: @mcollina 
Could you tell me, how to close a current WebSocket connection?
I didn't find an interface for it :cry:

lib/gateway.js
iviaks: I have added a simple cache for it

```js
if (this._serviceSDLs === undefined) {
  this._serviceSDLs = serviceSDLs.join('')
}
```

If we need to use LRU or something else, tell me and I will change it.

lib/gateway.js
iviaks: I have updated `.refresh()` method for getting information from other services in parallel.

test/gateway/pollingInterval.js
iviaks: Yes, you're right. I have removed unsued once call and everything work fine!
Thank you :+1:

lib/gateway.js
iviaks: `Promise.allSettled()` works only with Node.js 12.9.0+

test/gateway/pollingInterval.js
mcollina: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close or `stream.destroy()`.

mcollina:
Wdyt @asciidisco

index.js
```diff
@@ -146,10 +146,27 @@ const plugin = fp(async function (app, opts) {
     schema = gateway.schema
     entityResolversFactory = gateway.entityResolversFactory
 
+    let gatewayInterval = null
+
+    if (gateway.pollingInterval !== undefined) {
```
asciidisco, line 151: Shouldn't this better test for a numeric value like `typeof gateway.pollingInterval === 'number'` or similar?
I was just thinking that it's an easy mistake to apply `null` or `false` to the config option & given how `setInterval` treats this values (falling back to it's default of `0` for the delay afaik) it would go ahead & call the function a gazillion without any delay executing tons of requests, bringing the attached services to their knees in the worst case.

asciidisco:
Let me throw my 2 cents in here; but first, please take my remakrs discouraging in any way. When I first saw the issue you opened & the PR I was super grateful, because this is something we'd also need for the software we're building on top of `fastify-gql`, so I'm super happy to see someone tacklingt that problemspace.

First question that popped up is, what happens in the error case, to be more precise, what happens when one of the requested services isn't able to answer in time when re-building the schema on the fly?
I think we can all agree on, that the network is not a 100% reliable source & there are many variables that could lead to a request not being answered in time. As far as I undertand the code, the whole federation will fail then, meaning, the gateway will bail out with an error. Please let me know if my assumption is wrong.

When starting the gateway, this might not be a bnig of a problem (and I would kinda expect the gateway to error out in such a scenario), as someone should be there to take care of this. When polling the downstream services, this problem could happen in the middle of the night, with nobody on call being able to take immediate action.

First impulse here might be, to just leave the failing node out of the gateway & just build the federated schema without it, as one of the next polls might be able to fix this problem by itself. Unfortuanately, this would work with schema stitching, but not with federation, as some services might extend other types & queries from others, depending on their availability.

For example, if you have a `me` query in a `User` service & maybe have a `my Comments` query in a `Comments` service, extending the `me` query from the `User` service, the federated schema won't be able to build if the `User` service is not available. So we might need to think of a concept dealign with this kind of errors during the schema build up.

F.e.: We flag some of our services in our service map with a `{ mandatory: true }` option, telling our federation server which kind of services are necessary for a minimal working API. In the example above, `User` would be mandatory, while `Comments` wouldn't. This is also a thing you want, if not all of your federated services are under your control. F.e. when you start up an integration test environment for each of your pull requests (like we do), some of the services that are federation might take longer to boot up than the `Gateway` after the deplyoment & then we can start the Gateway & re-integrate them later, after the Gateway has booted, falgging them as  `{ mandatory: false }`. 

Having the service erroring out with a `DOWNSTREAM` exception on runtime, on the other hand, is some scenario that I think is perfectly fine. to me, this is even one of the benefits of federation, the API continous to work partially, even when one of the services is temprorarily unavailable. We treat this as an expected outcome of each operation in our software.

I'd also like to see the `configuration` being an object literal rather than a top level `pollingInterval`, the reason for this is that it is quite likely that this federated runtime schema change mechanic will be extended in the future.

Which is also the reason why it might be a good idea to export a function that allows for retriggering the schema fetch (basically exporting what is done withing the `setInterval`).
If, for some reason, users do not want to use periodic polling to re-fetch their schemas (f.e. we use a Kafka event to trigger that behaviour) they can either implement it on their own, or a 3rd party solution could be made available as a package.


Remarks in TL;DR form:

* Error cases should be handled in case re-fetching goes wrong (or some way to handle this should be exposed in userland)
* The configuration should be an object literal, so that it can be easier
* The re-fetch should be available via a programmatic interface
* Eventually, the `serviceList`could take an additional flag, marking it as `mandatory`, but of course, that depends on the error resolution to be settled on

index.js
iviaks: Yea, I agree. I will change it, thanks :+1:

iviaks:
@asciidisco 
Thank you for your detailed explanation of the problem.
I agree with your ideas and we can do it.

> Error cases should be handled in case re-fetching goes wrong (or some way to handle this should be exposed in userland)

Error cases are an important part of each application. I agree that we need to handle it. We can do it by adding additional handler (that can be passed from properties) and a developer can configure it by himself.

> Eventually, the `serviceList` could take an additional flag, marking it as mandatory, but of course, that depends on the error resolution to be settled on

We can extend the existing object of service parameters by adding `gateway.services.mandatory` parameter. It's quite good to have.

> The re-fetch should be available via a programmatic interface

For now, we have ability to use it (https://github.com/mcollina/fastify-gql/pull/206/files#diff-168726dbe96b3ce427e7fedce31bb0bcR169). A developer is able to use `server.graphql.gateway.refresh()`.

test/gateway/pollingInterval.js
iviaks: @mcollina 

I tried to close websocket connection and it works but I can't test it.
Could you take a look at the current implementation of it? Is it correct?

asciidisco:
@iviaks Wow, that was quick 😄 
If you want, I can do a review of the additions soon-ish.
At a first glance, it looks quite good, the only thing I've noticed that the documentation for that feature is missing, if you'd like, I could give you a hand with that one if needed.

@mcollina What should be done with the failing `node 10.x` tests? @iviaks changed the code to use `Promise.allSettled` as requested, but the 10.x tests are failing now of course.

iviaks:
Thank you, @asciidisco 
I will try to write proper documentation for this feature :+1:

mcollina:
> @mcollina What should be done with the failing node 10.x tests? @iviaks changed the code to use Promise.allSettled as requested, but the 10.x tests are failing now of course.

We need to keep supporting Node 10. https://www.npmjs.com/package/promise.allsettled can do the job.

index.js
```diff
@@ -147,10 +147,27 @@ const plugin = fp(async function (app, opts) {
     schema = gateway.schema
     entityResolversFactory = gateway.entityResolversFactory
 
+    let gatewayInterval = null
+
+    if (typeof gateway.pollingInterval === 'number') {
```
asciidisco, line 152: Should we inform the user (throwing an error or smth. like this) if their `pollingInterval` option is not a numeric value? Maybe that would help them to track down errors in case they've added the number as a string etc.

lib/gateway.js
```diff
@@ -281,6 +288,47 @@ async function buildGateway (gatewayOpts, app) {
...
+    async refresh () {
+      if (this._serviceSDLs === undefined) {
+        this._serviceSDLs = serviceSDLs.join('')
+      }
+
+      await allSettled(
+        Object.values(serviceMap).map((service) =>
+          service.refresh().catch((err) => {
+            if (service.mandatory) {
+              throw Object.assign(err, service)
```
asciidisco, line 301: I'd like to hear @mcollina opinion on that as well, as I'd like to hear yours @iviaks:

I think it might be beneficial if we somehow had the ability to log (or somehow react on) the situation when a service (that's not flagged as `mandatory`) isn't reachable. Maybe just to add it to my metrics, so that I could identify patterns or to my monitoring in order to identify potential problems.

index.js
iviaks: Yes, It's a good point! :+1:

lib/gateway.js
iviaks: It's a good thing.

We can use `errorHandler` function for getting everything and the developer will be able to use it as he wants.
For example, Developer wants to log all mandatory errors, so `errorHandler` should be:

```js
const errorHandler = (error, service) => {
  if (service.mandatory) {
    logger.warn(error)
  }
}
```

Changes according to my words.

```diff
diff --git a/lib/gateway.js b/lib/gateway.js
index b9185a0..cdeac26 100644
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -297,16 +297,10 @@ async function buildGateway (gatewayOpts, app) {
       await allSettled(
         Object.values(serviceMap).map((service) =>
           service.refresh().catch((err) => {
-            if (service.mandatory) {
-              throw Object.assign(err, service)
-            }
+            errorHandler(err, service)
           })
         )
-      ).then((result) => {
-        result
-          .filter(({ reason }) => reason !== undefined)
-          .forEach(({ reason }) => errorHandler(reason, reason.service))
-      })
+      )
 
       const _serviceSDLs = Object.values(serviceMap)
         .map((service) => service.schemaDefinition)
```

asciidisco: Could you add that to the docs when writing them?

iviaks: Yes, I can :slightly_smiling_face:

asciidisco:
@iviaks I'm  back from vacation & should have some spare time this week; so if there's anything I can assist you with, let me know. If it's possible for you somehow, I'd like to get this feature mergable this week.

iviaks:
@asciidisco 
I hope we will merge this feature this week :)

lib/gateway.js
```diff
@@ -281,6 +288,47 @@ async function buildGateway (gatewayOpts, app) {
...
+            }
+          })
+        )
+      ).then((result) => {
+        result
+          .filter(({ reason }) => reason !== undefined)
+          .forEach(({ reason }) => errorHandler(reason, reason.service))
+      })
```
mcollina, line 309: I think if some of the services are offline we should not start the gateway.

asciidisco: @mcollina That's why I asked for the `mandatory` flag. In some scenarios it might be that the network has a temprorary glitch, or some of the services (which might not be under the control of the gateway development team) eventually start after the Gateway has already been booted.

We fac ethis issue, as we create a new integration test environment for each of our projects merge requests; the Java based services of another team (which are part of the federation) usually take longer to boot than our services; so we integrate them after the Gateway has already been started.
Their functionality is not mandatory for the system basics to work, so it's fine if they're not present.

F.e. we work on a Smart Home product, the core functionality is steering devices directly & execute "if this than that" like rules. We also have additional features, like historic data of device states; this isn't considered mandatory. It's "fine" if that feature, which is also part of the federation, is temporarily down, as long as the main functionality works.

this way, we can have a resiliant system which makes sure that the main functionality of the system is working fine & while it's temporarily okay for 2nd level systems, with non critical functionailty, to fail.

iviaks: but, what if offline service is not mandatory or service is restarting right now?

for example, you have the next services:
* User-Service (mandatory)
* Company-Service (not mandatory)
* Gateway Service (core)

You're developing Company-Service and it's restarting. Gateway Service at the same time tries to fetch new schema by using `pollingInterval`. Company-Service can't answer because it's restarting. Should we shutdown Gateway-Service in this case? After some time, Company-Service will be ready to work, so Gateway-Service will fetch correct schema and can continue working with it.

I guess this [solution](https://github.com/mcollina/fastify-gql/pull/206#discussion_r461401701) will be best. In this case, the Developer can operate with any error by himself.

What do you think?

PacoDu: I also think that allowing services "hot-plug" to the gateway would make it a lot more flexible for this kind of environment ! But what makes a service "mandatory" or "not mandatory" ?

mcollina: Having a `mandatory` flag is good.

iviaks: Good. Now we have it! :+1:

iviaks:
I have added documentation for this feature and fixed merge conflicts with master

mcollina:
@asciidisco wdyt?

lib/gateway.js
```diff
@@ -281,6 +289,47 @@ async function buildGateway (gatewayOpts, app) {
...
+        this._serviceSDLs = serviceSDLs.join('')
+      }
+
+      await allSettled(
+        Object.values(serviceMap).map((service) =>
+          service.refresh().catch((err) => {
+            errorHandler(err, service)
+          })
+        )
+      ).then((result) => {
```
mcollina, line 304: nit: can you remove `then()` and use another `await`?

lib/gateway.js
```diff
@@ -281,6 +289,47 @@ async function buildGateway (gatewayOpts, app) {
...
+        result
+          .filter(({ status }) => status === 'rejected')
+          .forEach(({ reason }) => {
+            throw reason
+          })
```
mcollina, line 309: nit: we might be losing some errors here. Should we report all of them?

lib/gateway.js
iviaks: Yea, I will replace `.then()` with variable. I wanted to avoid extra variables there.

lib/gateway.js
iviaks: Yea, I agree
I will throw an error with information about all errors

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
-      url: 'http://localhost:4002/graphql'
-    }]
+    ]
   }
 })
 
 await gateway.listen(4000)
 ```
 
+#### Automatically refresh schema inside Gateway mode
```
asciidisco, line 678: I would propose to change the headline to something like this: `#### Periodically refresh federated schemas in Gateway mode`

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+
+Service which uses Gateway mode can obtain a new version of schema automatically. A developer should provide a few additional parameters into the Gateway config:
```
asciidisco, line 680: I'd like to propose a change to description, something along the lines of:

`The Gateway service can obtain new versions of federated schemas automatically within a defined polling interval using the following configuration:`

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+
+- `gateway.pollingInterval` is an interval between fetching new schema (if schema didn't change, a gateway will use cached one)
```
asciidisco, line 682: I'd like to make this a bit clearer, f.e. the unit of the polling interval; so I propose to change it to something like this:
` - `gateway.pollingInterval` defines the interval (in milliseconds) the gateway should use in order to look for schema changes from the federated services. If the received schema is unchanged, the previously cached version will be reused.`

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+      },
+    ],
+    pollingInterval: 2000,
+  },
+});
+
+gateway.listen(3001);
+```
+
+#### Manually refresh schema inside Gateway mode
```
asciidisco, line 703: I think `programmatically` fits the headline better than `manually`

` #### Programmatically refresh federated schemas in Gateway mode`

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+
+Service which uses Gateway mode can obtain a new version of schema manually. A developer needs to execute method `.refresh()` from `application.graphql.gateway`. This method can return a new schema or null (when it's not changed).
```
asciidisco, line 705: I'd propose to rephrase this to something along the lines of:

`The service acting as the Gateway can manually trigger re-fetching the federated schemas programmatically by calling the `application.graphql.gateway.refresh()` method. The method either returns the newly generated schema or `null` if no changes have been discovered.`

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+setTimeout(async () => {
+  const schema = await server.graphql.gateway.refresh();
+
+  if (schema !== null) {
+    server.graphql.replaceSchema(schema);
+  }
+}, 10000);
+```
+
+#### Mark service as mandatory inside Gateway mode
```
asciidisco, line 740: `#### Flag a service as mandatory in Gateway mode´ fits the description better IMHO.

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+
+Service which uses Gateway mode can require 1+ correct services for making correct requests. A developer can provide a flag (`gateway.services.mandatory`) for better processing issues.
```
asciidisco, line 742: It think the intent of this flag needs to be explained a bit more in detail; thinking of something like:

`A Gateway service can handle the federated services in 2 different modes, `mandatory` or not by utilizing the `gateway.services.mandatory` configuration flag.
If a service is not considered mandatory, creating the federated schema will succeed even if the service isn't capable of delivering a schema.
By default, all services are considered `mandatory`.
Note: At least 1 service is necessary in order to create a valid federated schema.`

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+      },
+    ],
+  },
+  pollingInterval: 2000,
+});
+
+server.listen(3002);
+```
+
+#### Use custom errorHandler for Gateway services
```
asciidisco, line 771: I think we need to make the intent clearer, something like `#### Using a custom errorHandler for handling downstream service errors in Gateway mode`

README.md
```diff
@@ -632,46 +632,178 @@ A GraphQL server can act as a Gateway that composes the schemas of the underlyin
...
+
+Service which uses Gateway mode can process different types of issues that can be obtained from remote services (for example, Network Error). A developer can provide a function (`gateway.errorHandler`) that can process these errors.
```
asciidisco, line 773: Just a bit of rephrasing, kinda feel like we need to place the keyword `downstream` in the docs, to improve the "grep-ability" when looking for this exact thing with a browser page search function.

iviaks: Excuse me, but I didn't get you :(

asciidisco: Sorry, re-reading it, I can understand why the comment isn't really clear:

When I would be new to the library (or I'd be using it & have a specific issue with errors in federation), I'd probably do a full text search in the readme for certain keywords. One of those would be `downstream` (as for some errors scenarios, you'd get a `Downstream error` when trying to communicate with an errorneous service within the federation); so I think it might be helpful if the Readme would mention `downstream` as a keyword for people who'd try the same approach as me (using a full text search on the readme).

iviaks: Thanks, I got it now :)

asciidisco: Thank you 🙏

iviaks:
Thanks for your comments according to README @asciidisco 
Also, Thanks a lot @mcollina for your help with this PR

Now it looks much better :)

aldis-ameriks:
@iviaks @asciidisco @mcollina 
Hi,
I'm a little confused by the "mandatory" flag.
I'm trying to initialize gateway with 2 downstream services, both with `mandatory: false` flags, but I'm still getting an error (unhandled exception) when trying to start gateway if at least one service is running.
Have I misunderstood the purpose of the flag?

Running: `"fastify-gql": "5.3.3"`

```
app.register(GQL, {
  gateway: {
    services: [
      {
        name: 'api',
        url: process.env.GATEWAY_API_URL || 'http://localhost:4000/graphql',
        mandatory: false
      },
      {
        name: 'cms',
        url: process.env.GATEWAY_CMS_URL || 'http://localhost:1337/graphql',
        mandatory: false
      }
    ],
    pollingInterval: 30000
  }
})
```

PR #217, opened 2020-07-16
fix resolvers ts definition

PabloSzx:
[There is support for executableSchemas](https://github.com/mcollina/fastify-gql#makeexecutableschema-support), but the type definitions forces to define resolvers when it shouldn't.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,10 +3,6 @@ import {
   FastifyReply,
   FastifyRequest,
   FastifyInstance,
-  RawServerBase,
-  RawRequestDefaultExpression,
-  RawReplyDefaultExpression,
-  FastifyPlugin,
 } from "fastify";
 import {
   DocumentNode,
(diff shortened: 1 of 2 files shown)


asciidisco:
Same goes for `schema` when used as a Federation Gateway,

see: https://github.com/mcollina/fastify-gql#use-graphql-server-as-a-gateway-for-federated-schemas & https://github.com/mcollina/fastify-gql/blob/master/index.d.ts#L73

Might make sense to tackle this within this PR as well.

Also, I'm wondering, if it might be possible to define a function overload for https://github.com/mcollina/fastify-gql/blob/master/index.d.ts#L192 & check for the type of `schema` and/or the existence of `gateway` within the opts to adapt the types to the scenario.

PabloSzx:
> Same goes for `schema` when used as a Federation Gateway,
> 
> see: https://github.com/mcollina/fastify-gql#use-graphql-server-as-a-gateway-for-federated-schemas & https://github.com/mcollina/fastify-gql/blob/master/index.d.ts#L73
> 
> Might make sense to tackle this within this PR as well.
> 
> Also, I'm wondering, if it might be possible to define a function overload for https://github.com/mcollina/fastify-gql/blob/master/index.d.ts#L192 & check for the type of `schema` and/or the existence of `gateway` within the opts to adapt the types to the scenario.


Done


I was following https://github.com/mcollina/fastify-gql#use-graphql-server-as-a-gateway-for-federated-schemas and https://github.com/mcollina/fastify-gql#makeexecutableschema-support

PR #218, opened 2020-07-16
fix operationName schema type

PabloSzx:
Respecting the GraphQL spec the operationName can be `null`.

https://spec.graphql.org/October2016/#GetOperation()

This error caused that when executing unnamed operations and the client sends the request with operationName = null, the servers responds:

```
{
  "error": {
    "errors": [
      {
        "message": "Unknown operation named \"\"."
      }
    ],
    "data": null
  }
}
```

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -53,7 +53,7 @@ const requestProperties = {
     type: 'boolean'
   },
   operationName: {
-    type: 'string'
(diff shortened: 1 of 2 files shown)


PabloSzx:
For the test added, if you remove 'null' from the schema, it doesn't pass

WindowsTerminal_EAlrZ7hSIv

Issue #184, opened 2020-06-12
Using fastify-gql in place of Apollo Gateway does not work


flux627:
Hi,

As a proof of concept, I tried taking [this demo repository](https://github.com/apollographql/federation-demo) and replace Apollo Gateway with fastify-gql. The federated services start fine (as expected). Then, when I try to start the Gateway server, I get this error:

```
Error: The directive "@key" can only be used once at this location.
    at assertValidSDLExtension (/Users/julienheller/projects/federation-demo/node_modules/graphql/validation/validate.js:124:11)
    at extendSchema (/Users/julienheller/projects/federation-demo/node_modules/graphql/utilities/extendSchema.js:78:45)
    at buildFederationSchema (/Users/julienheller/projects/federation-demo/node_modules/fastify-gql/lib/federation.js:253:22)
    at buildGateway (/Users/julienheller/projects/federation-demo/node_modules/fastify-gql/lib/gateway.js:151:18)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async fp.name (/Users/julienheller/projects/federation-demo/node_modules/fastify-gql/index.js:119:15)
```

My guess is that both the original definition for `Product` and its extended clause have a `@key` directive, which are trying to be merged together. I wonder how Apollo handles this situation?

[Here's my fork to try it for yourself.](https://github.com/flux627/federation-demo)

```bash
npm i
npm run start-services
npm run start-fastify-gateway
```

mcollina:
Wow, good finding and thanks for checking this out.

I’m a bit unsure how that would work... would you like to send a Pull Request to address this issue? Remember to add unit tests.

flux627:
I would help, but I don't know the proper expected implementation of the Gateway (in this case the expected way to handle conflicting directives when creating the combined schema). I tried looking for a specification, but could only find an [implementation specification for the Federated services](https://www.apollographql.com/docs/apollo-server/federation/federation-spec/), and an [API specification for the Gateway](https://www.apollographql.com/docs/apollo-server/api/apollo-federation/), but not a specification for implementing a Gateway. Under what basis was the Gateway implementation designed? By just reverse engineering the Apollo Gateway implementation?

I could do my best to try different things regarding directives using the Apollo implementation and copy the observed functionality over to fastify-gql, if there's no better expected methodology.

flux627:
I just found [this](https://www.apollographql.com/docs/apollo-server/federation/gateway/#type-system-directives):

[...]

PR #220, opened 2020-07-17
Fix schema.

voslartomas:


mcollina:
@frikille would you have a few minutes to take a look?

asciidisco:
@mcollina @voslartomas I tend to believe that this is only a sympton of the issue described in https://github.com/mcollina/fastify-gql/issues/184

paolochiodi:
I agree with @asciidisco. The error this PR is fixing seems to be related to #184. 

As I explained in https://github.com/mcollina/fastify-gql/issues/219#issuecomment-660157369 we get an error similar to the one described in #184. 
I think there's slight difference here though: we get away with removing `@key` on just one of the extend User. We can either remove it from https://github.com/mcollina/fastify-gql/blob/master/examples/gateway-subscription.js#L111 or from https://github.com/mcollina/fastify-gql/blob/master/examples/gateway-subscription.js#L173 and in both cases the error goes away and the gateway boots.

So, it is true that this PR will fix the immediate issue with the gateway subscription and will make the server boot, but I think it is not the correct fix because as far as I can see the schema is a legitimate one and should work.

The underlying cause should be fixed instead.

paolochiodi:
> So, it is true that this PR will fix the immediate issue with the gateway subscription and will make the server boot, but I think it is not the correct fix because as far as I can see the schema is a legitimate one and should work.
> 
> The underlying cause should be fixed instead.

Actually, let me take that back and ask the questions: is this a valid schema?

In this case it seems so because `@key` holds the same definition in both federated services. But what if the two services had different definitions? How could the gateway merge them?

And what's the best way to signal this to the user? Should the gateway silently ignore the conflicts?

asciidisco:
@paolochiodi Which emphasizes again on what @flux627 https://github.com/mcollina/fastify-gql/issues/184#issuecomment-644174562 said & something that probably @frikille can answer best, as the original author of that code.

One could ask how `apollo` behaves in such cases, but then would also need to ask if `apollo` should be used as the reference implementation here, or not.

paolochiodi:
@asciidisco yep, that's correct.

For the record, I found this in the apollo docs at https://www.apollographql.com/docs/apollo-server/federation/entities/#referencing

> The @key directive indicates that Product uses the upc field as its primary key. This value must match the value of exactly one @key defined in the entity's originating service, even if the entity defines multiple primary keys.

Also worth noting that `fastify-gql`'s gateway seems to be doing the right thing* when an entity is extended only once. The problem occurs when it's extended twice.

 * the right thing: one can replace it with "the thing apollo also does"

flux627:
> One could ask how `apollo` behaves in such cases, but then would also need to ask if `apollo` should be used as the reference implementation here, or not.

Apollo is the authority here, and this is about more than implementation details. They have docs describing the system, but since they don't have an official spec, then we need to take their docs + implementation together as a de-facto spec. Otherwise, we're implementing something different than Federation. That's ok, but we shouldn't claim that this is Federation in that case. I personally think it's in this library's best interest to match the functionality of Apollo and not diverge into its own flavor of schema delegation (especially if it's marketed as Federation and is almost Federation).

FWIW, I was originally looking into this library for its subscription support in combination with Federation, but have since moved on to using schema stitching via the new `graphql-tools` `stitchSchemas()` type merging.

paolochiodi:
I've done some more research and I may have found the route cause... but I'm moving the discussion to https://github.com/mcollina/fastify-gql/issues/184 as it seems more relevant there.

voslartomas:
> > One could ask how `apollo` behaves in such cases, but then would also need to ask if `apollo` should be used as the reference implementation here, or not.
> 
> Apollo is the authority here, and this is about more than implementation details. They have docs describing the system, but since they don't have an official spec, then we need to take their docs + implementation together as a de-facto spec. Otherwise, we're implementing something different than Federation. That's ok, but we shouldn't claim that this is Federation in that case. I personally think it's in this library's best interest to match the functionality of Apollo and not diverge into its own flavor of schema delegation (especially if it's marketed as Federation and is almost Federation).
> 
> FWIW, I was originally looking into this library for its subscription support in combination with Federation, but have since moved on to using schema stitching via the new `graphql-tools` `stitchSchemas()` type merging.

How exactly are you doing this please? You have one server witch is stitching federated schema and subscriptions? Another server is Federation and then services?

flux627:
I have a gateway server (not an *Apollo Federation Gateway™*)  that merges the types from multiple other subschemas. No directives required, but you do need to implement queries on the subschemas to fetch types via some identifier/key. [See the docs here.](https://www.graphql-tools.com/docs/schema-stitching/)

As an exercise, I took [this Apollo federation demo](https://github.com/apollographql/federation-demo) and converted it to use [`graphql-tools` schema stitching](https://github.com/flux627/federation-demo/tree/stitched), if you want to see a direct comparison.

For subscriptions, check out the code from [this comment](https://github.com/ardatan/graphql-tools/issues/864#issuecomment-649067840).

voslartomas:
@flux627  I see, so it's the old approach before federation. Or am I missing something?

flux627:
The `stitchSchemas` function, along with its type merging abilities, is brand new to `graphql-tools` v6. The old method of schema stitching (and the one that Apollo talks about as deprecated/obsolete) is much more cumbersome than what this provides.

voslartomas:
@flux627 I see, will take a look, thanks.

asciidisco:
@voslartomas As the conversation has been going into the direction of fixing the root cause for this, I think this could be closed.

PR #221, opened 2020-07-17
Fix to federation with multiple services extending the same entity

paolochiodi:
This is a quick fix for https://github.com/mcollina/fastify-gql/issues/184

See https://github.com/mcollina/fastify-gql/issues/184#issuecomment-660192118 for explanation.

This is a quick hack intended to demonstrate the issue. I'm happy to extend it to a proper PR with tests if maintainers are happy with the direction and intend to solve this in this repo rather then reporting to graphql-js.

### DIFF
--- a/lib/federation.js
+++ b/lib/federation.js
@@ -250,10 +250,12 @@ function buildFederationSchema (schema, isGateway) {
   })
 
   // Add all extensions
-  federationSchema = extendSchema(federationSchema, {
-    kind: Kind.DOCUMENT,
-    definitions: extensions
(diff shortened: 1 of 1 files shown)


paolochiodi:
To test the PR you can temporarily run `node examples/gateway-subscription.js`: master will fail with an error and this branch will work.

asciidisco:
@paolochiodi Really good work with https://github.com/mcollina/fastify-gql/pull/227 thank you a lot. I think we can close this one now, correct?

PR #222, opened 2020-07-17
Allow hasDirective to handle nodes that do not contain directives.

HW13:
### Summary

Currently if the SDL returned from the gateway's [`ServiceInfo`](https://github.com/mcollina/fastify-gql/blob/99a06a9ff480bc5aa8a6ab4b925b6d876f125645/lib/gateway/service-map.js#L67) query contains any directive definitions, the following error will be be thrown:
```js
TypeError: Cannot read property 'some' of undefined
```
This error originates from the [`hasDirective`](https://github.com/mcollina/fastify-gql/blob/99a06a9ff480bc5aa8a6ab4b925b6d876f125645/lib/gateway/service-map.js#L13) function, and is thrown because a [`DirectiveNode`](https://github.com/graphql/graphql-js/blob/6012d28b3b1fe8e002eefed9a3f013d4b352f08d/src/language/ast.js#L414) does not contain a `directives` field. 

Changes in this PR allow `hasDirective` to handle any [`ASTNode`](https://github.com/graphql/graphql-js/blob/6012d28b3b1fe8e002eefed9a3f013d4b352f08d/src/language/ast.js#L144) with a non-existent `directives` field.

### Test Plan
A custom directive has been added to the `userServicePort` schema in the schema test. Without the changes made to `hasDirective`, this test would fail (with the aforementioned error being thrown).

### DIFF
--- a/lib/gateway/service-map.js
+++ b/lib/gateway/service-map.js
@@ -11,7 +11,8 @@ const { buildRequest, sendRequest } = require('./request')
 const SubscriptionClient = require('../subscription-client')
 
 function hasDirective (directiveName, node) {
-  return node.directives.some(directive => directive.name.value === directiveName)
+  const { directives = [] } = node
+  return directives.some(directive => directive.name.value === directiveName)
 }
 
 function createFieldSet (definition, filterFn = () => false) {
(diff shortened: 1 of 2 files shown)


HW13:
Not sure what the convention is here - I hope no one minds that I opened this without creating an issue first. 😃

HW13:
No problem, thanks for the review! 😎 👍

PR #226, opened 2020-07-23
Add playground to type definitions

delaneyj:


delaneyj:
Just realized I was misreading how to enable the playground.  Technically its `graphiql:'playgroud'` but that seems very confusing as graphiql and playground aren't the same thing (though it looks like they might be merging in the future).  Would it make more since to rename the graphiql to `editor` and then be able to support playground,graphiql,altair, etc?

mcollina:
there is already the `ide` option that does that.

PR #227, opened 2020-07-27
Remove UniqueDirectivesPerLocationRule from validation for federated schemas

paolochiodi:
This fixes https://github.com/mcollina/fastify-gql/issues/184

As discussed with @mcollina with this PR I'm going with the simplest fix just for the linked issue. 
We replicate Apollo's behaviour by moving validations in our code and skipping the `UniqueDirectivesPerLocationRule` rule.

### DIFF
--- a/lib/federation.js
+++ b/lib/federation.js
@@ -34,6 +34,8 @@ const {
   isTypeExtensionNode,
   isObjectType
 } = require('graphql')
+const { validateSDL } = require('graphql/validation/validate')
+const compositionRules = require('./federation/compositionRules')
 
 const BASE_FEDERATION_TYPES = `
   scalar _Any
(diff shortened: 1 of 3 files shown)


lib/federation/compositionRules.js
```diff
@@ -0,0 +1,8 @@
+const { specifiedSDLRules } = require('graphql/validation/specifiedRules')
```
mcollina, line 1:
```suggestion
'use strict'

const { specifiedSDLRules } = require('graphql/validation/specifiedRules')
```

mcollina:
@flux627 can you check?

mcollina:
cc @asciidisco

mcollina:
cc @pscarey

lib/federation.js
```diff
@@ -250,10 +252,25 @@ function buildFederationSchema (schema, isGateway) {
...
     definitions: extensions
-  })
+  }
+
+  // instead of relying on extendSchema internal validations
+  // we run validations in our code so that we can use slightly different rules
+  // as extendSchema internal rules are meant for regular usage
+  // and federated schemas have different constraints
+  const errors = validateSDL(extensionsDocument, federationSchema, compositionRules)
+  if (errors.length === 1) {
```
pscarey, line 265: I think this error handling is messy.

I would suggest always throwing an error with the message `Invalid Federation Schema` or similar, and always setting the errors property.

paolochiodi: I copied the approach from https://github.com/mcollina/fastify-gql/blob/master/index.js#L158, but I'm happy to simplify it if we don't mind taking a different approach.

(or I guess we can change that code to have the same behaviour?)

mcollina: What we have found in several other issues is that devs will not look into the `errors` property and open issues about invalid schema :/. See https://github.com/mcollina/fastify-gql/issues/215 and https://github.com/mcollina/fastify-gql/pull/216. I agree this is far from ideal but it provides a better developer experience.

pscarey:
@mcollina I'm not an expert in schema federation (to clarify we're not using the federation aspect of fastify-gql but between JIT and Fastify have hit a 7x performance improvement over Apollo server on some critical workloads) but this fix seems true to Apollo's solution/workaround. 

It may be worth filing a further issue for more test cases and tweaks to validate the behaviour around the other rules which Apollo skips, found [here](https://github.com/apollographql/apollo-server/blob/00887a1e57a26075483ba397098f6d088a706499/packages/apollo-federation/src/composition/rules.ts).

@paolochiodi Nice work getting to the root of the issue.

mcollina:
> It may be worth filing a further issue for more test cases and tweaks to validate the behaviour around the other rules which Apollo skips, found here.

Can you open such an issue @paolochiodi ?

PR #228, opened 2020-07-27
Handle subscriptions with stream

PacoDu:
This PR tries to solve the message processing order issue encountered in #167 by handling subscription as stream.

I've modified `SubscriptionConnection` to use an async iteration on the connection instead of the `socket.on('data')`. With some modifications to `test/subscription-connection.js` it seems to work. But when multiple messages are sent through the ws it is not working as expected, the handler receives multiple json in one iteration (see: https://github.com/PacoDu/fastify-gql/runs/914696422#step:5:1119)

```
# Subtest: subscription server sends update to subscriptions
        ok 1 - should not error
    message raw: {"type":"connection_init"}{"id":1,"type":"start","payload":{"query":"\n          subscription {\n            notificationAdded {\n              id\n              message\n            }\n          }\n        "}}{"id":2,"type":"start","payload":{"query":"\n          subscription {\n            notificationAdded {\n              id\n              message\n            }\n          }\n        "}}{"id":2,"type":"stop"}
```

My knowledge with streams is limited, any help/review/idea to solve this issue would be welcomed :)

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const { on } = require('events')
 const { subscribe, parse } = require('graphql')
 const { SubscriptionContext } = require('./subscriber')
 const { kEntityResolvers } = require('./gateway/make-resolver')
@@ -13,7 +14,6 @@ const {
   GQL_COMPLETE,
   GQL_STOP
 } = require('./subscription-protocol')
-const { on } = require('events')
 
 module.exports = class SubscriptionConnection {
   constructor (socket, {
@@ -24,8 +24,8 @@ module.exports = class SubscriptionConnection {
     entityResolvers
   }) {
     this.fastify = fastify
-    this.schema = schema
     this.socket = socket
+    this.schema = schema
     this.lruGatewayResolvers = lruGatewayResolvers
     this.entityResolvers = entityResolvers
     this.subscriber = subscriber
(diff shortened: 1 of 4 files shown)


PacoDu:
I've changed the async iterator as you suggested:
```js
try {
      for await (const message of on(this.socket, 'message')) {
        await this.handleMessage(message)
      }
    } catch (err) {
      this.fastify.log.error(err)
      this.handleConnectionClose()
    }
```
but tests hangs at https://github.com/mcollina/fastify-gql/pull/228/checks?check_run_id=915124640#step:5:356

PacoDu:
The test hangs after the subscribe https://github.com/mcollina/fastify-gql/pull/228/files#diff-d0a65557fba21a77676fdc6c8ab721c7L105

```
subscription {
      notificationAdded {
      id
      message
   }
 }
```

PacoDu:
> Let me know If you can't find a polyfill for Node 10 for that.

Do you have any recommended polyfill for Node 10 for `on`, I couldn't find one

lib/subscription-connection.js
```diff
@@ -30,13 +31,17 @@ module.exports = class SubscriptionConnection {
...
-    this.socket.on('message', (message) => {
-      this.handleMessage(message).catch(e => {
-        this.fastify.log.error(e)
+    this.socket.on('error', this.handleConnectionClose.bind(this))
+    this.handleConnection()
+  }
+
+  async handleConnection () {
+    for await (const message of on(this.socket, 'message')) {
+      this.handleMessage(message).catch(err => {
```
mcollina, line 40: you need to await this

PacoDu: > The test hangs after the subscribe https://github.com/mcollina/fastify-gql/pull/228/files#diff-d0a65557fba21a77676fdc6c8ab721c7L105
> 
> ```
> subscription {
>       notificationAdded {
>       id
>       message
>    }
>  }
> ```

As said earlier, I don't understand why but if I await this line it blocks everything

PacoDu: I think it's because the `handleMessage` function doesn't return if a subscription is called with `handleGQLStart`: https://github.com/mcollina/fastify-gql/blob/master/lib/subscription-connection.js#L105

PacoDu: Do you have any suggestion ? This solution doesn't work because of the `handleMessage` will not return until the subscription is dropped, therefore blocking the iteration if we await `handleMessage`.

PacoDu: The subscription context is store in `this.subscriptionContexts` and deleted if the connection is closed so I've removed the `await` before `handleGQLStart`, is this safe ?

mcollina:
You can use https://www.npmjs.com/package/events.on to support node 10.x

lib/subscription-connection.js
mcollina: This is ok, yes.

PacoDu:
Waiting for https://github.com/mcollina/events.on/pull/1

mcollina:
shipped events.on v1.0.1

Issue #207, opened 2020-07-10
Federation subscriptions: check socket state before sending data.


SkeLLLa:
https://github.com/mcollina/fastify-gql/blob/master/lib/subscription-client.js#L138

Here before sending anything to socket it needs to be checked if it's in ready/connected state. Otherwise it will throw an error that is not cached anywhere properly.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #234, opened 2020-08-01
Fix async subscription context

### LINKED ISSUE #208: Federation: Handle situation when graphql subscribe result is not AsyncIterable.
https://github.com/mcollina/fastify-gql/blob/master/lib/subscription-connection.js#L88 

Graphql's subscribe method doesn't return AsyncIterable always. It also may return simple ExecutionResult. And if it will return it, we'll get an error.

It's pretty much the same case as in https://github.com/mcollina/fastify-gql/pull/174, when ExecutionResult just ignored.

### LINKED ISSUE #235: Async subscription context kills the subscriptions.
I opened a draft PR (#234)  showing a bug that happens when the context for the subscriptions is an async function, I discovered it trying to make the context based on an database request, so I simulated the situation adding an artificial delay on a existing test and it did fail as expected.

In theory the code is awaiting (https://github.com/mcollina/fastify-gql/blob/master/lib/subscription.js#L27), but there seems to be a weird race condition somewhere.

PabloSzx:
This PR is just a draft for a failing test that shows the async subscription context bug.

It basically the same as the existing test for custom context, but with an artificial delay simulating a common http request delay.

EDIT: Added fix.

Fixes #235
Fixes #208

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -24,7 +24,8 @@ module.exports = class SubscriptionConnection {
     lruGatewayResolvers,
     entityResolvers,
     context,
-    onConnect
+    onConnect,
+    resolveContext
   }) {
     this.fastify = fastify
     this.socket = socket
@@ -35,6 +36,8 @@ module.exports = class SubscriptionConnection {
     this.onConnect = onConnect
     this.subscriptionContexts = new Map()
     this.context = context
+    this.isReady = false
+    this.resolveContext = resolveContext
 
     this.socket.on('error', this.handleConnectionClose.bind(this))
     this.handleConnection()
@@ -69,11 +72,17 @@ module.exports = class SubscriptionConnection {
       case GQL_CONNECTION_TERMINATE:
         this.handleConnectionClose()
         break
-      case GQL_START:
-        this.handleGQLStart(data).catch(e => {
-          this.sendMessage(GQL_ERROR, id, e.message)
-        })
+      case GQL_START: {
+        if (this.isReady) {
+          this.handleGQLStart(data).catch(e => {
+            this.sendMessage(GQL_ERROR, id, e.message)
+          })
+        } else {
+          this.sendMessage(GQL_CONNECTION_ERROR, undefined, { message: 'Connection has not been established yet.' })
+          return this.handleConnectionClose()
+        }
         break
+      }
       case GQL_STOP:
         this.handleGQLStop(data)
         break
@@ -83,9 +92,9 @@ module.exports = class SubscriptionConnection {
   }
 
   async handleConnectionInit (data) {
-    if (this.context instanceof Promise) {
+    if (typeof this.resolveContext === 'function') {
       try {
-        this.context = await this.context
+        this.context = await this.resolveContext()
       } catch (err) {
         this.fastify.log.error(err)
         // 1011: Internal Error
@@ -116,6 +125,7 @@ module.exports = class SubscriptionConnection {
     }
 
     this.sendMessage(GQL_CONNECTION_ACK)
+    this.isReady = true
   }
 
   async handleGQLStart (data) {
@@ -147,6 +157,11 @@ module.exports = class SubscriptionConnection {
       variables,
       operationName
     )
+
+    if (result.errors) {
+      this.fastify.log.error(result.errors)
+      throw result.errors[0]
+    }
 
     // TODO implement backpressure
     for await (const value of result) {
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -83,6 +83,17 @@ module.exports = class SubscriptionConnection {
   }
 
   async handleConnectionInit (data) {
+    if (this.context instanceof Promise) {
+      try {
+        this.context = await this.context
+      } catch (err) {
+        this.fastify.log.error(err)
+        // 1011: Internal Error
+        this.socket.close(1011)
+        return
+      }
+    }
+
     if (typeof this.onConnect === 'function') {
       let authorize = false
       try {
--- a/lib/subscription.js
+++ b/lib/subscription.js
@@ -24,7 +24,15 @@ function createConnectionHandler ({ schema, subscriber, fastify, onConnect, lruG
     let context = {}
     if (subscriptionContextFn) {
       try {
-        context = await subscriptionContextFn(connection, request)
+        context = subscriptionContextFn(connection, request)
+
+        if (context instanceof Promise) {
+          context.catch(err => {
+            fastify.log.error(err)
+            // 1011: Internal Error
+            socket.close(1011)
+          })
+        }
       } catch (err) {
         fastify.log.error(err)
         // 1011: Internal Error
--- a/lib/subscription.js
+++ b/lib/subscription.js
@@ -21,24 +21,11 @@ function createConnectionHandler ({ schema, subscriber, fastify, onConnect, lruG
       return
     }
 
-    let context = {}
-    if (subscriptionContextFn) {
-      try {
-        context = subscriptionContextFn(connection, request)
+    const context = {}
+    let resolveContext
 
-        if (context instanceof Promise) {
-          context.catch(err => {
-            fastify.log.error(err)
-            // 1011: Internal Error
-            socket.close(1011)
-          })
-        }
-      } catch (err) {
-        fastify.log.error(err)
-        // 1011: Internal Error
-        socket.close(1011)
-        return
-      }
+    if (subscriptionContextFn) {
+      resolveContext = () => subscriptionContextFn(connection, request)
     }
 
     const subscriptionConnection = new SubscriptionConnection(socket, {
@@ -48,7 +35,8 @@ function createConnectionHandler ({ schema, subscriber, fastify, onConnect, lruG
       onConnect,
       lruGatewayResolvers,
       entityResolvers: entityResolversFactory && entityResolversFactory.create(),
-      context
+      context,
+      resolveContext
     })
 
     /* istanbul ignore next */
--- a/test/subscription.js
+++ b/test/subscription.js
@@ -508,3 +508,234 @@ test('subscription connection is closed if context function throws', t => {
     })
   })
 })
+
+test('subscription server sends update to subscriptions with custom async context', t => {
+  const app = Fastify()
+  t.tearDown(() => app.close())
+
+  const sendTestQuery = () => {
+    app.inject({
+      method: 'POST',
+      url: '/graphql',
+      body: {
+        query: `
+          query {
+            notifications {
+              id
+              message
+            }
+          }
+        `
+      }
+    }, () => {
+      sendTestMutation()
+    })
+  }
+
+  const sendTestMutation = () => {
+    app.inject({
+      method: 'POST',
+      url: '/graphql',
+      body: {
+        query: `
+          mutation {
+            addNotification(message: "Hello World") {
+              id
+            }
+          }
+        `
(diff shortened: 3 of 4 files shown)


mcollina:
Would you mind to also fix it? I'm not sure what's the problem:/

PabloSzx:
> Would you mind to also fix it? I'm not sure what's the problem:/

image

image

I found this solution (that could be improved with an `isPromise` helper function), which is to basically let the socket connection be created and delay the await until `handleConnectionInit`.

I'd like to know if it sounds good to you.

lib/subscription.js
```diff
@@ -24,7 +24,15 @@ function createConnectionHandler ({ schema, subscriber, fastify, onConnect, lruG
     let context = {}
     if (subscriptionContextFn) {
       try {
-        context = await subscriptionContextFn(connection, request)
+        context = subscriptionContextFn(connection, request)
+
+        if (context instanceof Promise) {
```
PabloSzx, line 29: The async error was not being catched without doing this

lib/subscription-connection.js
```diff
@@ -83,6 +83,17 @@ module.exports = class SubscriptionConnection {
   }
 
   async handleConnectionInit (data) {
+    if (this.context instanceof Promise) {
+      try {
+        this.context = await this.context
+      } catch (err) {
+        this.fastify.log.error(err)
```
PabloSzx, line 90: These couple of lines are not being tested, but I'm not sure if it's completely safe to just simply remove the try/catch, just because of the catch in `lib/subscription.js`

PabloSzx:
> (or `instanceof Promise` if it should only support native Promises).

In the last changes I just assumed this, I honestly don't know if I can assume this

lib/subscription-connection.js
PabloSzx: Nevermind, I just sent the message 'connection_init' in the intended test and it covered it. 100% coverage.

PabloSzx:
I found another kinda important thing, when an error happens inside the execution of a subscription the error is not being catched properly, and it just replies with `the result is not async iterable` .

Following the [type definitions docs](https://github.com/graphql/graphql-js/blob/master/src/subscription/subscribe.d.ts#L22) in `graphql-js` `subscribe` function, those errors are returned in the execution results without the promise itself being thrown.

I just added
```js
if (result.errors) {
      throw result.errors[0];
}
``` 

[in subscription-connection.js](https://github.com/mcollina/fastify-gql/blob/master/lib/subscription-connection.js#L139)

And it works as expected.


I don't know if I should add it to this PR 🤷‍♂️

asciidisco:
@PabloSzx I wonder if your latest finding might be related to https://github.com/mcollina/fastify-gql/issues/208

PabloSzx:
> @PabloSzx I wonder if your latest finding might be related to #208

Yes, it is,  it's ok if you assigned yourself this fix yourself👍 

The `graphql` package itself validates the error thrown in the execution, it means there should be no need of validating the error inside result.errors, but I am not sure if it's needed to handle that there might be more than a single error.

<details><summary>Test I made in subscription.js if it's helpful at all</summary>
<p>

```js
test('subscription server send correct error if execution throws', t => {
  const app = Fastify()
  t.tearDown(() => app.close())

  const emitter = mq()
  const schema = `
    type Notification {
      id: ID!
      message: String
    }

    type Query {
      notifications: [Notification]
    }

    type Subscription {
      notificationAdded: Notification
    }
  `

  const resolvers = {
    Subscription: {
      notificationAdded: {
        subscribe: () => {
          throw Error('custom execution error')
        }
      }
    }
  }

  app.register(GQL, {
    schema,
    resolvers,
    subscription: {
      emitter
    }
  })

  app.listen(0, err => {
    t.error(err)

    const ws = new WebSocket('ws://localhost:' + (app.server.address()).port + '/graphql', 'graphql-ws')
    const client = WebSocket.createWebSocketStream(ws, { encoding: 'utf8', objectMode: true })
    t.tearDown(client.destroy.bind(client))
    client.setEncoding('utf8')

    client.write(JSON.stringify({
      type: 'connection_init'
    }))

    client.write(JSON.stringify({
      id: 1,
      type: 'start',
      payload: {
        query: `
          subscription {
            notificationAdded {
              id
              message
            }
          }
        `
      }
    }))

    client.on('data', chunk => {
      const data = JSON.parse(chunk)

      if (data.id === 1 && data.type === 'error') {
        t.equal(chunk, JSON.stringify({
          type: 'error',
          id: 1,
          payload: 'custom execution error'
        }))

        client.end()
        t.end()
      }
    })
  })
})
```

</p>
</details>

PacoDu:
I confirm the issue if the contextFn is async the subscription listener is missing the messages sent in the 1st PR test, but the 2nd test works. I guess we can't do async stuff before binding the message handler.

I think it would be better to remove the context creation from `createConnectionHandler` and pass the `subscriptionContextFn` to `SubscriptionConnection` constructor. 
But this rises an other issue concerning `connection_init` and message order.

Based on apollo implementation https://github.com/apollographql/subscriptions-transport-ws/blob/52758bfba6190169a28078ecbafd2e457a2ff7a8/src/server.ts and https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md, if an `onConnect` handler is defined and returns `false` a `GQL_START` message should return a `GQL_CONNECTION_ERROR`.

asciidisco:
@PabloSzx I just assigned myself as an ToDo item, because the person that has created the issue seems to be not interested in fixing & I have a slight feeling that this can hit us as well once we're done migrating from Apollo to Fastify. That being said, I would suggest that you can re-assign this to you & I also think it would be reasonable to fix it within this PR (of course, this is just my subjective opinion)

lib/subscription.js
mcollina: I don't understand why using `await` would not have worked. can you articulae?

PabloSzx:
> I confirm the issue if the contextFn is async the subscription listener is missing the messages sent in the 1st PR test, but the 2nd test works. I guess we can't do async stuff before binding the message handler.
> 
> I think it would be better to remove the context creation from `createConnectionHandler` and pass the `subscriptionContextFn` to `SubscriptionConnection` constructor.
> But this rises an other issue concerning `connection_init` and message order.
> 
> Based on apollo implementation https://github.com/apollographql/subscriptions-transport-ws/blob/52758bfba6190169a28078ecbafd2e457a2ff7a8/src/server.ts and https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md, if an `onConnect` handler is defined and returns `false` a `GQL_START` message should return a `GQL_CONNECTION_ERROR`.

related to #207 I think





> @PabloSzx I just assigned myself as an ToDo item, because the person that has created the issue seems to be not interested in fixing & I have a slight feeling that this can hit us as well once we're done migrating from Apollo to Fastify. That being said, I would suggest that you can re-assign this to you & I also think it would be reasonable to fix it within this PR (of course, this is just my subjective opinion)

I will do, no problem, plus the #207

lib/subscription.js
PabloSzx: awaiting the context here doesn't let the socket connection be properly created, the possible promise should be started, but it's result is not used until the socket messages 'connection_init' / 'start'.

mcollina: Then let's move where we do this operation. If the context should be created and awaited during `connection_init`, then let's do it there. This adds complexity that could bite us in the future.

lib/subscription-connection.js
```diff
@@ -35,6 +35,7 @@ module.exports = class SubscriptionConnection {
     this.onConnect = onConnect
     this.subscriptionContexts = new Map()
     this.context = context
+    this.isReady = false
```
mcollina, line 38: where is `isReady = true` set?

PabloSzx: at the end of handleConnectionInit, it's to verify the message order as @PacoDu mentioned

lib/subscription.js
PabloSzx: The thing is, as I mentioned, should the context be created at socket creation or in `connection_init`?, as the test `subscription connection is closed if context function **throws**` suggests, it's still possible to make the context without the `connection_init` being received.

As I said previously:

> EDIT: The Context should be made (and it's possible errors, handled) in the request context, and the subcription connection should just receive the promise, since like the test 'subscription connection is closed if context function throws' in `subscription.js` test file suggests, it's possible to start the socket connection first without the 'connection_init' and the context should already be in process. Just passing the function to be executed in `subcription-connection.js` delays the promise unnecessarily.

lib/subscription-connection.js
```diff
@@ -83,6 +90,17 @@ module.exports = class SubscriptionConnection {
   }
 
   async handleConnectionInit (data) {
+    if (this.context instanceof Promise) {
```
mcollina, line 93: remove this, there is no need for this check.

lib/subscription.js
mcollina: I would move it to https://github.com/mcollina/fastify-gql/pull/234/files#diff-d0a65557fba21a77676fdc6c8ab721c7R95

PabloSzx: Then I will have to modify the mentioned test and give a function to the subscription connection that resolves the context instead of just receiving it.

test/subscription.js
```diff
@@ -508,3 +513,316 @@ test('subscription connection is closed if context function throws', t => {
...
+    }
+  `
+
+  const resolvers = {
+    Query: {
+      add: (parent, { x, y }) => x + y
+    }
+  }
+
+  const wait = (amount = 0) => new Promise(resolve => setTimeout(resolve, amount))
```
mcollina, line 714: please use util.promisify instead.

test/subscription.js
```diff
@@ -508,3 +513,316 @@ test('subscription connection is closed if context function throws', t => {
...
+
+  app.register(GQL, {
+    schema,
+    resolvers,
+    subscription: {
+      context: async function () {
+        await wait(200)
```
mcollina, line 721: Why is this timeout needed?

wesselvdv:
I am hit big time by this atm, good to see that it's almost merged!

test/subscription.js
PabloSzx: this timeout triggers the race condition of the socket not being created because of the context creation await, this simulates the reason this PR was created

mcollina: Then you can use the following instead:

```js
const immediate = promisify(setImmediate)

// in the test
await immediate()
```

PabloSzx: with setTimeout(() => {}, 0), which is what setImmediate does, the race condition was not happening, the number should be high for it to happen.

PabloSzx: the problem was not happening just because it was async, was because of the delay intrinsic the async, which is we are simulating with the delay 200, you can test this problem yourself removing the code that solves this and keeping the tests.

mcollina: Can we use the fake timers instead then? setTimeout introduce flakyness.

PabloSzx: are they being installed/used already somewhere in the code?

mcollina: https://github.com/mcollina/fastify-gql/blob/1335259f51710521b3d81cfb01e78d84d241577c/test/gateway/pollingInterval.js#L5 use the fake timers.

PabloSzx: I was checking the timers and they don't do what the test needs need to do, we need to actually make the socket wait, and the number needs to be 50ms+, the flakyness doesn't matter

PR #244, opened 2020-08-09
update(types): add cache map

RafaelGSS:
ref: https://github.com/fastify/help/issues/227

I'll investigate later it has a bug on the cache. This should be added btw


RafaelGSS:

I added to `types` test, but I don't know if is about these tests you said.

PR #246, opened 2020-08-11
fix: fix ide path routing based on path option

gtolarc:


gtolarc:
When I think about it again, it can be solved with a combination of path and prefix options. However, there is an error when using graphiql/playground, but I will upload a new PR that has been corrected.

PR #248, opened 2020-08-11
Fix inconsistency in handling errors

mcollina:
Fixes https://github.com/fastify/help/issues/227

### DIFF
--- a/index.js
+++ b/index.js
@@ -426,9 +426,7 @@ const plugin = fp(async function (app, opts) {
     if (cached && cached.jit !== null) {
       const execution = await cached.jit.query(root, context, variables || {})
 
-      maybeFormatErrors(execution, reply)
-
-      return execution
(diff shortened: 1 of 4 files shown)


mcollina:
cc @paul590 @RafaelGSS

index.js
```diff
@@ -425,15 +424,11 @@ const plugin = fp(async function (app, opts) {
...
 
-      if (res.errors) {
-        const { response: { data, errors } } = errorFormatter(res)
-        res.data = data
-        res.errors = errors
-      }
+      maybeFormatErrors(execution, reply)
 
-      return res
+      return execution
```
pscarey, line 431: Would it be neater to make the API `return formatResult(execution, reply)`, and update `maybeFormatErrors` to `formatResult`, with an extra `return execution` at the end?

pscarey:
Other than comment, looks good to me.

Issue #81, opened 2019-10-28
Add examples for using subscriptions


mcollina:
With:

* [ ] redis https://github.com/mcollina/mqemitter-redis
* [ ] mongodb https://github.com/mcollina/mqemitter-mongodb
* [ ] cluster/multiple process on the same machine https://github.com/mcollina/mqemitter-child-process

santiagoalmeidabolannos:
looks like there is a redis example in the Readme file, are you looking for something different?

mcollina:
I think this is done

PR #250, opened 2020-08-12
add mqemitter mongodb subscription example (#81)

santiagoalmeidabolannos:
starting with the mongodb example, trying something a bit different

### DIFF
--- a/examples/subscription/mqemitter-mongodb-subscription.js
+++ b/examples/subscription/mqemitter-mongodb-subscription.js
@@ -1,0 +1,90 @@
+'use strict'
+
+const GQL = require('fastify-gql')
+const Fastify = require('fastify')
+const mongodbMQEmitter = require('mqemitter-mongodb')
+
+const app = Fastify({ logger: true })
+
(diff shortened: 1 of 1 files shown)


examples/subscription/mqemitter-mongodb-subscription.js
```diff
@@ -0,0 +1,120 @@
+const GQL = require('fastify-gql')
```
mcollina, line 1: please add `'use strict'`

examples/subscription/mqemitter-mongodb-subscription.js
```diff
@@ -0,0 +1,120 @@
...
+const Fastify = require('fastify')
+const mongoose = require('mongoose')
```
mcollina, line 3: I would prefer to not use mongoose in an example.

santiagoalmeidabolannos: done, I was trying an example with a different place where emit events but I understand your point

PR #251, opened 2020-08-14
Added logging to the default error formatter

mcollina:
As titled


paolochiodi:
Tested, and it works. 

But should the logging be part of the error formatter?

Seems like a side effect that may not be evident to users, and thus they'd fail to reimplement it when they use a custom formatter.
I think logging should be a separate step unrelated to formatting.

lib/errors.js
```diff
@@ -74,13 +83,15 @@ function defaultErrorFormatter (err) {
     }).reduce((acc, val) => acc.concat(val), [])
   }
 
-  return {
+  const res = {
```
paolochiodi, line 86: can't we just return here?

mcollina:
> Seems like a side effect that may not be evident to users, and thus they'd fail to reimplement it when they use a custom formatter.

This come from experience. There is a significant amount of devs that asked for a similar feature to be _removed_ from Fastify. If I move it outside of the errorFormatter, I would need to add a new option. By putting it in the errorFormatter, I leave it to the user by default.

paolochiodi:
> > Seems like a side effect that may not be evident to users, and thus they'd fail to reimplement it when they use a custom formatter.
> 
> This come from experience. There is a significant amount of devs that asked for a similar feature to be _removed_ from Fastify. If I move it outside of the errorFormatter, I would need to add a new option. By putting it in the errorFormatter, I leave it to the user by default.

Yeah, but then you force devs that don't want the logs to rewrite the error formatter. To me the two features should just not be linked.

pscarey:
In terms of logging in the error formatter, we re-implemented the error formatter with logging & error capturing (i.e. Sentry) only for specific errors which are unrecognised.

I'd argue that the default error formatter is intended for development or basic requirements, but anyone who's running a production env will quite likely want to capture & report errors. Fastify itself also logs requests out.

Possibly the `errorFormatter` is could be re-concieved as e.g. a series of `onGraphQLError` hooks, but that's probably beyond the scope of this change.

I do get the argument that it could possibly be clearer, or there could be more options, but I'd also argue that it's still a net positive change this way,  and further updates can be made.

index.js
```diff
@@ -448,12 +448,13 @@ const plugin = fp(async function (app, opts) {
       operationName
     )
 
-    return maybeFormatErrors(execution, reply)
+    return maybeFormatErrors(execution, context)
   }
 
-  function maybeFormatErrors (execution, reply) {
+  function maybeFormatErrors (execution, context) {
```
stephenh, line 454: @mcollina could you update the index.d.ts definition of `errorFormatter`? Very coincidental timing on your PR but I'm _just_ now trying out/setting up fastify-gql and wanted to access the context from my error logging. :-) Thanks!

mcollina: If you have time to send a PR, it'd be quicker.

Issue #243, opened 2020-08-07
Switch to fake timers for test/subscription.js


mcollina:
As titled.

PabloSzx:
fixed in #254

PR #254, opened 2020-08-15
Switch to fake timers for test/subscription.js (#243)

meel-io:
Replacing `wait` calls with `tickAsync` from `FakeTimers`.

The issue doesn't have much description at the moment and it's possible that you had a different idea to achieve this. If that's the case, I'll be glad to get the feedback and change it accordingly.

Thanks!


mcollina:
great work!

PR #255, opened 2020-08-15
Fix errorFormatter's TypeScript typing.

stephenh:


stephenh:
@mcollina sure, I believe I've done that. Let me know more specifically what you'd like if that's not enough.

PR #256, opened 2020-08-15
feat: add asyncIterator & fix pubsub serialize/deserialize

gtolarc:
I added asyncIterator to the subscription context so that we can process multiple topics at once.

And I added json serialize/deserialize to the pubsub sending/receiving process. If we don't do this, a problem occurs when we perform a field-based permission check in middleware such as [graphql-shield](https://github.com/maticzav/graphql-shield).

I am not 100% sure that this PR will not be a problem. It would be good to apply after judging whether there is a problem with asynchronous processing or performance. @mcollina

### DIFF
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -49,8 +49,9 @@ class SubscriptionContext {
     if (typeof topics === 'string') {
       return this.pubsub.subscribe(topics, this.queue).then(() => this.queue)
     }
-    topics.forEach((topic) => this.pubsub.subscribe(topic, this.queue))
-    return this.queue
+    return Promise.all(topics.map((topic) => this.pubsub.subscribe(topic, this.queue).then(() => this.queue))).then(
+      (values) => values[0]
+    )
   }
 
   publish (event) {
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -10,7 +10,7 @@ class PubSub {
   subscribe (topic, queue) {
     return new Promise((resolve, reject) => {
       function listener (value, cb) {
-        queue.push(JSON.parse(value.payload))
+        queue.push(value.payload)
         cb()
       }
 
@@ -30,7 +30,6 @@ class PubSub {
   }
 
   publish (event, callback) {
-    event.payload = JSON.stringify(event.payload)
     this.emitter.emit(event, callback)
   }
 }
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -10,7 +10,7 @@ class PubSub {
   subscribe (topic, queue) {
     return new Promise((resolve, reject) => {
       function listener (value, cb) {
-        queue.push(value.payload)
+        queue.push(JSON.parse(value.payload))
         cb()
       }
 
@@ -30,6 +30,7 @@ class PubSub {
   }
 
   publish (event, callback) {
+    event.payload = JSON.stringify(event.payload)
     this.emitter.emit(event, callback)
   }
 }
@@ -47,6 +48,11 @@ class SubscriptionContext {
 
   subscribe (topic) {
     return this.pubsub.subscribe(topic, this.queue).then(() => this.queue)
+  }
+
+  asyncIterator (topics) {
+    topics.forEach((topic) => this.pubsub.subscribe(topic, this.queue))
+    return this.queue
   }
 
   publish (event) {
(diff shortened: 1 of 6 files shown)


gtolarc:
First of all, it was wrong to think that the problem was caused by middleware.
In my case, the "Maximum call stack size exceeded" error occurs in the following part.
https://github.com/mcollina/mqemitter-redis/blob/master/mqemitter-redis.js#L162

These are the issues related to this issue. When encoding circular structures, msgpack-lite has a problem.
https://github.com/kawanet/msgpack-lite/issues/63
https://github.com/kawanet/msgpack-lite/issues/74

I experienced an error when publish a mongoose model object, which is an object like this:
```
CoreMongooseArray [
        'transaction',
        [Symbol(mongoose#Array#_atomics)]: {},
        [Symbol(mongoose#Array#_path)]: 'scopes',
        [Symbol(mongoose#Array#_schema)]: [SchemaArray],
        [Symbol(mongoose#Array#_parent)]: [Circular] ]
```

As in the answer to the above issue, we should remove the circular structure by JSON.stringify.
I don't know if this is a good solution, but it works anyway. @mcollina

gtolarc:
https://github.com/davidyaha/graphql-redis-subscriptions/blob/master/src/redis-pubsub.ts#L78
Other pubsub implementations are also serializing json.

mcollina:
mqemitter is serializing as msgpack. There is no need to double serialize, just don't send Moongoose models as they are not serializable and they have a special thing for json.

gtolarc:
To be precise, the circular structure of the parent object is created in a specific situation in mongoose, but since the custom toJSON is well implemented mongoose internally, there is no error with JSON.stringify. However, custom serializers such as msgpack.encode don't call toJSON, so it seems to get the error.

As you said, don't send the mongoose object as it is, just run toJSON() on the object and publish it and it works fine. I will remove the JSON processing part and post PR.

lib/subscriber.js
```diff
@@ -49,6 +49,11 @@ class SubscriptionContext {
     return this.pubsub.subscribe(topic, this.queue).then(() => this.queue)
   }
 
+  asyncIterator (topics) {
```
mcollina, line 52: why is this called `asyncIterator`? Can't we just support arrays of topics in `subscribe`?

lib/subscriber.js
```diff
@@ -49,6 +49,11 @@ class SubscriptionContext {
...
+    topics.forEach((topic) => this.pubsub.subscribe(topic, this.queue))
```
mcollina, line 53: `subscribe` returns a promise, we should use a `Promise.all()`

lib/subscriber.js
gtolarc: I borrowed the naming of graphql-subscriptions. As far as I know, we should return asyncIterator to subscribe of GraphQLObjectType, and this is a naming meaning of custom asyncIterator.

However, as I try to add additional features like withFilter, it is not good to implement in the current PubSub and SubscriptionContext separate structure. So, as you say, it would be better to add only support arrays to existing subscribe function.

lib/subscriber.js
gtolarc: If I return only the redable queue without having to do promise.all, it would be fine, isn't it? Anyway, processing is asynchronous.

mcollina: it's not the same thing, those promises needs to be awaited.

gtolarc: If any error occurs, it has to stop, so promise.all seems to be correct as you say. I modified it...

lib/subscriber.js
```diff
@@ -49,8 +49,9 @@ class SubscriptionContext {
     if (typeof topics === 'string') {
       return this.pubsub.subscribe(topics, this.queue).then(() => this.queue)
     }
-    topics.forEach((topic) => this.pubsub.subscribe(topic, this.queue))
-    return this.queue
+    return Promise.all(topics.map((topic) => this.pubsub.subscribe(topic, this.queue).then(() => this.queue))).then(
+      (values) => values[0]
```
mcollina, line 53:
```suggestion
    return Promise.all(topics.map((topic) => this.pubsub.subscribe(topic, this.queue))).then(() => this.queue)
```

mcollina:
docs are needed

mcollina (approved):
lgtm

Issue #229, opened 2020-07-29
Federation: strip common directives from federate services schema


paolochiodi:
While working on https://github.com/mcollina/fastify-gql/pull/227 and https://github.com/mcollina/fastify-gql/issues/184 I have noticed that the fastify-gql implementation is lacking at least two features from Apollo server.

One of them is that apollo server strips common and type system directives definitions (i.e the definition of `key` directive) from federated services sdl before combining them. 

As stated in their documentation this allows the federated service to provide the schema unchanged without needing to strip them out.

Documentation: https://www.apollographql.com/docs/apollo-server/federation/gateway/#type-system-directives
Apollo's code: 
https://github.com/apollographql/apollo-server/blob/1c58ea3cbf6f5ead1afd5e81a1113a75441ead26/packages/apollo-federation/src/composition/composeAndValidate.ts#L16
https://github.com/apollographql/apollo-server/blob/main/packages/apollo-federation/src/composition/normalize.ts#L20
https://github.com/apollographql/apollo-server/blob/main/packages/apollo-federation/src/composition/normalize.ts#L270

paolochiodi:
Tests: https://github.com/apollographql/apollo-server/blob/main/packages/apollo-federation/src/composition/__tests__/normalize.test.ts#L147

PR #262, opened 2020-08-24
fix: extendSchema for federation

PacoDu:
This PR fixes an issue with `extendSchema` that wasn't working correctly with federated schema if we tried to extend the schema with an external type reference.

### DIFF
--- a/index.js
+++ b/index.js
@@ -24,7 +24,10 @@ const {
 } = require('graphql')
 const { buildExecutionContext } = require('graphql/execution/execute')
 const queryDepth = require('./lib/queryDepth')
-const buildFederationSchema = require('./lib/federation')
+const {
+  buildFederationSchema,
+  extendFederationSchema
+} = require('./lib/federation')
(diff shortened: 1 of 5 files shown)


PacoDu:
Do not merge !

service resolver sdl is not properly updated

mcollina:
Any update on this?

PacoDu:
We need a function to remove the federation metadata from the sdl (this is related to #229). I tried to remove the directives from the schema before using `printSchema` but this also removes the directive statement from the type definition (which we need to keep for the sdl). I don't have the time to investigate this right now (I'm merging my schemas before as a workaround for now).

Also, I think this part is wrong: https://github.com/mcollina/fastify-gql/pull/262/files#diff-d9185daf57698ed90d572d1966e61b16R235

If anyone wants to fix the issue feel free to use this PR or start over. I'll try to fix this later if it's not resolved.

mcollina:
Closing for now.

PR #265, opened 2020-08-26
fix: handle gateway.service.wsConnectionParams

PacoDu:
This PR fixes the `gateway.service.wsConnectionParams` that is currently defined in the documentation but not handled.

It also adds the `connection_error` message handling for the `SubscriptionClient` and the `connectionInitPayload` option.

### DIFF
--- a/lib/subscription-client.js
+++ b/lib/subscription-client.js
@@ -4,6 +4,7 @@ const WebSocket = require('ws')
 const {
   GQL_CONNECTION_INIT,
   GQL_CONNECTION_ACK,
+  GQL_CONNECTION_ERROR,
   GQL_START,
   GQL_DATA,
   GQL_COMPLETE,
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

PacoDu:
A little change so that `gateway.service.wsConnectionParams.connectionInitPayload` can be an async function.

lib/subscription-client.js
```diff
@@ -36,17 +38,20 @@ class SubscriptionClient {
...
+    this.connectionInitPayload = connectionInitPayload
 
     this.connect()
   }
 
   connect () {
     this.socket = new WebSocket(this.uri, this.protocols)
 
-    this.socket.onopen = () => {
+    this.socket.onopen = async () => {
```
mcollina, line 49: Can you add a try/catch? The `connectionInitPayload` can throw, and it will lead to an unhandledRejection

mcollina (changes requested):
Please add tests for the error paths where the new function throws.

PacoDu:
is there anything that prevent this PR from being merged ?

mcollina (approved):
nothing,

LGTM

PR #267, opened 2020-08-28
Fix websocket connection issues with postgraphile

jemikanegara:
Some backend federated services like postgraphile.org send periodic keepalive message over websocket.

This fixes compatibility with such graphql servers.

### DIFF
--- a/test/subscription-client.js
+++ b/test/subscription-client.js
@@ -289,3 +289,81 @@ test('subscription client closes the connection if connectionInitPayload throws'
     }
   })
 })
+
+test('subscription client sending empty object payload on connection init', (t) => {
+  const server = new WS.Server({ port: 0 })
+  const port = server.address().port
+
+  server.on('connection', function connection (ws) {
+    ws.on('message', function incoming (message) {
+      const data = JSON.parse(message)
+      if (data.type === 'connection_init') {
+        t.deepEqual(data.payload, {})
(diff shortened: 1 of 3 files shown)


mcollina (commented):
Can you please add a unit test? Also, CI is failing.

heri16:
Should be GQL_CONNECTION_KEEP_ALIVE

The complete list is here: https://github.com/apollographql/subscriptions-transport-ws/blob/faa219cff7b6f9873cae59b490da46684d7bea19/src/message-types.ts#L7

test/subscription-client.js
```diff
@@ -215,3 +215,81 @@ test('subscription client multiple subscriptions unsubscribe removes only one su
...
+        t.end()
+      }
+
+      const operationId1 = client.createSubscription('query', {}, publish)
+
+      setTimeout(() => {
+        // Let keep alive payload sent twice
+        client.unsubscribe(operationId1)
+      }, keepAliveInterval * 2)
+    }
```
mcollina, line 293: Would you mind to use fake-timers instead? There are other examples in the tests.

mcollina (changes requested):

mcollina:
this would also need a rebase

test/subscription-client.js
jemikanegara: done

mcollina (approved):
lgtm

PR #270, opened 2020-08-31
fix: handle SubscriptionClient ready state properly

PacoDu:
This PR fixes the ready state of the `SubscriptionClient` that was not properly handled due to an invalid property set in the `close` function (`this.isReady = false` instead of `this.ready = false`). 
It also extends the `connectionCallback` usage to handle error payload in case of a `connection_error`.

### DIFF
--- a/lib/subscription-client.js
+++ b/lib/subscription-client.js
@@ -28,6 +28,7 @@ class SubscriptionClient {
       maxReconnectAttempts = Infinity,
       serviceName,
       connectionCallback,
+      failedConnectionCallback,
       failedReconnectCallback,
       connectionInitPayload
     } = config
@@ -38,6 +39,7 @@ class SubscriptionClient {
     this.serviceName = serviceName
     this.reconnectAttempts = 0
     this.connectionCallback = connectionCallback
+    this.failedConnectionCallback = failedConnectionCallback
     this.failedReconnectCallback = failedReconnectCallback
     this.connectionInitPayload = connectionInitPayload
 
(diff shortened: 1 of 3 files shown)


PacoDu:
It also throws an error if createSubscription is called and the connection is not ready, otherwise the gateway was accepting gql_start without any error even if the connection_init failed with the related service.

mcollina (changes requested):
Is `connectionCallback` called when a new connection is established? Maybe calling it in case of an error is not expected, we might want another callback?

PacoDu:
It is called when a new connection is established and I thought it was a good fit to also handle the error case like a 'classic' callback with the error defined if it failed, but this can be a breaking change for some users if not handled. 
If you prefer I can declare a new callback but I think it makes the interface more complicated and it's not necessary.

PacoDu:
I've added the `failedConnectionCallback` to prevent a breaking change on `connectionCallback`.

lib/subscription-client.js
```diff
@@ -204,6 +206,9 @@ class SubscriptionClient {
         break
       case GQL_CONNECTION_ERROR:
         this.close(this.tryReconnect, false)
+        if (this.failedConnectionCallback) {
+          this.failedConnectionCallback(data.payload)
```
mcollina, line 210: I would add an await here.

PacoDu: Done, I've not handled the case were `failedConnectionCallback` throws because we don't have the fastify instance to log the error in the `SubscriptionClient`, this would require a modification on `buildServiceMap`

mcollina (approved):
lgtm

Issue #268, opened 2020-08-28
Federation: subscription protocol, websocket and authentication


PacoDu:
I'm experimenting with federation and subscriptions and I noticed a flaw in the current implementation.

### Problem
When the gateway starts and `services.wsUrl` is defined, the gateway will open a ws connection with each services and send the `connection_init` payload. 
Once a client creates a websocket connection and sends the `connection_init` payload, the gateway will instantiate a `SubscriptionConnection` and uses the gateway's `subscription.onConnect` handler. Now, if the Client wants to start a subscription with the `gql_start`  payload (defined by a federated service), the gateway will use the existing connection of the federated service and forward the `gql_start` payload and then it will forward any message coming from this subscription to the original client. This means that the federated service uses the same `SubscriptionConnection` for every gateway clients.

The issue here is that the service that defined this subscription and holds the logic/validation/permissions etc. didn't received the `connection_init` payload. Thus services have no way to authenticate and authorize messages coming from the gateway (the issue is the same if we use websocket headers as the connection between service and gateway was establish before any client request).

### Solution ?
I think that the solution would be to establish a connection from gateway to service per client: 

The gateway will not open websocket connections at startup, but only if needed. It can then forward the `connection_init` payload, initially received during the `client <-> gateway` handshake, to the related service. The federated service will create a `SubscriptionConnection` that is properly isolated per user. 
This solution will create more connections between gateway and services but it should allow services to properly handle authentication/authorization and isolation without modifying the subscription protocol.

Before trying any implementation I wanted to share this issue because I'm not sure if it's a good idea to store the `connection_init` payload in-memory to be able to establish new connections but I currently don't see any other way that wouldn't require modification on the service side or on the subscription protocol. This would also decrease the overall performance of the gateway as it should handle more connections.

mcollina:
Good spot.

I think there are two cases:

1. the server is aware that is talking to a gateway, and they negotiate the auth in some app-specific way - this is essentially how the gateway currently work. We might want to add some extension on this protocol to be able to trace the user details for each message. The connection from the gateway to the server is privileged.

2. The server is not aware that is talking to a gateway, in this case a new connection for each client must be performed (currently not supported). This is where your issue is about.

Would you like to add an option for keeping one connection on each front? The problem would be scaling such a deployment as it significantly increase memory consumption of the gateway. I'm not opposed on adding an option to support this second mode of operation.

PacoDu:
[...]

PR #271, opened 2020-08-31
feat: GraphQL over WebSocket protocol extensions and 'connectionInit' extension

PacoDu:
This PR is an implementation of the solution proposed in #268 that should solve the `connection_init` handshake issue.

### DIFF
--- a/test/gateway/connectionInit-extension.js
+++ b/test/gateway/connectionInit-extension.js
@@ -1,0 +1,327 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('fastify')
+const WebSocket = require('ws')
+const GQL = require('../..')
+
+test('connectionInit extension e2e testing', t => {
+  t.plan(5)
+  const userContext = { name: 'test-user' }
+
+  function onConnect (data) {
+    const { payload } = data
+    if (typeof payload.headers === 'object' && payload.headers.authorize) {
+      if (payload.headers.authorize) {
+        return { user: userContext }
+      }
+    }
+
+    return false
+  }
+
+  const userService = Fastify()
+  const notificationService = Fastify()
+  const gateway = Fastify()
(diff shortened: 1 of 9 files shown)


PacoDu:
Waiting for #270, otherwise I think this PR is ready.

mcollina (commented):
Good work! Can you please add a full end-to-end test where this is fully tested?

PacoDu:
> Good work! Can you please add a full end-to-end test where this is fully tested?

This is already included in test/gateway/subscription.js (https://github.com/mcollina/fastify-gql/pull/271/files#diff-b9cdf3fa0a5c08b4c70395f7f82a9930R388). Is there anything missing ?

mcollina:
I think we should have a test with two websocket clients to validate the end use case.

PacoDu:
I think this is ready now 🚀 !

test/gateway/connectionInit-extension.js
```diff
@@ -0,0 +1,327 @@
...
+      })
+
+      client.write(JSON.stringify({
+        type: 'connection_init',
+        payload: {
+          headers: {
+            authorize: true
+          }
+        }
+      }))
```
mcollina: Can you please add _another_ client that uses a different `authorize` header (let's use `foo` and `bar`)?

PacoDu: The part you commented is not up to date but sure i'll add a different auth method

PacoDu: I've added 3 authentication method: one for each client and an other for gateways. Do you want me to add a third client ?

PacoDu:
Is there anything else that prevent the merge of this PR ? I'm currently testing this version with some of our services and the `connectionInit` extension works really well.

mcollina (approved):
lgtm

PR #277, opened 2020-09-04
add playground settings option

MomenNano:
fixed [272](https://github.com/mcollina/fastify-gql/issues/272)

this RP is WIP, its missing documentation and testing (I'm not sure how to test it)

but I want to make sure I got the settings option name right.

### DIFF
--- a/index.js
+++ b/index.js
@@ -199,7 +199,7 @@ const plugin = fp(async function (app, opts) {
       errorHandler: opts.errorHandler,
       errorFormatter: opts.errorFormatter,
       ide: optsIde,
-      ideSettings: opts.playgroundOpts,
+      ideSettings: opts.playgroundSettings,
       prefix: opts.prefix,
       path: opts.path,
       context: opts.context,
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -199,6 +199,7 @@ const plugin = fp(async function (app, opts) {
       errorHandler: opts.errorHandler,
       errorFormatter: opts.errorFormatter,
       ide: optsIde,
+      ideSettings: opts.playgroundOpts,
```
mcollina, line 202: Could you please add the option to the README? I would call it `playgroundSettings` as it's not available for graphiql.

mcollina (commented):
You can add a new test where the plugin is configured with `playgroundSettings` and then you use the snapshot test to check this (as it is right now for `undefined`).

index.js
MomenNano: okay

mcollina (approved):
lgtm

MomenNano:
first I want to thank you very much for helping me with this PR <3

I did all the things you mentioned. I think it's ready to be reviewed.

Issue #252, opened 2020-08-14
Incorrect Error object returned


paul590:
Hello fastify team @RafaelGSS , the issue that i am seeing is when i run a query and receive an error it looks as follows:

```
{
  "error": {
   "data": null,
   "errors": [
   {
   "message": "error message thrown",
   "locations": [
   {
   "line": 2,
    "column": 3
    }
    ],
    "path": [
     "test"
    ]
   }
  ]
 }
}
```

however, if i run the same query again I see the error as below:

```
{
 "data": null,
 "errors": [
  {
  "message": "error message thrown",
  "locations": [
   {
   "line": 2,
   "column": 3
   }
   ],
   "path": [
   "test"
   ]
  }
 ]
}
```

Notice how there is an error object that wraps the actual error at first but then disappears with the cache of the error. The issue happens when using custom errors. is there a way to remove this error wrapping and have the library return the same error? Thank you so much for your help!! Here is the link to the open repo: https://github.com/paul590/toSendGraphQL

RafaelGSS:
ref: https://github.com/fastify/help/issues/227 I'll take a look later.

RafaelGSS:
@paul590 I could not reproduce your first example. Always return the same value (expected) for me:

```
{
  "data": {
    "test": null
  },
  "errors": [
    {
      "message": "Error test",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "test"
      ]
    }
  ]
}
```

However, I think I understand your other problem, is about your error handler, not works for `throw new Error...` but works when you pass a function that not already defined right? e.g: `query { anyFunctionThatNotExistsHere(code: "test") }` in your code.

Some considerations on that, I think that this is expected because following the [GraphQL - Error Spec](http://spec.graphql.org/draft/#sec-Errors) is a pattern to response errors in with this format. So, if your first example works as you sent, sounds a bug, but here I could not reproduce that.

As I can understand, you would like a `errorHandler` to runtime (e.g throw new Error). I recommend to check the types doc, maybe these two properties will help you:

```ts
  /**
   * Change the default error handler (Default: true).
   * If a custom error handler is defined, it should return the standardized response format according to [GraphQL spec](https://graphql.org/learn/serving-over-http/#response).
   * @default true
   */
  errorHandler?:
    | boolean
    | ((
        error: FastifyError,
        request: FastifyRequest,
        reply: FastifyReply
      ) => ExecutionResult);
  /**
[...]

PR #280, opened 2020-09-15
fix(validationErrors): call errorFormatter on validationErrors

RafaelGSS:
This is reference on #252.

I don't sure if this is expected behavior, but when occurring a validation error we don't call `errorFormatter`. But, is it not the only path that could occur, on the `queryDepthLimit` is another example.

So, the question is:

We should continue to throw errors on `BadRequest` and not call the `errorFormatter` to user treat it on `errorHandler` or call `errorFormatter` for everything? The `errorHandler` on the second option is less called or never.

### DIFF
--- a/test/gateway/pollingInterval.js
+++ b/test/gateway/pollingInterval.js
@@ -119,9 +119,7 @@ test('Polling schemas', async (t) => {
   t.deepEqual(JSON.parse(res2.body), {
     errors: [
       {
-        message:
-          'Cannot query field "lastName" on type "User". Did you mean "name"?',
-        locations: [{ line: 6, column: 13 }]
+        message: 'Cannot query field "lastName" on type "User". Did you mean "name"?'
       }
     ],
     data: null
@@ -361,9 +359,7 @@ test("Polling schemas (if service is down, schema shouldn't be changed)", async 
     t.deepEqual(JSON.parse(body), {
       errors: [
         {
-          message:
-            'Cannot query field "lastName" on type "User". Did you mean "name"?',
-          locations: [{ line: 6, column: 15 }]
+          message: 'Cannot query field "lastName" on type "User". Did you mean "name"?'
         }
       ],
       data: null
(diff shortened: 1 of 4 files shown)


mcollina:
CI is failing and tests are missing.

I'm not really sure about your question. I think we should pass through the formatter.

RafaelGSS:
I think there are no tests to include, just change the actual. I have a doubt about this "fix", seems like a major considering that could break clients that treat `validationError` on `errorHandler` and not on `errorFormatter`.

There are other paths that `throw err` instead of call `maybeFormatErrors`, is expected?

test/gateway/pollingInterval.js
```diff
@@ -119,9 +119,7 @@ test('Polling schemas', async (t) => {
   t.deepEqual(JSON.parse(res2.body), {
     errors: [
       {
-        message:
-          'Cannot query field "lastName" on type "User". Did you mean "name"?',
-        locations: [{ line: 6, column: 13 }]
```
mcollina, line 124: why are we losing `locations`? I think it was quite handy and needed.

mcollina (changes requested):

test/gateway/pollingInterval.js
RafaelGSS: The locations are set by `errorHandler` and it is not called anymore, instead we call `errorFormatter` that calls `defaultErrorFormatter` in our tests. Maybe we can set locations on `defaultErrorFormatter`?

mcollina: I think should move the locations to the formatter.

paul590:
@RafaelGSS @mcollina will this fix also include the ability for errors to also display the GraphQL Error extensions?

mcollina:
> will this fix also include the ability for errors to also display the GraphQL Error extensions?

Not sure, maybe it should be a different PR.

RafaelGSS:
Not sure as well. This PR only call `errorFormatter` instead of `errorHandler` on validationErrors.

RafaelGSS:
Today I had time to dedicate myself more to this and after a quick search, the behavior on issue is expected, and the affirmation on the description of PR

>  when occurring a validation error we don't call errorFormatter

Is wrong. We always call `errorFormatter` on `errorHandler` **if none errorHandler is provided**.

Reference:
https://github.com/mercurius-js/mercurius/blob/master/lib/routes.js#L114

Closing this PR. I'll open another just to clarify in our docs this behavior.

paul590:
@RafaelGSS @mcollina thank you for your help! Lastly, will you open the PR or would you like me to do so?

PR #282, opened 2020-09-18
Fix nested fragment bug

jemikanegara:
Missing 'schema' parameter inside recursive, this cause error: 

> Cannot read property 'getType' of undefined


mcollina (changes requested):

jemikanegara:

done, I just modify the schema test a little bit

mcollina (approved):
lgtm

Issue #272, opened 2020-08-31
[Playground] set header and dark theme setting


saostad:
hi,
is there any way to set Http headers in the playground?
also, is there any way to change the default light theme to a dark theme?

asciidisco:
I guess you're referring to the settings mentioned here: https://github.com/prisma-labs/graphql-playground#settings

There's currently no way to do this via the `fastify-gql` settings; channeling my inner @mcollina here, leads to the question:
"would you like to send a PR to fix/improve this?"

There are basically 2 code parts you would need to spend your attention on, if you'd like to integrate this:

https://github.com/mcollina/fastify-gql/blob/master/index.js#L196
https://github.com/mcollina/fastify-gql/blob/master/lib/routes.js#L307

blimmer:
Is this fully implemented as part of 5.7.0? The type definitions don't seem to support `playgroundOptions` as of 5.7 for me.

mcollina:
I think the types just need to be updated, want to send a PR?

blimmer:
@mcollina Yep, no problem. See #285.

PR #285, opened 2020-09-20
Add `playgroundSettings` to Typescript types.

blimmer:
Relates to #272.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -213,6 +213,30 @@ export interface FastifyGQLCommonOptions {
    * receive an array of responses within a single request.
    */
   allowBatchedQueries?: boolean;
+
+  /**
+   * Settings for GraphQL Playground. These settings only apply if `graphiql` parameter is set to 'playground'.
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -213,6 +213,30 @@ export interface FastifyGQLCommonOptions {
    * receive an array of responses within a single request.
    */
   allowBatchedQueries?: boolean;
+
+  /**
+   * Settings for GraphQL Playground. These settings only apply if `graphiql` parameter is set to 'playground'.
+   * The most current GraphQL Playground code is loaded via CDN, so new configuration settings may be available.
+   * See https://github.com/prisma-labs/graphql-playground#usage for the most up-to-date list.
+   */
+  playgroundSettings?: {
```
blimmer, line 222: Since `graphql-playground` is not currently included as a dependency in this repo's `package.json`, I went ahead and just copied over the settings from the upstream repo. The other alternative is to include `graphql-playground-react` as a dependency for the types. It'd need to be a `dependency` vs. a `devDependency` if folks want to go that route.

mcollina (approved):
lgtm

PR #289, opened 2020-09-25
Rename to mercurius

mcollina:
renamed to mercurius! Let me know if I missed anything

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-# fastify-gql
+# mercurius
 
 ![CI workflow](https://github.com/fastify/fastify-oauth2/workflows/CI%20workflow/badge.svg)
 
-Fastify GraphQL adapter.
+Mercurius is [__GraphQL__](https://graphql.org/) adapter for [__Fastify__](https://www.fastify.io)
 
 Features:
 
@@ -18,7 +18,7 @@ Features:
 ## Install
 
 ```
-npm i fastify fastify-gql
+npm i fastify mercurius
 ```
 
 ## Example
@@ -27,7 +27,7 @@ npm i fastify fastify-gql
 'use strict'
 
 const Fastify = require('fastify')
-const GQL = require('fastify-gql')
+const mercurius = require('mercurius')
 
 const app = Fastify()
 
@@ -43,7 +43,7 @@ const resolvers = {
   }
 }
 
-app.register(GQL, {
+app.register(mercurius, {
   schema,
   resolvers
 })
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -1285,6 +1285,16 @@ async function run () {
...
 
+## Acknowledgements
+
+The project is kindly sponsored by:
+
+* [NearForm](https://www.nearform.com) for [Matteo](https://github.com/mcollina)'s time in maintaining this module.
+
+The mercurius name was gracefully donated by [Marco Castelluccio](https://github.com/marco-c).
+The usage of that library was described in https://hacks.mozilla.org/2015/12/web-push-notifications-from-irssi/, and
+you can find that codebase in https://github.com/marco-c/mercurius.
```
mcollina, line 1296: @marco-c PTAL

marco-c: Thanks ;)

asciidisco:
the repos `About` section still contains `Fastify GraphQL adapter` 
Scrennshot 2020-09-25 um 11 27 43

mcollina:
@asciidisco what would you recommend as text?

asciidisco:
> @asciidisco what would you recommend as text?

I'm not exactly sure, the question is what will benefit the user most. I see it more like a `GraphQL server library based on Fastify` rather than a `Fastify GraphQL adapter`. But that's such a minor thing, I'm not sure if it's even worth the time to discusss TBH.

mcollina:
What do you think of..

> Implement GraphQL servers and gateways with Fastify

examples/subscription/mqemitter-mongodb-subscription.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const GQL = require('fastify-gql')
+const mercurius = require('fastify-gql')
```
PacoDu, line 3:
```suggestion
const mercurius = require('mercurius')
```

or `../..` to match the other examples

index.d.ts
```diff
@@ -239,12 +239,12 @@ export interface FastifyGQLCommonOptions {
   };
 }
 
-export type FastifyGQLOptions = FastifyGQLCommonOptions & (FastifyGQLGatewayOptions | FastifyGQLSchemaOptions)
+export type MercuriusOptions = MercuriusCommonOptions & (MercuriusGatewayOptions | MercuriusSchemaOptions)
 
 declare function fastifyGQL
```
frikille, line 244: `fastifyGQL` should be renamed to `mercurius` for consistency

frikille (commented):
In `make-resolver.js` line 10 a symbol is created with the name `fastify-gql.entity-resolvers` that should be renamed for `mercurius.entity-resolvers` for consistency.

Apart from these two changes, I think the rest is good 👍

asciidisco:
> What do you think of..
> 
> > Implement GraphQL servers and gateways with Fastify

That sounds very descriptive to me 👍

index.d.ts
mcollina: Sorry I missed this, I'm doing this change on master.

PR #292, opened 2020-09-28
fix(gateway): should support multiple extends of the same type in the service SDL

PacoDu:
The gateway can overwrite already defined types when calling `buildServiceMap`. 
I noticed the bug occurs if you define two Query returning a scalar.

I encountered this bug because I join the schema defined by multiple module before passing it to mercurius, so the service SDL contains multiple `extend Query`. Once the gateway schema is built the first `extend Query` doesn't have any resolver and skips the resolution to return `null` without forwarding the request to the related service.

### DIFF
--- a/lib/gateway/service-map.js
+++ b/lib/gateway/service-map.js
@@ -15,8 +15,8 @@ function hasDirective (directiveName, node) {
   return directives.some(directive => directive.name.value === directiveName)
 }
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

asciidisco:
I had to workaround this issue for the time beeing in my codebase, so I was able to test it with an existing application, works as expected.

PR #293, opened 2020-09-29
doc(README): add note about errorHandler

RafaelGSS:
Reference: https://github.com/mercurius-js/mercurius/pull/280#issuecomment-700358125﻿


mcollina (approved):
lgtm

PR #302, opened 2020-10-06
feat: defineResolvers should throw if field is not defined in schema

PacoDu:
`mercurius.defineResolvers` can fail to bind a resolver if it is not defined in the schema, this PR fails properly with the type and field name to help for debug


mcollina (approved):
lgtm

PR #307, opened 2020-10-08
feat: support extensions from query for GET requests

sastan:
This PR allows to use automatic persisted queries using GET requests.


mcollina (approved):
lgtm

PR #308, opened 2020-10-08
test(gateway): support array references and empty reponse

PacoDu:
An empty `_entities` response for `LIST_TYPE` currently produces the following error:  `Cannot read property 'json' of undefined`.

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -107,12 +107,25 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
             } else {
               // check if the field is default field of the type
               if (serviceMap[serviceForType].typeMap[type.name].has(fieldName)) {
-                field.resolve = makeResolver({
-                  service: serviceMap[serviceForFieldType],
-                  createOperation: createEntityReferenceResolverOperation,
-                  transformData: makeEntityTransformData(field),
-                  isReference: true
-                })
+                // Check if field is nullable
+                const isNonNull = field.astNode.type.kind === Kind.NON_NULL_TYPE
+                const leafKind = isNonNull ? field.astNode.type.type.kind : field.astNode.type.kind
+
+                if (leafKind === Kind.LIST_TYPE) {
+                  field.resolve = makeResolver({
+                    service: serviceMap[serviceForFieldType],
+                    createOperation: createEntityReferenceResolverOperation,
+                    transformData: response => response ? response.json.data._entities : (isNonNull ? [] : null),
+                    isReference: true
+                  })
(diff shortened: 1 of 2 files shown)


lib/gateway.js
```diff
@@ -146,6 +144,25 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
   }
 }
 
+function makeEntityTransformData (field) {
+  const isNonNull = field.astNode.type.kind === Kind.NON_NULL_TYPE
+  const leafKind = isNonNull ? field.astNode.type.type.kind : field.astNode.type.kind
```
PacoDu, line 149: There may be a better/more generic way to find theses values but for now it solves the issue nullable/not nullable lists

lib/gateway.js
```diff
@@ -146,6 +144,25 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
...
+        return isNonNull ? [] : null
+      }
+
+      return response.json.data._entities
+    }
+  }
+
+  return function transformData (response) {
+    return response.json.data._entities[0]
+  }
```
mcollina, line 163: This function can be a top level one, as there is no need to allocate a closure.

lib/gateway.js
```diff
@@ -146,6 +144,25 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
...
+  const leafKind = isNonNull ? field.astNode.type.type.kind : field.astNode.type.kind
+
+  if (leafKind === Kind.LIST_TYPE) {
+    return function transformData (response) {
+      if (!response) {
+        return isNonNull ? [] : null
+      }
+
+      return response.json.data._entities
+    }
```
mcollina, line 158: I would prefer if we did not create a closure over the ast node, can you move this to two top level functions?

PacoDu:
related to #310

PacoDu:
Strange, idk why the CI failed here, it's working locally

mcollina (approved):
lgtm

test/gateway/schema.js
```diff
@@ -865,3 +865,281 @@ test('Should support multiple `extends` of the same type in the service SDL', as
...
+        },
+        {
+          name: 'user',
+          url: `http://localhost:${userServicePort}/graphql`
+        }
+      ]
+    }
+  })
+
+  await gateway.listen(0)
```
mcollina, line 1103: You can remove this, there is no need as you are using inject()

mcollina (approved):
lgtm

Issue #309, opened 2020-10-08
Replace JSON.parse with secure-json-parse


mcollina:
We do a few JSON.parse() that we should replace with https://github.com/fastify/secure-json-parse. It's not strictly a security issue, but it would be great to do it to protect against possible attacks.

lucadonnoh:
working on this

PacoDu:
Resolved by https://github.com/mercurius-js/mercurius/pull/311

PR #311, opened 2020-10-08
replaced JSON.parse with secure-json-parse across files

lucadonnoh:
#309 using the default configuration (throw a SyntaxError)


mcollina (commented):
There was no need to change the tests, could you restore them?

lucadonnoh:
yes

lucadonnoh:
hmm...

mcollina:
you can try adding a few `console.log(err)` in the try-catch blocks, maybe there is an error being caught there.

jkirkpatrick24:
Looks like an error is being thrown by `secure-json-parse` when attempting to parse a buffer [here](https://github.com/mercurius-js/mercurius/pull/311/files#diff-2050061250e7c969e3895288b8da74b5340874e78239e4a82806cf6f1ecc195eR170)

The error being thrown is `TypeError: text.charCodeAt is not a function` 

I've opened up a [PR in secure-json-parse](https://github.com/fastify/secure-json-parse/pull/13) that should hopefully fix this issue 😄

mcollina:
A better approach is to call `.toString()` before trying to parse with a `Buffer`.

jkirkpatrick24:
@lucadonnoh secure-json-parse has been updated. I think you should be able to update to the latest version and re-open this.

mcollina (approved):
lgtm

PR #313, opened 2020-10-09
fix(gateway): make parent field use @requires directive

jimmymintzer:
When extending a type defined in a separate service, the fields in the `@requires` directive are not being added to the query unless they are fetched by the client already.

This change looks at all the children of a field and adds any required fields to the parent's selection set

Example:
```
  id: ID! @external
  name: String @external
  biography: String @requires(fields: "id name")
}
```
Currently, only `id` will be passed to `biography` because it's defined on `@key` (or if it's part of the client's query)

With this change, both `id` & `name` will be passed to `biography`

### DIFF
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -30,19 +30,21 @@ function getDirectiveSelection (node, directiveName) {
 }
 
 function removeNonServiceTypeFields (selections, service, type, schema) {
-  const requiredFields = []
-
   return [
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PacoDu (approved):
lgtm

asciidisco (approved):
I've ran the branch against https://github.com/MichalLytek/type-graphql/blob/20c81f4f7ee82779595002d25784fce3a8ff8b9b/examples/apollo-federation/ with the default query given in the repo (which always failed because of the `requires` directive in `inventory/resolver.ts`) but now it's working flawlessly. Great job (and thank you @jimmymintzer)

PR #317, opened 2020-10-14
feat(gateway): add initHeaders option for services

PacoDu:
This PR adds an option to set request headers for the _service SDL query from gateway to services, similar to connectionInitPayload.

### DIFF
--- a/lib/gateway/service-map.js
+++ b/lib/gateway/service-map.js
@@ -99,6 +99,13 @@ async function buildServiceMap (services) {
         }
       },
       async init () {
+        let headers
+        if (typeof initHeaders === 'function') {
+          headers = await initHeaders()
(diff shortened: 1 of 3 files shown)


lib/gateway/service-map.js
```diff
@@ -67,7 +67,7 @@ async function buildServiceMap (services) {
   const serviceMap = {}
 
   for (const service of services) {
-    const { name, mandatory = false, ...opts } = service
+    const { name, mandatory = false, initHeaders, ...opts } = service
```
mcollina, line 70: Can you please document this new option?

mcollina (changes requested):

mcollina (approved):
lgtm

PR #321, opened 2020-10-16
Add schemaTransforms and transformSchema utilities

mcollina:


package.json
```diff
@@ -31,6 +31,7 @@
     "@typescript-eslint/parser": "^3.6.1",
     "autocannon": "^6.0.0",
     "fastify": "^3.0.2",
+    "graphql-constraint-directive": "^2.1.0",
```
asciidisco, line 34: Question: Is it necessary to pull in a new 3rd party package to test the functionality?

index.js
```diff
@@ -340,6 +344,16 @@ const plugin = fp(async function (app, opts) {
     fastifyGraphQl.defineResolvers(resolvers)
   }
 
+  fastifyGraphQl.transformSchema = function (transforms) {
+    if (!Array.isArray(transforms)) {
+      transforms = [transforms]
+    }
+
+    for (const fn of transforms) {
+      fastifyGraphQl.replaceSchema(fn(fastifyGraphQl.schema))
```
asciidisco, line 353: `fn` seems oddly unspecifc, I believe it would help with the readability of the code if that would be renamed to `transformFn` or something like that.

PacoDu (approved):

PR #305, opened 2020-10-07
setup docsify

AlexNiallMalin:
@mcollina  Let me know if there's anything else to add.

PR #323, opened 2020-10-17
implement docsify

PabloSzx:
Preview here: https://pabloszx.github.io/mercurius/#/

Alternative of #305 following the existing file structure


mcollina:
The other PR had a few more css styles in it? Can we meld the two together? cc @AlexNiallMalin

PabloSzx:
> The other PR had a few more css styles in it? Can we meld the two together? cc @AlexNiallMalin

@mcollina done 👍 

Also added light / dark theme based on OS (it seems to fallback to light theme), search plugin & copy to clipboard plugin.

mcollina (approved):
lgtm

PR #326, opened 2020-10-18
feat(subscription): implement subscription filters

PacoDu:
This PR adds an utility function for filtered subscriptions, inspired by apollo's [withFilter](https://www.apollographql.com/docs/apollo-server/data/subscriptions/#subscription-filters)

### DIFF
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -84,6 +84,7 @@ function withFilter (subscribeFn, filterFn) {
           yield payload
         }
       } catch (err) {
+        context.app.log.error(err)
         continue
       }
     }
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -75,7 +75,23 @@ class SubscriptionContext {
   }
 }
 
+function withFilter (subscribeFn, filterFn) {
+  return async function * (root, args, context, info) {
+    const subscription = await subscribeFn(root, args, context, info)
+    for await (const payload of subscription) {
+      try {
+        if (await filterFn(payload, args, context, info)) {
+          yield payload
+        }
+      } catch (err) {
+        continue
+      }
+    }
+  }
+}
+
 module.exports = {
   PubSub,
-  SubscriptionContext
+  SubscriptionContext,
+  withFilter
 }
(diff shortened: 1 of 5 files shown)


test/subscription.js
```diff
@@ -1097,3 +1097,309 @@ test('subscription works properly if onConnect is not defined and connectionInit
...
+    },
+    Subscription: {
+      notificationAdded: {
+        subscribe: withFilter(
+          (_, __, { pubsub }) => pubsub.subscribe('NOTIFICATION_ADDED'),
+          (payload, { contains }) => {
+            if (contains && !payload.notificationAdded.message.includes(contains)) {
+              throw new Error('fail')
+            }
+            return true
```
mcollina, line 1310: Can you add a test where this returns a Promise? I'm not really sure this should be `async`, it might just be a synchronous implementation instead.

lib/subscriber.js
```diff
@@ -75,7 +75,23 @@ class SubscriptionContext {
...
 
+function withFilter (subscribeFn, filterFn) {
+  return async function * (root, args, context, info) {
+    const subscription = await subscribeFn(root, args, context, info)
+    for await (const payload of subscription) {
+      try {
+        if (await filterFn(payload, args, context, info)) {
+          yield payload
+        }
+      } catch (err) {
```
mcollina, line 86: I think we should log this error.

mcollina (changes requested):

test/subscription.js
PacoDu: If we go for a synchronous implementation, I think this is achievable with a stream Transform. But IMO this should be async if some async stuff is required to filter the notification (apollo's implementation allows async filter). I'll add an async test to see if it works as expected.

mcollina: Unfortunately a stream transform is slower than an async iterator, so better go with the async iterator anyway.

mcollina: Let's add a test for the async version

PacoDu: > Unfortunately a stream transform is slower than an async iterator, so better go with the async iterator anyway.

Ok, good to know ^^

mcollina (approved):
lgtm

PacoDu:
@mcollina can we have a minor bump for this feature please ?

PR #329, opened 2020-10-19
Add GrtaphQl extensions

paul590:
Hi Matteo,

As per our conversation I uploaded a version which fails when getting a 400 error. As you can see the error is already coming without the extensions portion that GraphQL adds to the error.

### DIFF
--- a/test/errors.js
+++ b/test/errors.js
@@ -7,6 +7,63 @@ const { ErrorWithProps } = GQL
 const { FederatedError } = require('../lib/errors')
 const split = require('split2')
 
+test('errors - incorrect query error', async (t) => {
+  const schema = `
+    type Query {
+      error: String
+      successful: String
+    }
+  `
+
+  const resolvers = {
+    Query: {
+      error (err) {
+        console.log(JSON.stringify(err))
+        throw new ErrorWithProps('Error', { code: 'ERROR', additional: 'information', other: 'data' })
+      },
+      successful () {
+        return 'Runs OK'
+      }
+    }
+  }
+
+  const app = Fastify()
+
+  app.register(GQL, {
+    schema,
+    resolvers
+  })
+
+  await app.ready()
+
+  const res = await app.inject({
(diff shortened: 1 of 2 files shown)


test/errors.js
```diff
@@ -7,6 +7,63 @@ const { ErrorWithProps } = GQL
...
+  app.register(GQL, {
+    schema,
+    resolvers
+  })
+
+  await app.ready()
+
+  const res = await app.inject({
+    method: 'GET',
+    url: '/graphql?querys={error,successful}'
```
mcollina, line 41: You have a typo here. You are performing a non-existent query. After you perform the correct query, the extensions are correctly reported.

```suggestion
    url: '/graphql?query={error,successful}'
```

mcollina:
Closing as it does not seem an actual bug.

paul590:
Hi @mcollina Thank you for looking into this. I did the typo on purpose to show that when an error occurs the extensions portion is not getting filled. Even though this is a graphql validation error it does not have the extensions variable. At some point in the code it seems we are removing the extension from all errors. by the time it reaches mercurius error handler the extensions does not exist in the error anymore. I added a print statement on the error.js file defaultErrorFormatter function to verify how the error is coming.

mcollina:
I don't understand where the extensions are supposed to come from in your example. You are allocating them in https://github.com/mercurius-js/mercurius/pull/329/files#diff-5cb4ab11a7c5fd6f646ea7e6fb3ff28c0eea4ff9b5bb4431513c0c2c2b2ca766R22 which is never called.

paul590:
@mcollina I could be misunderstanding, I believe GraphQL errors when they occur inject extensions to their errors. The application seems to get this error at some point and remove the extensions?

mcollina:
The handling of syntax errors happens https://github.com/mercurius-js/mercurius/blob/4f00f60a321f1afc5b2008376149831e6980af5e/index.js#L388-L394. They have no extensions.

PR #330, opened 2020-10-19
doc(README): add Federated subscriptions support

PacoDu:
I think we should add this feature to the README, I don't know about any other graphql server that is capable of this feature for now.


mcollina (approved):
lgtm

PR #331, opened 2020-10-19
Move to graphql-tools scope packages

Urigo:
Not a big deal but can save some bytes in the install scripts.
Also changed the docs as it's a better practice


mcollina (approved):
lgtm

PR #334, opened 2020-10-21
feat(loaders): expose info field

PacoDu:
This PR tries to expose the info field to the loader function. This is currently not working because it broke the caching strategy. 

@mcollina could we have a way to ignore the info key in `single-user-cache` ?

Use case: I want to access the info field of each query in my loader to find the minimal merged projection required for this batch (i.e.   `const projection = extend(true, {}, ...queries.map(({ info }) => introspectFields(info)))` )

### DIFF
--- a/index.js
+++ b/index.js
@@ -319,11 +319,11 @@ const plugin = fp(async function (app, opts) {
 
     function defineLoader (name) {
       // async needed because of throw
-      return async function (obj, params, { reply }) {
+      return async function (obj, params, { reply }, info) {
         if (!reply) {
           throw new Error('loaders only work via reply.graphql()')
         }
-        return reply[kLoaders][name]({ obj, params })
+        return reply[kLoaders][name]({ obj, params, info })
       }
     }
 
--- a/test/loaders.js
+++ b/test/loaders.js
@@ -64,23 +64,23 @@ test('loaders create batching resolvers', async (t) => {
   const loaders = {
     Dog: {
       async owner (queries, { reply }) {
+        const objs = queries.map(({ obj }) => obj)
+        const params = queries.map(({ params }) => params)
+
         // note that the second entry for max is cached
-        t.deepEqual(queries, [{
-          obj: {
-            name: 'Max'
-          },
-          params: {}
-        }, {
-          obj: {
-            name: 'Charlie'
-          },
-          params: {}
-        }, {
-          obj: {
-            name: 'Buddy'
-          },
-          params: {}
-        }])
+        t.deepEqual(objs, [
+          { name: 'Max' },
+          { name: 'Charlie' },
+          { name: 'Buddy' }
+        ])
+        t.deepEqual(params, [
+          { },
+          { },
+          { }
+        ])
+
+        queries.map(({ info }) => t.true(typeof info === 'object'))
+
         return queries.map(({ obj }) => owners[obj.name])
       }
     }
@@ -135,28 +135,25 @@ test('disable cache for each loader', async (t) => {
     Dog: {
       owner: {
         async loader (queries, { reply }) {
+          const objs = queries.map(({ obj }) => obj)
+          const params = queries.map(({ params }) => params)
+
           // note that the second entry for max is NOT cached
-          t.deepEqual(queries, [{
-            obj: {
-              name: 'Max'
-            },
-            params: {}
-          }, {
-            obj: {
-              name: 'Charlie'
-            },
-            params: {}
-          }, {
-            obj: {
-              name: 'Buddy'
-            },
-            params: {}
-          }, {
-            obj: {
-              name: 'Max'
-            },
-            params: {}
-          }])
+          t.deepEqual(objs, [
+            { name: 'Max' },
+            { name: 'Charlie' },
+            { name: 'Buddy' },
+            { name: 'Max' }
+          ])
+          t.deepEqual(params, [
+            {},
+            {},
+            {},
+            {}
+          ])
+
+          queries.map(({ info }) => t.true(typeof info === 'object'))
+
           return queries.map(({ obj }) => owners[obj.name])
         },
         opts: {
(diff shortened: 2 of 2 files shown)


mcollina:
> @mcollina could we have a way to ignore the info key in single-user-cache ?

I don't understand this, can you please articulate it?

PacoDu:
> > @mcollina could we have a way to ignore the info key in single-user-cache ?
> 
> I don't understand this, can you please articulate it?

I did not dive into `single-user-cache` source code but I guess the cache key is based on the query keys (obj and params for mercurius loaders). The issue is that `info` is a heavy object that can differ between queries.

See: https://github.com/mercurius-js/mercurius/pull/334/files#diff-ffa2d5d0109b010a154b14c64adad546b8bafd660d4dfdf877c8550253d26adcR74 
This test is not working currently, `{ name: 'Max' }` is not cached. 
CI line: https://github.com/mercurius-js/mercurius/pull/334/checks?check_run_id=1287125281#step:5:497

PacoDu:
```
# Subtest: loaders create batching resolvers
        not ok 1 - should be equivalent
          ---
          diff: |
            --- expected
            +++ actual
            @@ -8,4 +8,7 @@
               Object {
                 "name": "Buddy",
               },
            +  Object {
            +    "name": "Max",
            +  },
             ]
          at:
            line: 71
            column: 11
            file: test/loaders.js
            type: Wrapper
            function: owner
            method: func
          stack: |
            Wrapper.owner [as func] (test/loaders.js:71:11)
            node_modules/single-user-cache/index.js:87:12
          source: |2
                    // note that the second entry for max is cached
                    t.deepEqual(objs, [
            ----------^
                      { name: 'Max' },
                      { name: 'Charlie' },
          ...
```

PacoDu:
Here is the key generated for the 1st test by single-user-cache:

```js
{
  obj: { name: 'Max' },
  params: {},
  info: {
    fieldName: 'owner',
    fieldNodes: [ [Object] ],
    returnType: Human,
    parentType: Dog,
    path: { prev: [Object], key: 'owner', typename: 'Dog' },
    schema: GraphQLSchema {
      __validationErrors: [],
      description: undefined,
      extensions: undefined,
      astNode: undefined,
      extensionASTNodes: [],
      _queryType: Query,
      _mutationType: undefined,
      _subscriptionType: undefined,
      _directives: [Array],
      _typeMap: [Object: null prototype],
      _subTypeMap: [Object: null prototype] {},
      _implementationsMap: [Object: null prototype] {}
    },
    fragments: [Object: null prototype] {},
    rootValue: {},
    operation: {
      kind: 'OperationDefinition',
      operation: 'query',
      name: undefined,
      variableDefinitions: [],
      directives: [],
      selectionSet: [Object],
      loc: [Object]
    },
    variableValues: {}
  }
}
```

The info field is part of the key and it doesn't not match between representations of the same graphql Type and key

mcollina:
Why do you need the `info` object in the loader?

I'm not really convinced this is a bug after all. Essentially the `info` object will  have different data depending on where the loader is used in the query graph - by caching things we will lose every access after the first. The `info` object was not passed around on purpose.

---

The cache key is just the list of ids you pass for loading. It is normalized to a string in https://github.com/mcollina/single-user-cache/blob/ce555fef21168ebcfd68e431b9dc235173fe6fb2/index.js#L59. As you see, I used a stable json stringification algorithm. This becomes a problem if you pass the `info` object around because that's specific to _where_ the loader is called, which means it makes it impossible for things to be cached.

There is probably something to change/fix there and separate the cache key from the passed in object in https://github.com/mcollina/single-user-cache/blob/ce555fef21168ebcfd68e431b9dc235173fe6fb2/index.js#L58-L59. This would allow us to pass the `info` object around and cache only the first access. We should also add another option to disable the cache altogether but keep the coalescing effect.

PacoDu:
> I'm not really convinced this is a bug after all. Essentially the `info` object will have different data depending on where the loader is used in the query graph - by caching things we will lose every access after the first. The `info` object was not passed around on purpose.
>
> The cache key is just the list of ids you pass for loading. It is normalized to a string in https://github.com/mcollina/single-user-cache/blob/ce555fef21168ebcfd68e431b9dc235173fe6fb2/index.js#L59. As you see, I used a stable json stringification algorithm. This becomes a problem if you pass the `info` object around because that's specific to _where_ the loader is called, which means it makes it impossible for things to be cached.

Yes, this is why I didn't want the info to be used to define the key for the stringification. I was looking for a way to know the projection to avoid over fetching from the database the fields that will not be used in the query. But you are right this does not fit the loader usage ... 

Let me elaborate my use case and why I was looking for a way to know the projection:

```js
const schema =`
type Data @key(fields: "id"){
 id
 name
 nested: Nested
}

type Nested {
  prop: String
  var: Int
}`

const loaders = {
  Data: {
    async __resolveReference: async (queries, params) => {
       // It would be nice to avoid fetching unnecessary fields here
       // const projection = introspectField(info)
       const ids = queries.map(({ obj }) => obj.id)
       const objs = await context.service.search({ ids }, projection)
       return objs
    }
  }
}

const resolver = {
  Data: {
    async __resolveReference (reference, args, context, info) {
      // This is highly inefficient
       const projection = introspectField(info)
       const obj = await context.service.read(reference.id, projection)
       return obj
     }
  }
}
```

But I did not tough to circular references at first ... I'll have to test this but, if we use this kind of schema:
```js

const schema =`
type Data @key(fields: "id"){
 id
 name
 nested: Nested
}

type Nested {
  prop: String
  var: Int
  data: Data
}`

const loaders = {
  Data: {
    async __resolveReference: async (queries, params) => {
       // It would be nice to avoid fetching unnecessary fields here
       // const projection = introspectField(info)
       const ids = queries.map(({ obj }) => obj.id)
       const objs = await context.service.search({ ids }, projection)
       return objs
    }
  }
}

const resolver = {
  Data: {
    async __resolveReference (reference, args, context, info) {
      // This is highly inefficient
       const projection = introspectField(info)
       const obj = await context.service.read(reference.id, projection)
       return obj
     }
  }
}

```

 I guess it will not work because the loader will be called multiple times and the projection may vary for the same key...

PacoDu:
Closing, for now I don't have the time to investigate further.

mcollina:
I think the use case is valid - you need a flag to to single-user-cache to disable the caching and keep the coalescing.

PR #335, opened 2020-10-22
fix(gateway): handle interface referenced multiple times in services

PacoDu:


PacoDu:
The gateway was messing with interfaces due to the way it was handling Enums. They were treated as a Type but an Enum does not need a resolver it has to be defined in every services which requires it: https://www.apollographql.com/docs/federation/value-types/

mcollina (approved):
lgtm

PR #340, opened 2020-10-26
improve subscription types

PabloSzx:
- The included `pubsub.subscribe` actually returns a Node.js Readable (technically from `readable-stream`, but both are pretty much the same types)
- `withFilter` types were missing

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -17,9 +17,10 @@ import {
 } from "graphql";
 import { SocketStream } from "fastify-websocket"
 import { IncomingMessage, IncomingHttpHeaders, OutgoingHttpHeaders } from "http";
+import { Readable } from "stream";
 
 export interface PubSub {
-  subscribe<TResult = any>(topics: string | string[]): Promise<AsyncIterator<TResult>>;
+  subscribe<TResult = any>(topics: string | string[]): Promise<Readable & AsyncIterableIterator<TResult>>;
   publish<TResult = any>(event: { topic: string; payload: TResult }, callback?: () => void): void;
 }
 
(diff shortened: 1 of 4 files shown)


index.d.ts
```diff
@@ -17,9 +17,10 @@ import {
 } from "graphql";
 import { SocketStream } from "fastify-websocket"
 import { IncomingMessage, IncomingHttpHeaders, OutgoingHttpHeaders } from "http";
+import { Readable } from "stream";
 
 export interface PubSub {
-  subscribe<TResult = any>(topics: string | string[]): Promise<AsyncIterator<TResult>>;
+  subscribe(topics: string | string[]): Promise<Readable>;
```
mcollina, line 23: Most people does not know that a Readable is an asyncIterator :/.

mcollina (commented):
+1, just a note.

index.d.ts
PabloSzx: I was playing with the types and it seems like `Readable & AsyncIterable<TResult>` seems to be a good middle-ground, while allowing to access the readable functions & keeping the async iterable stuff with the data type

https://codesandbox.io/s/summer-butterfly-xgxdv?file=/src/index.ts

mcollina: Agree!

PabloSzx:
I had to add TypeScript as dev dep because of `Error: Failed to load parser '@typescript-eslint/parser' declared in 'CLIOptions': Cannot find module 'typescript'` was showing up 🤷‍♂️ 

And to test async iterable stuff properly, the target should be at least set `es2018`

mcollina (approved):
lgtm

tsconfig.json
```diff
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "es2018",
```
mcollina, line 3: why this change?

PabloSzx: The async iterable stuff comes with typescript es2018, without it the tests won't recognize it's keywords

Issue #179, opened 2020-06-07
Feature Request: Tracing support


tHBp:
Consider adding tracing support to enable checking runtime performance of queries and downstream API/Database calls.

Apollo supports the same [via a plugin](https://github.com/apollographql/apollo-server/tree/master/packages/apollo-tracing).

It can be either directly part of the core, or be implemented as a plugin. But for the latter, I guess we'd have to roll out a plugin architecture.

mcollina:
Would you like to work on this? I think it would be a nice feature to add.

tHBp:
Sure, will try to complete a first revision over the weekend.

xamgore:
Precious information can be found in the [apollo-tracing](https://github.com/apollographql/apollo-tracing) repository.

pscarey:
@xamgore @tHBp

I'd like to flag that the only solution here isn't just Apollo's approach, but also the tracing could be implemented with [OpenTelemetry](https://opentelemetry.io/). Adding an OpenTelemetry exporter would be a viable way of supporting many more tracing backends in a performant way.

Wrapping all resolvers in a Schema to achieve detailed tracing (by analysing the GraphQL `info` object which includes `type`, `path`), is viable, but can have *large* performance impacts. We have implemented this approach on certain APIs for advanced Audit Logging, so it's worth the hit, but ensuring performance is not overly impacted is key for tracing.

A viable approach here may be:
- Wrapping and timing only the key schema/resolvers (i.e. where a `resolve` function has been provided).
- Only tracing execution time a configurable percentage of the time.

It might be that timing every resolver is possible, but performance impacts should be kept in mind (Apollo is certainly able to achieve this in Apollo Engine / Graph Manager).

I can provide some additional info on how you might go about wrapping the schema if either of you is interested in this.

mcollina:
I would actually prefer adding OpenTelemtry support.

PacoDu:
I think we will just need some documentation that explains how to use opentelemetry with mercurius. Some packages are about to be released for graphql instrumentation: [@opentelemetry/instrumentation-graphql](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/master/plugins/node/opentelemetry-instrumentation-graphql). I've not tested this package yet.

PacoDu:
> Could we ship that as a small module with a few tests in this org?

[...]

PR #341, opened 2020-10-31
doc(opentelemetry): add tracing documentation with zipkin and opentelemetry

PacoDu:
Waiting for `@opentelemetry/instrumentation-graphql` release.

### DIFF
--- a/docs/integrations/README.md
+++ b/docs/integrations/README.md
@@ -4,3 +4,4 @@
 
 - [@nexus/schema](/docs/integrations/nexus-schema.md) - Declarative, code-first and strongly typed GraphQL schema construction for TypeScript & JavaScript
 - [mercurius-integration-testing](/docs/integrations/mercurius-integration-testing.md) - Utility library for writing mercurius integration tests.
+- [@opentelemetry](/docs/integrations/open-telemetry) - A framework for collecting traces and metrics from applications.
--- a/docs/integrations/open-telemetry.md
+++ b/docs/integrations/open-telemetry.md
@@ -60,7 +60,7 @@ const opentelemetry = require('@autotelic/fastify-opentelemetry')
 service.register(opentelemetry, { serviceName })
 service.register(mercurius, {
   schema: `
-  type Query {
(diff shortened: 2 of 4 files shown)


PabloSzx:
Maybe it's better for `open-telemetry.md` to be in `docs/integrations/`

PacoDu:
Closes #179

PacoDu:
TODO: switch to @autotelic/fastify-opentelemetry

mcollina:
@adamovittorio PTAL

adamovittorio:
@projectjudge FYI, we can try it with the gcp/prometheus exporter and see if we can remove the custom code for the headers propagation in fastify. Then in the ctx we have access to the tracer that we can use to instantiate custom spans.

projectjudge:
@adamovittorio for sure, we can give it a go and update the decorator for our custom spans as well if needed

PacoDu:
I switched to `@autotelic/fastify-opentelemetry` for the documentation. 

It's not perfect, here is the trace I get with the configuration described in the documentation:

For the `_service { sdl }` requested by the `gateway` from `service-add` I've a clean trace: 
image

But for the example query `{ add(x: 1, y: 2)`:
image

PacoDu:
Ok ... the gateway doesn't work as expected with a query returning a Scalar, it returns null. Modifying the schema to return an object fixed the issue.

Opentelemetry is perfectly working 🎉 :
image

mcollina (approved):
lgtm

mcollina:
I think we should document that this would not work with graphql-jit. Or does it?

PacoDu:
> I think we should document that this would not work with graphql-jit. Or does it?

I didn't test this with jit enabled, I suppose it will not work for the graphql trace, but we will still get the http trace at least. 
Maybe we can add some mercurius specific spans / attributes in another PR.

PacoDu:
Normal:
image

Jitted:
image

mcollina (approved):
lgtm

PR #346, opened 2020-11-11
feat: opentelemetry integration

PacoDu:
Mercurius works with existing tools like `@autotelic/fastify-opentelemetry` and `@opentelemetry/instrumentation-graphql`. But it's missing some mercurius specific spans and attributes that we can add with the opentelemetry API.

As a reference, here is an example of a mercurius trace with a gateway and a service and without any additional opentelemetry integration:
image

And here is an example with a simple mercurius - graphql span added to the main fastifyGraphql function and a cached attribute (see PR changes):
image
 

We need to define what we want to expose in opentelemetry traces (spans and attributes). And maybe add a `telemetry` option to mercurius so user can enable/disable telemetry ?

### DIFF
--- a/docs/integrations/README.md
+++ b/docs/integrations/README.md
@@ -4,3 +4,4 @@
 
 - [@nexus/schema](/docs/integrations/nexus-schema.md) - Declarative, code-first and strongly typed GraphQL schema construction for TypeScript & JavaScript
 - [mercurius-integration-testing](/docs/integrations/mercurius-integration-testing.md) - Utility library for writing mercurius integration tests.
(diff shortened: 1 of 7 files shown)


PacoDu:
Depends on #341

mcollina:
@PabloDu any updates on this one?

PacoDu:
This is a demonstration of a span usage inside the mercurius library, I don't have time right now to push it further unfortunately.
If you want I can cleanup this PR and just add the tracer and the base mercurius span.

PacoDu:
superseded by #363

Issue #348, opened 2020-11-12
fix: extendSchema does not work with federationMedata set to true


PacoDu:
The `_service { sdl }` is not updated on extendSchema with federationMetadata set to true, which means that extendSchema does not work at all if federation is enabled because the sdl does not reflect the changes. 

My current workaround is to merge schemas and resolvers from different module before registering mercurius. But that makes it difficult to split code in different modules without some custom wrapper.

IMO we should clearly state in the documentation that extendSchema does not work with federation and throw if used or better fix the issue.

I think this is related to #273 #229 and my first attempt to fix this #262 

I'll try an other fix by implementing `stripCommonPrimitives` based on apollo implementation as suggested in #229. This should simplify the logic of `buildFederatedSchema` and `extendFederatedSchema` proposed in #262

mcollina:
I think the correct path is to throw in `extendSchema` if federation is enabled and then have a better fix that adds support for it.

simoneb:
Any news on this one?

PR #349, opened 2020-11-13
fix(extendSchema): throw if federationMetadata is enabled

PacoDu:
1 step to address the issue #348 

This PR prevent the usage of `extendSchema` if `federationMedata` is enabled


mcollina (approved):
lgtm

Issue #356, opened 2020-12-05
fix: gateway doesn't strip non-service types when querying with inline fragments


projectjudge:
The gateway doesn't strip non-service types when querying federated services using inline fragments.

For the following schema:

Service 1:

```graphql
type Query @extends {
  me: UserUnion
}

type User @key(fields: "id") {
  id: ID!
  name: String!
}

type Admin {
  id: ID!
}

union UserUnion = User | Admin
```

Service 2:

```graphql
type Post @key(fields: "pid") {
  pid: ID!
  author: User
}

type User @key(fields: "id") @extends {
  id: ID! @external
  topPosts(count: Int!): [Post]
}
```

Gateway:

```graphql
type Query {
  me: UserUnion
}

type User {
  id: ID!
  name: String!
  topPosts(count: Int!): [Post]
}

type Post {
  pid: ID!
  author: User
}

type Admin {
  id: ID!
}

union UserUnion = User | Admin
```

If we query using an inline fragment:

```graphql
query {
  me {
    ... on User {
      id
      name
      topPosts(count: 2) {
        pid
        author {
          id
          topPosts(count: 1) {
            pid
          }
        }
      }
    }
  }
}
```

We get the following error:

```json
"Cannot query field \"topPosts\" on type \"User\"."
```

Looking into the Mercurius code, the issue appears to be with the `removeNonServiceTypeFields` function which doesn't strip non-service types for inline fragments. As a result, the gateway sends a query to the user service which includes `topPosts` and errors.

## Proposed solution

Remove non-service types for inline fragments by including logic for inline fragments.

I have prepared a PR in my fork which has the following changes: https://github.com/projectjudge/mercurius/commit/e31269c3667b94f0ea7a65239af9949c3ef2b3f3

If you are happy with the method, I'll submit this fix as a PR :)

mcollina:
Go ahead and open a PR!

PR #357, opened 2020-12-05
fix: gateway doesn't strip non-service types when querying with inline fragments

projectjudge:
fixes: https://github.com/mercurius-js/mercurius/issues/356


mcollina (approved):
lgtm

PacoDu (approved):
lgtm

PR #363, opened 2020-12-14
feat: base opentelemetry integration

PacoDu:
This PR adds a base mercurius span to the fastifyGraphql decorator and the cached attribute.

image

### DIFF
--- a/index.js
+++ b/index.js
@@ -38,6 +38,7 @@ const {
   MER_ERR_METHOD_NOT_ALLOWED,
   MER_ERR_INVALID_METHOD
 } = require('./lib/errors')
+const tracer = require('./lib/tracer')
 
 const kLoaders = Symbol('mercurius.loaders')
 const kFactory = Symbol('mercurius.loadersFactory')
@@ -382,12 +383,18 @@ const plugin = fp(async function (app, opts) {
   }
 
   async function fastifyGraphQl (source, context, variables, operationName) {
+    const span = tracer.startSpan('mercurius - graphql', { parent: tracer.getCurrentSpan() })
+
     context = Object.assign({ app: this, lruGatewayResolvers }, context)
     const reply = context.reply
 
     // Parse, with a little lru
     const cached = lru !== null && lru.get(source)
     let document = null
+
+    // Add opentelemetry attributes
+    span.setAttribute('mercurius.cached', cached)
+
     if (!cached) {
       // We use two caches to avoid errors bust the good
       // cache. This is a protection against DoS attacks
@@ -397,6 +404,7 @@ const plugin = fp(async function (app, opts) {
         // this query errored
         const err = new MER_ERR_GQL_VALIDATION()
         err.errors = cachedError.validationErrors
+        span.end()
         throw err
       }
 
@@ -405,6 +413,7 @@ const plugin = fp(async function (app, opts) {
       } catch (syntaxError) {
         const err = new MER_ERR_GQL_VALIDATION()
         err.errors = [syntaxError]
+        span.end()
         throw err
       }
 
@@ -425,6 +434,7 @@ const plugin = fp(async function (app, opts) {
         }
         const err = new MER_ERR_GQL_VALIDATION()
         err.errors = validationErrors
+        span.end()
         throw err
       }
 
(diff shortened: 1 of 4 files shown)


mcollina (commented):
I would generically prefer if we provided hooks (similar to fastify, or even simpler) for APMs to plug into. This is totally ok but it's non-scalable as there are a few different implementations for this.

PacoDu:
> I would generically prefer if we provided hooks (similar to fastify, or even simpler) for APMs to plug into. This is totally ok but it's non-scalable as there are a few different implementations for this.

As stated in https://github.com/mercurius-js/mercurius/pull/346#issuecomment-739436886 this is just a cleanup of the previous PR with the tracer declaration and the base mercurius span. Unfortunately, I don't have time to push this implementation further for now.

But, I'm not sure to understand what you mean by providing hooks, as opentelemetry is already providing an API for the end-user to extend the tracing and how it is plugged. This implementation is based on opentelemetry recommendations: https://open-telemetry.github.io/opentelemetry-js/#library-authors

mcollina:
> But, I'm not sure to understand what you mean by providing hooks, as opentelemetry is already providing an API for the end-user to extend the tracing and how it is plugged. This implementation is based on opentelemetry recommendations: https://open-telemetry.github.io/opentelemetry-js/#library-authors

Some APMs do not implement OpenTelemetry _but_ they have similar concepts/libraries. They'd need to put logic in the same places. Instead of calling opentracing directly, we could do:

```js
app.graphql.setTracing({
  ... // all the methods for those points
})
```

Sytten:
I also support hooks, I am trying to build an SDK for graphmetrics.io and we are not openmetrics compliant for now (we will consider it when the spec stabilizes a bit).
I think we can get quite far with the new 6.12.0 hooks, but we are missing a hook on the field level to measure their resolution time (not applicable for gateway/federation obviously).

mcollina:
@Sytten this is probably not the right issue to talk about. Maybe you can open a new one? Or just send a PR.

airhorns:
FWIW I think the aspirations of opentelemetry is to avoid having different APM products all have their own interfaces, and instead give library authors one generally accepted API to integrate against. Datadog, Honeycomb, Lightstep, and others all support the otel format for stuff and they are all collaborating on the client libraries explicitly so that we don't all have to try to build support for each individually or none at all. The idea would be if you want to export telemetry from your system to whatever new APM product comes along, you write a new opentelemetry exporter instead of a new integration for each piece of your system. I don't know if it will work and it is early days yet, but at least in the kubernetes space, a lot of communities have standardized on opentelemetry being *the* way to get telemetry data out of their system, and users use whatever exporter they want (or the opentelemetry collector, which is like a network attached exporter that can redirect to yet more places).

package.json
```diff
@@ -28,6 +28,7 @@
     "@graphql-tools/merge": "^6.2.4",
     "@graphql-tools/schema": "^6.2.4",
     "@graphql-tools/utils": "^6.2.4",
+    "@opentelemetry/api": "^0.13.0",
```
airhorns, line 31: `@opentelemetry/api` version `0.16.0` changes the API significantly, if we do want to get this in we should try to build against the most up to date tracer API and active-context-getting functions they have available (the `getSpan`, `setSpan` stuff)

index.js
```diff
@@ -397,6 +404,7 @@ const plugin = fp(async function (app, opts) {
         // this query errored
         const err = new MER_ERR_GQL_VALIDATION()
         err.errors = cachedError.validationErrors
+        span.end()
```
SkeLLLa, line 407: Usually on erros error info is added to span as status code

```
const { SpanStatusCode } = require('@opentelemetry/api')

span.setStatus({
  code: SpanStatusCode.ERROR,
  message: error.message,
});
```

As well as `OK` status on success.
```
span.setStatus({
  code: SpanStatusCode.OK,
});
```

Also to avoid multiple `span.end()` I prefer following pattern:

```js
const result = tracer.startActiveSpan(
  `<span_name>`,
  { attributes: <spanAttributes> },
  async (span) => {
    try {
      // do something
      span.setStatus({ code: SpanStatusCode.OK });
      // return if needed
    } catch (error) {
      span.setStatus({
        code: SpanStatusCode.ERROR,
        message: error.message,
      });
      throw error;
    } finally {
      span.end();
    }
  },
);
return result;
```

But it might be little bit less performant. And this one is available in most recent opentelemetry API, not sure about version from package.json.

Issue #283, opened 2020-09-18
unable to run subscriptions


nabsofken:
I am tying out the example for subscriptions but I am getting no luck. The mutation works fine, however, the subscription resolver never seems to get executed. Not sure what I am doing wrong. Any ideas?.
 Here is my code:
```Mutation: {
    addNewTeam: async (_, { teamName }, { pubsub } ) => {
      const newTeam = { name: teamName };
      myTeams.push(newTeam);
      console.log("publishing the team");
      await pubsub.publish({
        topic: "TEAM_ADDED",
        payload: {
          teamAdded: newTeam,
        },
      });
      console.log(pubsub);
      return newTeam;
    },
  },
  Subscription: {
    teamAdded: {
      subscribe: async (root, args, { pubsub }) => {
        const subs = await pubsub.subscribe("TEAM_ADDED");
        console.log(subs);
        return subs;
      },
    },
  },


This is my index.js file


import Fastify from "fastify";
import GQL from "fastify-gql";
import cors from "fastify-cors";
import { schema } from "./schema";
import { resolvers } from "./resolvers";

const app = Fastify({ logger: true });

app.log.info("Starting server");
const start = async () => {
  app.register(GQL, {
    schema: schema,
    resolvers: resolvers,
    subscription: true,
    graphiql: true,
  });
  app.register(cors, { origin: "http://localhost:4000" });
  await app.listen(3000);
};
start();

mcollina:
Can you please upload a full example to reproduce?

nabsofken:
[demo_backend_copy.zip](https://github.com/mcollina/fastify-gql/files/5247516/demo_backend_copy.zip)
@mcollina I have uploaded the example

nabsofken:
node version v14.9.0 if this helps as well

frikille:
I've run your code and it works fine.

<img width="1300" alt="Screenshot 2020-09-21 at 15 54 15" src="https://user-images.githubusercontent.com/569278/93775604-e6d49300-fc22-11ea-9f92-122b3f0fb7cd.png">

(You can access this screen if you change the `graphiql` option from `true` to `'playground'` and the visit `localhost:3000/playground` instead of `localhost:3000/graphiql`)

As soon as you create a subscription and then run the mutation, you should see the new team is pushed to the client.

nabsofken:
@frikille thanks for that! Yes I got it to work by setting graphiql to playground. Thanks!

PR #365, opened 2020-12-14
docs(api/options): add a note to graphiql option

RafaelGSS:
I had some problems with `subscriptions` on graphiql until I found #283. It would be better if have some notes to help instead of search for each every time. This PR provides it.


mcollina (approved):
lgtm

PR #367, opened 2020-12-22
feat: GraphQL playground headers

simoneb:
Adds a `playgroundHeaders` option which works with GraphQL playground and allows adding HTTP headers such as authorization.

It supports objects or, probably more useful, functions. The function version relies on the ability to serialize the function definition to a string, which is injected in the client side JS and invoked with the window object, basically allowing to extract headers from browser's storage or any place accessible to the client.

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -40,6 +40,7 @@
 
 - `playgroundSettings` Object. that allow you to configure GraphQL Playground with [playground
   options](https://github.com/prisma-labs/graphql-playground#usage). it works if the graphiql is set to `'playground'`.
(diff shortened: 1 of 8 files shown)


mcollina (approved):
lgtm

PabloSzx:
there is missing the types, it could be something like 

```ts
playgroundHeaders?: ((window: Window) => object) | object
```

in line 314 of index.d.ts

simoneb:
> there is missing the types, it could be something like
> 
> ```ts
> playgroundHeaders?: ((window: Window) => object) | object
> ```
> 
> in line 314 of index.d.ts

good catch. done (had to change tsconfig to allow it to recognize the Window object)

PabloSzx (approved):

PR #371, opened 2020-12-26
feat: add subscription.onDisconnect callback for subscriptions

leonzalion:
I was switching from Apollo Server to Mercurius, but I realized that the onDisconnect functionality that I used in Apollo Server which I used for cleanup in my app was missing from Mercurius. I saw a similar feature addition in #167 with the addition of onConnect, so I decided to add an onDisconnect alongside most places where I found onConnect, and so far it seems to be working.

In my use case, I need onDisconnect to be called with the context created by onConnect which contains the information I would need for cleanup, but I'm not sure if that's the best argument to pass to onDisconnect.

Thank you for your hard work on mercurius and I'm looking forward to using it! If there's any changes you'd like me to make with this PR, do let me know!

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -25,6 +25,7 @@ module.exports = class SubscriptionConnection {
     entityResolvers,
     context = {},
     onConnect,
+    onDisconnect,
     resolveContext
   }) {
     this.fastify = fastify
@@ -33,6 +34,7 @@ module.exports = class SubscriptionConnection {
     this.entityResolvers = entityResolvers
     this.subscriber = subscriber
     this.onConnect = onConnect
+    this.onDisconnect = onDisconnect
     this.subscriptionContexts = new Map()
     this.context = context
     this.isReady = false
@@ -208,6 +210,10 @@ module.exports = class SubscriptionConnection {
   }
 
   handleConnectionClose () {
+    if (typeof this.onDisconnect === 'function') {
+      this.onDisconnect(this.context)
+    }
+
     Array
       .from(this.subscriptionContexts.values())
       .map(subscriptionContext =>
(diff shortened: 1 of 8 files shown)


mcollina (changes requested):
Would you mind to add a couple test when onDisconnect() returns a promise (success and error case)? I don't think that is handled in the code.

mcollina:
Good work!

leonzalion:
Thanks :)

I made the handleConnectionClose async and added a try wrapper around the onDisconnect call which logs the error if the callback throws or the promise rejects. I'm not sure if making it async was the best way, but based on other parts of the code it looked to be common. I also moved the onDisconnect call after the closing of the sockets and subscriptionContexts, which I think is more appropriate. I'm not too sure if the way I wrote the tests is how you envisioned it, since I couldn't find any other similar examples of overriding the log.error function in other tests. In any case, do let me know if you would like any changes!

lib/subscription-connection.js
```diff
@@ -207,12 +209,20 @@ module.exports = class SubscriptionConnection {
     this.subscriptionContexts.delete(data.id)
   }
 
-  handleConnectionClose () {
+  async handleConnectionClose () {
```
mcollina, line 212: who is calling this function? Are you sure they are using `await` on it? I think a better approach is add a `.catch()` handler instead.

leonzalion: Yeah, I agree that using .catch() would be better, but initially I wasn't entirely too sure how to use `.catch()` on a callback that could be potentially both synchronous and asynchronous. After a bit of searching, I ended up using `Promise.resolve().then(() => cb()).catch((e) => { log(e) })`, but I'm not sure if this is the best approach.

Let me know if you know of a better way to handle this!

mcollina (approved):
lgtm

Issue #369, opened 2020-12-23
How to use subscriptions with fastify-websocket?


acro5piano:
First of all, thank you for the great library.

I'm trying to work both mercurius + fastify-websocket using `subscription: true`, but stacking.

My code is like this:

```ts
const Fastify = require('fastify')
const fastifyWebsocket = require('fastify-websocket')
const mercurius = require('mercurius')

const { schema, resolvers } = require('./schema') // The same one from official docs

const app = new Fastify({
  logger: true,
})

// If we register `fastifyWebsocket` on our own, it raises:
//    Error: server.handleUpgrade() was called more than once with the same socket,
//    possibly due to a misconfiguration
// app.register(fastifyWebsocket)

app.register(mercurius, {
  schema,
  resolvers,
  subscription: true,
})

app.get('/ws', { websocket: true }, (conn) => {
  console.log(conn)
})

app.listen(6767)
```

Then connect to ws://127.0.0.1:6767/ws and I'm seeing the following output:

```
$ wscat -c ws://127.0.0.1:6767/ws

Connected (press CTRL+C to quit)
< {"error":"unknown route"}
Disconnected (code: 1005, reason: "")
```

I've created reproductive repo.

https://github.com/acro5piano/mercurius-with-fastify-websocket

Do you have any hints on it? 

Possible solutions are:

- Allow to pass my `wss` and pass it to `fastify-websocket`.
- Allow use multiple WebSocket server and use `server.on('upgrade')`.

Best regards,

acro5piano:
Oh I've found a way to do it.

If I edit this line and 


https://github.com/acro5piano/mercurius/blob/c835596557e867d990a009a95ef3e7ff5f7c7c54/lib/subscription.js#L8

https://github.com/acro5piano/mercurius/blob/master/lib/subscription.js#L8

acro5piano:
I've saw the whole code and my assumption is the following. (I'm not so familiar with Fastify plugins)

- fastify-websocket adds WebSocket route using `onRoute` hook
- Fasity does not fire `onRoute` hook from plugins. However, if we wrap the plugin with `fp`, it fires.
- Currently plugins are multi-layers.
    - 1. mercurius (index.js)
    - 2. routes.js
    - 3. subscriptions.js
    - 4. fasitfy-websocket
- `route.js` isn't wrapped with `fp`, so `onRoute` hook which written at `fastify-gql` does not invoke

Is this correct? If so, are there anything I can do?

[...]

PR #372, opened 2020-12-27
fix: Do not register fastify-websocket if it is already registered

acro5piano:
Hi, thank you for the great library.

This PR fixes https://github.com/mercurius-js/mercurius/issues/369, which enables `mercurius` to work with `fastify-websocket`.

As @mcollina suggested on https://github.com/mercurius-js/mercurius/issues/369#issuecomment-751381695, I've added a check if `fastify-websocket` is already registered or not.

- My determination is that checking the decorator `fastify.websocketServer` is `undefined` or not. I think there should be a method to check a plugin registered or not, but this is what I know about Fastify.
- I've added a simple test for coverage. I've confirmed subscription is working even if `fastify-websocket` is already registered on my machine, but I'll happy to add more tests if needed.
 
Thanks!

### DIFF
--- a/test/subscription.js
+++ b/test/subscription.js
@@ -3,6 +3,7 @@ const Fastify = require('fastify')
 const WebSocket = require('ws')
 const mq = require('mqemitter')
 const { EventEmitter } = require('events')
+const fastifyWebsocket = require('fastify-websocket')
 const GQL = require('..')
 
 const FakeTimers = require('@sinonjs/fake-timers')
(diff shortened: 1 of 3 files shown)


test/subscription.js
```diff
@@ -1751,3 +1752,33 @@ test('`withFilter` tool works with async filters', t => {
     })
   })
 })
+
+test('subscription server does not register fastify-websocket if already registered', t => {
```
mcollina, line 1756: could you please expand this test by using the graphql subscription and adding another non-graphql websocket route?

acro5piano: ok!

acro5piano: @mcollina I've added the test, so could you check it again? Thank you for your review!

mcollina (approved):
lgtm, good work!

acro5piano:
@mcollina Thank you!

PR #373, opened 2020-12-28
add mac & windows ci workflows, replace eol with lf

PabloSzx:
> also added all the different lockfiles options to the .gitignore


mcollina (approved):
lgtm

PR #377, opened 2021-01-11
fix: Don't use extendSchema's validations in buildFederationSchema.

HW13:
### Summary
Currently a gateway will throw a validation error if more than one federated service has the same directive defined - excluding directives added by `federationMetadata`. (ie. `Error: There can be only one directive named ...`)

While `compositionRules` does remove the `UniqueDirectivesPerLocationRule`, it is being circumvented by not passing `assumeValidSDL: true` to the `extendSchema` calls prior to the `validateSDL` call.

[This](https://github.com/mercurius-js/mercurius/blob/1ae98ad2d577e1a4d83a098544a21b358374d97c/lib/federation.js#L248) call of `extendSchema` is the direct cause of the issue, but based on [this](https://github.com/mercurius-js/mercurius/blob/1ae98ad2d577e1a4d83a098544a21b358374d97c/lib/federation.js#L259) comment I assume the intention is to not have `extendSchema` do any validation - so I've added `assumeValidSDL: true` to the other `extendSchema` calls in `buildFederationSchema` as well.

### Test plan
The existence of this issue can be confirmed by running the added test without the changes to `buildFederationSchema`.

### DIFF
--- a/lib/federation.js
+++ b/lib/federation.js
@@ -233,7 +233,11 @@ function buildFederationSchema (schema, isGateway) {
     query: undefined
   })
 
-  federationSchema = extendSchema(federationSchema, parse(isGateway ? BASE_FEDERATION_TYPES : FEDERATION_SCHEMA))
+  federationSchema = extendSchema(
+    federationSchema,
+    parse(isGateway ? BASE_FEDERATION_TYPES : FEDERATION_SCHEMA),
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

HW13:
Thanks! 😄

Issue #65, opened 2019-08-06
Authentication and Authorization Example


mthota15:
Any Example on how to leverage fastify-jwt plugin with fastify-gql for authentication and Authorization example?

thiagokroger:
Here's an example using fastify-jwt and fastify-auth

`
const fastify = require("fastify")();

const users = [
  { id: 1, name: "matteo", email: "matteo@email.com", password: "12345" },
  { id: 2, name: "thiago", email: "thiago@email.com", password: "123456" }
];

const getToken = (email, password) => {
  const user = users.find(u => u.email === email && u.password === password);
  return user ? fastify.jwt.sign({ id: user.id }) : null;
};

const getUserName = id => {
  const user = users.find(u => u.id === id);
  return user ? user.name : null;
};

const verifyJWT = async (request, reply) => {
  try {
    await request.jwtVerify();
  } catch (err) {
    reply.send(err);
  }
};

const registerRoutes = () => {
  fastify
    .post("/login", async (request, reply) => {
      const { email, password } = request.body;
      try {
        return reply.graphql(
          `{ token(email: "${email}", password: "${password}") }`
        );
      } catch (error) {
        return reply.error(error);
      }
    })
    .route({
      method: "GET",
      url: "/user-name",
      preHandler: fastify.auth([fastify.verifyJWT]),
      handler: (request, reply) => {
        const userId = request.user.id;
        try {
          return reply.graphql(`{ userName(id: ${userId}) }`);
        } catch (error) {
          return reply.send;
        }
      }
    });
};

fastify
  .register(require("fastify-auth"))
  .register(require("fastify-jwt"), { secret: "supersecret" })
  .register(require("fastify-gql"), {
    schema: `
      type Query {
        token(email: String, password: String): String,
        userName(id: Int): String
      }
    `,
    resolvers: {
      Query: {
        token: async (_, { email, password }) => getToken(email, password),
        userName: async (_, { id }) => {
          return getUserName(id);
        }
      }
    }
  })
  .decorate("verifyJWT", verifyJWT)
  .after(registerRoutes)
  .listen(9000, (err, address) => {
    if (err) {
      console.log(err);
      process.exit(1);
    }
    console.log("Server running at: " + address);
  });
`

Full code here: [fastify-graphql-jwt-example](https://github.com/thiagokroger/fastify-graphql-jwt-example)

Eomm:
I have started a basic example that fit your needs for authentication in the example repo.

[...]

PR #378, opened 2021-01-20
Add graphql-shield authorization example

jmfury:
#### This PR adds an example using `graphql-shield` 
This is in reference to #65 

- The [`graphql-shield`](https://github.com/maticzav/graphql-shield) library provides a middleware for permission / authorization at the GraphQL type or field level
- This leverages `applyMiddleware` from [`graphql-middleware`](https://github.com/maticzav/graphql-middleware) since that is the applicable path

**Note:**
Also uses `makeExecutableSchema` since again, that is the applicable path here

#### Test queries for Playground
Here are a couple of test queries to run as any given "user" (via `authorization` header)
```graphql
 query Customers {  # try as Alice (admin)
    customers {
      id
    }
  }

  query Fruits { # try as Alice or Bob (admin or editor)
    fruits {
     name
    }
  }

  query FrontPage { # omit authorization header (public user)
    frontPage {
      count
      name
    }
  }

```

Happy to clean this up, clarify, or add a README as needed! :smile:


package.json
```diff
@@ -34,9 +34,11 @@
     "@typescript-eslint/eslint-plugin": "^3.6.1",
     "@typescript-eslint/parser": "^3.6.1",
     "autocannon": "^6.0.0",
-    "graphql-tools": "^6.2.6",
     "docsify-cli": "^4.4.1",
     "fastify": "^3.0.2",
+    "graphql-middleware": "^6.0.2",
+    "graphql-shield": "^7.5.0",
+    "graphql-tools": "^6.2.6",
```
jmfury, line 41: I believe the `graphql-tools` change here is my editor doing a _reorder_ or deps there.

mcollina (approved):
lgtm... nice!

PR #381, opened 2021-01-25
fix(gateway/refresh): set typeToServiceMap Query, Mutation and Subscription properties to null

### LINKED ISSUE #379: Gateway schema refresh issue with TypeGraphQL
Hey there,

First thank you for the good work on this awesome library 🙏 , the fastify eco-system is just neat !

----

I have an issue that's making me crazy and I'm seeking your help on this.

I have a gateway running in federation mode like this:

```
server.register(mercurius, {
    jit: 1,
    graphiql: isProduction ? false : 'playground',
    gateway: {
      services: [
        {
          name: 'my-service',
          url: 'http://localhost:8080/graphql',
          rewriteHeaders: (headers: IncomingHttpHeaders) => {
            if (headers.authorization) {
              return {
                authorization: headers.authorization
              };
            }
            return {};
          },
          mandatory: true
        }
      ],
      pollingInterval: POOLING_INTERVAL_IN_MS
    }
  });
```

And a downstream service using mercurius (not in gateway mode) with schema generated through [type-graphql](https://github.com/MichalLytek/type-graphql).

The generated schema is:

```
# -----------------------------------------------
# !!! THIS FILE WAS GENERATED BY TYPE-GRAPHQL !!!
# !!!   DO NOT MODIFY THIS FILE BY YOURSELF   !!!
# -----------------------------------------------

directive @extends on INTERFACE | OBJECT

directive @external on FIELD_DEFINITION | OBJECT

directive @key(fields: String!) on INTERFACE | OBJECT

directive @provides(fields: String!) on FIELD_DEFINITION

directive @requires(fields: String!) on FIELD_DEFINITION

type User {
  email: String!
  id: ID!
  name: String!
}

type Query {
  findUserByEmail(email: String!): User!
}

```

**Issue**:
1. The gateway and the downstream service are running
2. I add a new mutation/query whatever to the downstream service
3. I restart the downstream service to regenerate the schema
4. And when I try to query the gateway with the newly query/mutation or even with the existing one, I got an error saying "**Cannot return null for non-nullable field Query.myNewQuery"**

The only way to make it works is to restart the gateway (or triggering a new deployment) which is not very convenient 😞 .

I'm suspecting a schema issue since the schema is not regenerated even on `playground`.

Do you have any inputs ?

Thanks for your time ! 😄 

Franck

franck-romano:
Closes #379

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -368,6 +368,10 @@ async function buildGateway (gatewayOpts, app) {
 
       const schema = buildFederatedSchema(_serviceSDLs, true)
 
+      typeToServiceMap.Query = null
+      typeToServiceMap.Mutation = null
+      typeToServiceMap.Subscription = null
+
       defineResolvers(schema, typeToServiceMap, serviceMap, typeFieldsToService)
 
       return schema
--- a/test/gateway/pollingInterval.js
+++ b/test/gateway/pollingInterval.js
@@ -978,3 +978,204 @@ test('Polling schemas (subscriptions should be handled)', async (t) => {
 
   clock.uninstall()
 })
+
+test('Polling schemas (should properly regenerate the schema when a downstream service restarts)', async (t) => {
+  const clock = FakeTimers.install({
+    shouldAdvanceTime: true,
+    advanceTimeDelta: 40
+  })
+  const oldSchema = `
+    directive @extends on INTERFACE | OBJECT
+
+    directive @external on FIELD_DEFINITION | OBJECT
+
+    directive @key(fields: String!) on INTERFACE | OBJECT
+
+    directive @provides(fields: String!) on FIELD_DEFINITION
+
+    directive @requires(fields: String!) on FIELD_DEFINITION
+
+    type Query {
+      me: User
+    }
+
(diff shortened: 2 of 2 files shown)


mcollina (changes requested):
Please add a unit test.

franck-romano:
Yup, did it.

However, it seems that I am not able to make the test fail, event if I remove the `null` assignment  that I did before.
So I am not very convinced by this test since it won't prevent the problem to happen again.

Curious to know your thoughts on the best strategy to reproduce the issue in test files.

I will iterate over time on it !

mcollina (dismissed):
lgtm

mcollina:
> Curious to know your thoughts on the best strategy to reproduce the issue in test files.

I would recommend to copy the schema generated by typegraphql and past it in

mcollina:
It seems the test is failing even if there is the proposed fix.

franck-romano:
> It seems the test is failing even if there is the proposed fix.

Yup, seems that it's now passing ✅

mcollina (approved):
lgtm

Issue #380, opened 2021-01-21
Add hooks to Mercurius


projectjudge:
In order to define gateway authentication and authorization as discussed in https://github.com/mercurius-js/mercurius/issues/343, we should add GraphQL hooks support to Mercurius so we can process data at different points in the request lifecycle.

## Requirements

- [ ] All necessary hooks triggered in Mercurius at the correct stage of the GraphQL request lifecycle
- [ ] One can define multiple hooks of the same type where they run sequentially in the order of their definition
- [ ] Make sure these hooks can be easily extended
- [ ] Make sure these hooks are cleaned up correctly when needed (e.g. errors, timeouts)
- [ ] Make sure just promises are supported
- [ ] Documentation, TypeScript types and tests in Mercurius

## Research and Exploration

Fastify currently uses hooks to allows users to define custom functionality at certain events within the codebase. You can define multiple hooks of the same name in this way. They are executed sequentially from the order at which they are defined.

## Proposed solution

In Mercurius, we will add support for hooks by implementing core functionality also defined in Fastify. This will allow users of Mercurius and plugins to utilise the defined GraphQL lifecycle events. The hooks we will add will be:

- `preParsing`: before queries are extracted from the input request. It will contain the following data:
	- MercuriusContext
	- Request
- `preValidation`: before the queries are validated against the schema. It will contain the following data:
	- Extracted query data
	- MercuriusContext
	- Request
- `preExecution`: this hook is triggered before queries are sent to the federated services (in the case of a gateway) or executing the query on the current GraphQL service. It will contain the following data:
	- Query AST(s) split by service
	- Destination schema AST(s) split by service
	- MercuriusContext
	- Request
- `preResolution`: before the responses from each federated service or single service is crafted into a GraphQL response. It will contain the following data:
	- Response(s) split by service
	- MercuriusContext
	- Request

All new code with be 100% covered by tests and documentation will also be provided. The new hooks will be correctly typed. I have already started work on an implementation of this and will submit a PR for this once it is done and everyone is agreed on the solution :)

mcollina:
Thanks for the long description, it's fantastic. SGTM with the plan, are you going to work on it?

A few notes: fastify supports both callbacks and promises. We should just support promises for this as the whole GraphQL ecosystem is all promise-based.

projectjudge:
Yep, I planned on starting the implementation today if that's okay with you?

Thanks for the tips about the callbacks/promise support - I'll make sure just promises are supported :)

PR #385, opened 2021-01-29
Add hooks to Mercurius

projectjudge:
Closes: https://github.com/mercurius-js/mercurius/issues/380

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -17,7 +17,6 @@ By using hooks you can interact directly with the GraphQL lifecycle of Mercurius
 
 ## GraphQL Request Hooks
 
-<!-- TODO -->
 It is pretty easy to understand where each hook is executed by looking at the [lifecycle page](lifecycle.md).<br>
 
 There are four different hooks that you can use in GraphQL Request *(in order of execution)*:
@@ -30,8 +29,7 @@ await fastify.ready()
 
 ### preParsing
 
-<!-- TODO -->
-If you are using the `preParsing` hook, you can transform the GraphQL query string before it is parsed. It receives the schema and context objects as other hooks.
+If you are using the `preParsing` hook, you can access the GraphQL query string before it is parsed. It receives the schema and context objects as other hooks.
 
 For instance, you can register some tracing events:
 
@@ -101,7 +99,7 @@ If you get an error during the execution of your hook, you can just throw an err
 **Notice:** there is one exception to this with the `preGatewayExecution` hook, which will continue execution of the rest of the query and append the error to the errors array in the response.
 
 ```js
-fastify.addHook('onResponse', async (request, reply) => {
+fastify.addHook('preParsing', async (schema, source, context) => {
   throw new Error('Some error')
 })
 ```
@@ -132,7 +130,7 @@ Note, the original query will still execute. Adding the above will result in the
 ```json
 {
   "data": {
-    ...
+    foo: "bar"
   },
   "errors": [
     {
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -1,0 +1,146 @@
+# Hooks
+
+Hooks are registered with the `fastify.graphql.addHook` method and allow you to listen to specific events in the GraphQL request/response lifecycle. You have to register a hook before the event is triggered, otherwise the event is lost.
+
+By using hooks you can interact directly with the GraphQL lifecycle of Mercurius. There are GraphQL Request hooks:
+
+- [GraphQL Request Hooks](#graphql-request-hooks)
+  - [preParsing](#preparsing)
+  - [preValidation](#prevalidation)
+  - [preExecution](#preexecution)
+  - [preGatewayExecution](#pregatewayexecution)
+  - [onResolution](#onresolution)
+  - [Manage Errors from a hook](#manage-errors-from-a-hook)
+  - [Add errors to the GraphQL response from a hook](#add-errors-to-the-graphql-response-from-a-hook)
+
+**Notice:** these hooks are only supported with `async`/`await` or returning a `Promise`.
+
+## GraphQL Request Hooks
+
+<!-- TODO -->
+It is pretty easy to understand where each hook is executed by looking at the [lifecycle page](lifecycle.md).<br>
+
+There are four different hooks that you can use in GraphQL Request *(in order of execution)*:
+
+When registering hooks, you must wait for Mercurius to be registered in Fastify.
+
+```js
+await fastify.ready()
+```
+
+### preParsing
+
+<!-- TODO -->
+If you are using the `preParsing` hook, you can transform the GraphQL query string before it is parsed. It receives the schema and context objects as other hooks.
+
+For instance, you can register some tracing events:
+
+```js
+fastify.graphql.addHook('preParsing', async (schema, source, context) => {
+  await registerTraceEvent()
+})
+```
+
+### preValidation
+
+By the time the `preValidation` hook triggers, the query string has been parsed into a GraphQL Document AST.
+
+```js
+fastify.addHook('preValidation', async (schema, document, context) => {
+  await asyncMethod()
+})
+```
+
+### preExecution
+
+In the `preExecution` hook, you can modify the following items by returning them in the hook definition:
+  - `document`
+  - `errors`
+
+```js
+fastify.addHook('preExecution', async (schema, document, context) => {
+  const { modifiedDocument, errors } = await asyncMethod(document)
+
+  return {
+    document: modifiedDocument
+    errors
+  }
+})
+```
+
+### preGatewayExecution
+
+In the `preGatewayExecution` hook, you can modify the following items by returning them in the hook definition:
+  - `document`
+  - `errors`
+
+This hook will only be triggered in gateway mode. When in gateway mode, each hook definition will trigger multiple times in a single request just before executing remote GraphQL queries on the federated services.
+
+```js
+fastify.addHook('preGatewayExecution', async (schema, document, context) => {
+  const { modifiedDocument, errors } = await asyncMethod(document)
+
+  return {
+    document: modifiedDocument
+    errors
+  }
+})
+```
+
+### onResolution
+
+```js
+fastify.addHook('onResolution', async (execution, context) => {
+  await asyncMethod()
+})
+```
+
+### Manage Errors from a hook
+If you get an error during the execution of your hook, you can just throw an error and Mercurius will automatically close the GraphQL request and send the appropriate errors to the user.`
+
+**Notice:** there is one exception to this with the `preGatewayExecution` hook, which will continue execution of the rest of the query and append the error to the errors array in the response.
+
+```js
+fastify.addHook('onResponse', async (request, reply) => {
+  throw new Error('Some error')
+})
+```
+
+### Add errors to the GraphQL response from a hook
+
+The following hooks support adding errors to the GraphQL response. These are:
+
+ - `preExecution`
+ - `preGatewayExecution`
+
+```js
+fastify.addHook('preExecution', async (schema, document, context) => {
+  return {
+    errors: [new Error('foo')]
+  }
+})
+
+fastify.addHook('preExecution', async (schema, document, context) => {
+  return {
+    errors: [new Error('bar')]
+  }
+})
+```
+
+Note, the original query will still execute. Adding the above will result in the following response:
+
+```json
+{
+  "data": {
+    ...
+  },
+  "errors": [
+    {
+      "message": "foo"
+    },
+    {
+      "message": "bar"
+    }
+  ]
+}
+```
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -446,11 +446,12 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
 
     if (isQuery) {
       // TODO: Trigger per service preExecution
-      await preExecutionHandler({ schema, document: operation, context }, {})
+      // TODO: trigger new query if present
+      const { modifiedQuery } = await preExecutionHandler({ schema, document: operation, context })
       return service.sendRequest({
         method: 'POST',
         body: JSON.stringify({
-          query,
+          query: modifiedQuery || query,
           variables
         }),
         originalRequestHeaders: reply.request.headers
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -445,7 +445,10 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
 
     if (isQuery) {
       // Trigger preGatewayExecution hook
-      const { modifiedQuery } = await preGatewayExecutionHandler({ schema, document: operation, context })
+      let modifiedQuery
+      if (context.preGatewayExecution !== null) {
+        ({ modifiedQuery } = await preGatewayExecutionHandler({ schema, document: operation, context }))
+      }
 
       return service.sendRequest({
         method: 'POST',
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -373,8 +373,7 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
     if (cached) {
       variableNamesToDefine = cached.variableNamesToDefine
       query = cached.query
-      // TODO: work out whether to set in the cache or parse the query here
-      operation = parse(query)
+      operation = cached.operation
     } else {
       // Remove items from selections that are not defined in the service
       const selections = fieldNodes[0].selectionSet ? removeNonServiceTypeFields(fieldNodes[0].selectionSet.selections, service, type, schema) : []
@@ -414,7 +413,7 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
         }
       }
 
-      lruGatewayResolvers.set(`${__currentQuery}_${resolverKey}`, { query, variableNamesToDefine })
+      lruGatewayResolvers.set(`${__currentQuery}_${resolverKey}`, { query, operation, variableNamesToDefine })
     }
 
     const variables = {}
@@ -445,9 +444,9 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
     }
 
     if (isQuery) {
-      // TODO: Trigger per service preGatewayExecution
-      // TODO: trigger new query if present
+      // Trigger preGatewayExecution hook
       const { modifiedQuery } = await preGatewayExecutionHandler({ schema, document: operation, context })
+
       return service.sendRequest({
         method: 'POST',
         body: JSON.stringify({
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -6,7 +6,7 @@ const {
   parse,
   Kind
 } = require('graphql')
-const { preExecutionHandler } = require('../handlers')
+const { preGatewayExecutionHandler } = require('../handlers')
 
 const kEntityResolvers = Symbol('mercurius.entity-resolvers')
 
@@ -445,9 +445,9 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
     }
 
     if (isQuery) {
-      // TODO: Trigger per service preExecution
+      // TODO: Trigger per service preGatewayExecution
       // TODO: trigger new query if present
-      const { modifiedQuery } = await preExecutionHandler({ schema, document: operation, context })
+      const { modifiedQuery } = await preGatewayExecutionHandler({ schema, document: operation, context })
       return service.sendRequest({
         method: 'POST',
         body: JSON.stringify({
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -6,6 +6,7 @@ const {
   parse,
   Kind
 } = require('graphql')
+const { preExecutionHandler } = require('../handlers')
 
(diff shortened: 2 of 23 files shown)


lib/hooks.js
```diff
@@ -0,0 +1,161 @@
...
+      }
+    }
+    await next(null, newDocument, newErrors)
+  }
+
+  function handleReject (err) {
+    cb(err, request, executionResult)
+  }
+
+  await next()
```
mcollina, line 137: Could you please avoid the recursion and use a loop instead? We are just using promises here, there is no need to have a callback.

lib/hooks.js
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  function handleReject (err) {
+    cb(err, request)
+  }
+
+  await next()
+}
+
+async function preExecutionHookRunner (functions, request, executionResult, cb) {
```
mcollina, line 77: don't mix callbacks and async

lib/hooks.js
```diff
@@ -0,0 +1,161 @@
...
+
+  async function handleResolve () {
+    await next()
+  }
+
+  function handleReject (err) {
+    cb(err, request)
+  }
+
+  await next()
```
mcollina, line 74: use a loop, don't recurse with async.

test/gateway/hooks.js
```diff
@@ -0,0 +1,1075 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('fastify')
+const { GraphQLSchema, parse } = require('graphql')
+const GQL = require('../..')
+
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
```
mcollina, line 8: Please use `promisify(setTimeout)` instead

mcollina (changes requested):
I've done a quick pass, I need to get deeper but I think there are a few things to fix already.

lib/hooks.js
projectjudge: yep, no problem

lib/handlers.js
```diff
@@ -0,0 +1,67 @@
...
+const { print } = require('graphql')
+
+async function preParsingHandler (request) {
+  if (request.context.preParsing !== null) {
+    await hooksRunner(
+      request.context.preParsing,
+      preParsingHookRunner,
+      request
+    )
+  }
```
mcollina, line 12: The if should be moved to the caller instead.

lib/handlers.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+async function preValidationHandler (request) {
+  if (request.context.preValidation !== null) {
+    await hooksRunner(
+      request.context.preValidation,
+      hookRunner,
+      request
+    )
+  }
```
mcollina, line 22: The if should be moved to the caller instead.

lib/handlers.js
```diff
@@ -0,0 +1,67 @@
...
+      request.context.preExecution,
+      request
+    )
+    if (errors.length > 0) {
+      addErrorsToContext(request.context, errors)
+    }
+    if (typeof modifiedDocument !== 'undefined') {
+      return { modifiedDocument, modifiedQuery: print(request.document) }
+    }
+  }
```
mcollina, line 37: The if should be moved to the caller instead.

lib/handlers.js
```diff
@@ -0,0 +1,67 @@
...
+      request.context.preGatewayExecution,
+      request
+    )
+    if (errors.length > 0) {
+      addErrorsToContext(request.context, errors)
+    }
+    if (typeof modifiedDocument !== 'undefined') {
+      return { modifiedDocument, modifiedQuery: print(modifiedDocument) }
+    }
+  }
```
mcollina, line 53: The if should be moved to the caller instead.

lib/handlers.js
```diff
@@ -0,0 +1,67 @@
...
+}
+
+async function onResolutionHandler (request) {
+  if (request.context.onResolution !== null) {
+    await hooksRunner(
+      request.context.onResolution,
+      onResolutionHookRunner,
+      request
+    )
+  }
```
mcollina, line 64: The if should be moved to the caller instead.

mcollina (changes requested):
Good work! How do the hooks play with subscriptions?

projectjudge:
Thanks! I have tested on a very basic level and seems to be playing okay. I'll write some thorough tests for subscriptions for both normal server and gateway mode to make sure it is all working as expected :) I also noticed that the CI checks are intermittently failing for some of the subscription tests. At first glance it seems to be an environmental issue (other checks for the same node version pass fine) - did you have any thoughts on this? (Example run: https://github.com/mercurius-js/mercurius/pull/385/checks?check_run_id=1791951386)

mcollina:
I've seen the flaky tests... can't do much about it right now. If you'd like to investigate and fix them, I'd be very grateful!

projectjudge:
For sure, very happy to take a look at the tests! :)

lib/hooks.js
```diff
@@ -0,0 +1,91 @@
...
+  this.preValidation = []
+  this.preExecution = []
+  this.preGatewayExecution = []
+  this.onResolution = []
+}
+
+Hooks.prototype.validate = function (hook, fn) {
+  if (typeof hook !== 'string') throw new MER_ERR_HOOK_INVALID_TYPE()
+  if (typeof fn !== 'function') throw new MER_ERR_HOOK_INVALID_HANDLER()
+  if (supportedHooks.indexOf(hook) === -1) {
```
mcollina, line 18:
```suggestion
  if (lifecycleHooks.indexOf(hook) === -1) {
```

I think you can remove the `supportedHooks` variable

test/hooks.js
```diff
@@ -0,0 +1,682 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('fastify')
+const { parse, GraphQLSchema } = require('graphql')
+const GQL = require('..')
+
+const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
```
mcollina, line 8:
```suggestion
const { promisify } = require('util')
const sleep = promisify(setTimeout)
```

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+
+// -----
+// hooks
+// -----
+test('hooks', async t => {
+  t.plan(16)
+  const app = await createTestServer(t)
+
+  app.graphql.addHook('preParsing', async function (schema, source, context) {
+    await sleep(1)
```
mcollina, line 51: why is this needed?

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+    }
+  }
+}
+
+const query = '{ add(x: 2, y: 2) }'
+
+async function createTestServer (t, customResolvers = resolvers) {
+  const app = Fastify()
+  t.tearDown(() => {
+    app.close()
```
mcollina, line 32:
```suggestion
    return app.close()
```

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+  })
+
+  app.register(GQL, { schema, resolvers: customResolvers })
+
+  // needed so that graphql is defined
+  await app.ready()
```
mcollina, line 38: This is actually not needed anymore, just returning app would do.

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.fail('this should not be called')
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 203: this is not needed at all.

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+  })
+
+  app.graphql.addHook('preParsing', async (schema, source, context) => {
+    t.type(schema, GraphQLSchema)
+    t.is(source, query)
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 240: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.fail('this should not be called')
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 282: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+  })
+
+  app.graphql.addHook('preValidation', async (schema, document, context) => {
+    t.type(schema, GraphQLSchema)
+    t.deepEqual(document, parse(query))
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 319: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.fail('this should not be called')
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 357: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+  })
+
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.type(schema, GraphQLSchema)
+    t.deepEqual(document, parse(query))
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 394: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+        }
+      ],
+      loc: { start: 0, end: 19 }
+    }
+    return {
+      document: modifiedDocument
+    }
+  })
+
+  await app.listen(0)
```
mcollina, line 464: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.type(schema, GraphQLSchema)
+    t.deepEqual(document, parse(query))
+    t.type(context, 'object')
+    return {
+      errors: [new Error('bar')]
+    }
+  })
+
+  await app.listen(0)
```
mcollina, line 504: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
```
mcollina, line 560: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+// -------------
+test('preGatewayExecution hooks should never be called in a non-gateway graphql server', async t => {
+  t.plan(1)
+  const app = await createTestServer(t)
+
+  app.graphql.addHook('preGatewayExecution', async (schema, document, context) => {
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 600: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+    t.type(execution, 'object')
+    t.type(context, 'object')
+    throw new Error('a onResolution error occured')
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 633: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+    context.foo = 'bar'
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.type(execution, 'object')
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 668: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+      services: [{
+        name: 'user',
+        url: `http://localhost:${userServicePort}/graphql`
+      }, {
+        name: 'post',
+        url: `http://localhost:${postServicePort}/graphql`
+      }]
+    }
+  })
+  await gateway.ready()
```
mcollina, line 160: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+  })
+
+  // This should still be called in the gateway
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.type(execution, 'object')
+    t.type(context, 'object')
+    t.ok('onResolution called')
+  })
+
+  await app.listen(0)
```
mcollina, line 751: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+    t.type(context, 'object')
+    t.ok('preExecution called')
+    const documentClone = JSON.parse(JSON.stringify(document))
+    documentClone.definitions[0].selectionSet.selections = [documentClone.definitions[0].selectionSet.selections[0]]
+    return {
+      document: documentClone
+    }
+  })
+
+  await app.listen(0)
```
mcollina, line 620: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+  })
+
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.type(schema, GraphQLSchema)
+    t.deepEqual(document, parse(query))
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 563: not  neeeded

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.fail('this should not be called')
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 526: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+  app.graphql.addHook('preExecution', async (schema, document, context) => {
+    t.fail('this should not be called')
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    await sleep(1)
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 426: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+  })
+
+  app.graphql.addHook('preParsing', async (schema, source, context) => {
+    t.type(schema, GraphQLSchema)
+    t.is(source, query)
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 358: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+
+  app.graphql.addHook('preExecution', async (schema, operation, context) => {
+    t.fail('this should not be called')
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.fail('this should not be called')
+  })
+
+  await app.listen(0)
```
mcollina, line 321: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+  })
+
+  app.graphql.addHook('preValidation', async (schema, document, context) => {
+    t.type(schema, GraphQLSchema)
+    t.deepEqual(document, parse(query))
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 463: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+    t.type(schema, GraphQLSchema)
+    t.type(document, 'object')
+    t.type(context, 'object')
+    t.ok('preGatewayExecution called for foo error')
+    return {
+      errors: [new Error(`bar - ${document.definitions[0].name.value}`)]
+    }
+  })
+
+  await app.listen(0)
```
mcollina, line 863: not needed

test/gateway/hooks.js
```diff
@@ -0,0 +1,1076 @@
...
+    context.foo = 'bar'
+  })
+
+  app.graphql.addHook('onResolution', async (execution, context) => {
+    t.type(execution, 'object')
+    t.type(context, 'object')
+    t.is(context.foo, 'bar')
+  })
+
+  await app.listen(0)
```
mcollina, line 1037: not needed

test/hooks.js
```diff
@@ -0,0 +1,682 @@
...
+  })
+
+  app.graphql.addHook('onResolution', async function (execution, context) {
+    await sleep(1)
+    t.type(execution, 'object')
+    t.type(context, 'object')
+    t.ok('onResolution called')
+  })
+
+  await app.listen(0)
```
mcollina, line 86: not needed

lib/gateway/make-resolver.js
```diff
@@ -454,7 +463,8 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
     return reply[kEntityResolvers][`${service.name}Entity`]({
       query,
       variables,
-      originalRequestHeaders: reply.request.headers
+      originalRequestHeaders: reply.request.headers,
+      context
     }).then(transformData)
```
mcollina, line 468: please change then to `await`, `makeResolver` is now an async function (also a few lines above)

mcollina (commented):
Good work! I did another pass and I've left a few comments.

test/hooks.js
projectjudge: I wanted to have an explicit promise to await within the hooks function so we can simulate some work - similar to how fastify tests hooks. I'll remove it however, because it's not really needed for these test cases and what is currently there in the function definitions should be sufficient!

test/hooks.js
projectjudge: Nice, that's good to know!

lib/gateway/make-resolver.js
projectjudge: sure!

test/hooks.js
mcollina: You can do:

```js
const immediate = promisify(setImmediate).
```

projectjudge: ah nice - thanks! I'll push up the rest of these changes you've suggested now and get to work on adding thorough tests for subscriptions

lib/handlers.js
```diff
@@ -0,0 +1,57 @@
+const { hooksRunner, preExecutionHooksRunner, hookRunner, preParsingHookRunner, onResolutionHookRunner } = require('./hooks')
```
mcollina, line 1: use strict is missing

mcollina (commented):
Good work, one last nit.

lib/handlers.js
projectjudge: ah yes - fixed! :)

mcollina (changes requested):
Could you link the new docs pages from the README as well?

projectjudge:
> Could you link the new docs pages from the README as well?

Sure thing!

> Good work! How do the hooks play with subscriptions?

After a bit of experimentation, I think we need to add a few more hook events within `lib/subscription-connection.js`. From my observations with subscriptions so far, I have seen that `preGatewayExecution` is executed when using federated subscriptions. We don't see any other hooks triggered because subscriptions don't go through the normal request route. To fix this, I suggest we add `preParsing`, `preExecution` and `onResolution` (not `preValidation` because we don't do any validation in subscriptions) to `lib/subscription-connection.js` to make sure we capture the full GraphQL lifecycle for subscriptions.

My only concern with this approach is the meaning of context when applied to the hooks. Because it is not strictly associated with a request over the wire, one could argue that we should have subscription specific hooks for this situation instead. I wondered what your thoughts were on this?

Ps, apologies for my slower than normal responses, I'm on holiday this week! :)

mcollina (approved):
lgtm

PabloSzx (approved):

mcollina:
> My only concern with this approach is the meaning of context when applied to the hooks. Because it is not strictly associated with a request over the wire, one could argue that we should have subscription specific hooks for this situation instead. I wondered what your thoughts were on this?

It's better to have a subscription specific hook, ideally in a follow up pull request. This is good to ship right now, so I'm shipping it as a few people are waiting on this!

projectjudge:
Yep, makes sense - happy to do a follow up PR for this as you suggested! Thanks for shipping this afternoon! :)

docs/hooks.md
```diff
@@ -0,0 +1,144 @@
...
+  await registerTraceEvent()
+})
+```
+
+### preValidation
+
+By the time the `preValidation` hook triggers, the query string has been parsed into a GraphQL Document AST.
+
+```js
+fastify.addHook('preValidation', async (schema, document, context) => {
```
SimenB, line 47: should this be `fastify.graphql.addHook`? only the first example in this document has that, not the others

jonnydgreen: You're correct, yep - I'll submit a PR to fix this over lunch!

PR #388, opened 2021-02-10
Upgrade to fastify-websocket 3.0.0 for plugin encapsulation support

airhorns:
This uses the new version of `fastify-websocket` that has support for running hooks before upgrading to websockets inside mercurius, which means things like auth hooks will apply to both http and websocket routes registered by mercurius! Woop! See https://github.com/fastify/fastify-websocket/pull/95 for more details. This also means that custom `verifyClient` functions should be needed less frequently, as you can just rely on `preValidation` hooks wrapping mercurius to apply to all of it.

Only real change I had to made was removing the globally installed websocket handler that replied with an unknown route, which I don't think makes much sense to install anymore now that `fastify-websocket` uses the normal Fastify router and hooks system. 

I think this is semver major though because now hooks will apply to subscription routes registered by mercurius, whereas they wouldn't have before, and also this global not found route handler has been removed.

### DIFF
--- a/test/subscription.js
+++ b/test/subscription.js
@@ -598,43 +598,6 @@ test('subscription server sends update to subscriptions with custom context', t 
   })
 })
 
-test('subscription server register handle function arg is not empty', t => {
-  const app = Fastify()
-  const schema = `
-    type Query {
-      add(x: Int, y: Int): Int
-    }
-  `
-
-  const resolvers = {
-    add: async ({ x, y }) => x + y
-  }
-
-  t.tearDown(app.close)
-
-  app.register(GQL, {
-    schema,
-    resolvers,
-    subscription: true
-  })
-
-  app.listen(0, err => {
-    t.error(err)
-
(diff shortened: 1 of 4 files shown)


test/subscription.js
```diff
@@ -598,43 +598,6 @@ test('subscription server sends update to subscriptions with custom context', t
...
-  app.register(GQL, {
-    schema,
-    resolvers,
-    subscription: true
-  })
-
-  app.listen(0, err => {
-    t.error(err)
-
-    const url = 'ws://localhost:' + (app.server.address()).port + '/'
```
airhorns, line 624: This test was testing what happens if you connect to the root route with a websocket connection and expecting a not found response. I imagine this was mostly here for 100% test coverage, but I think that registering mercurius probably shouldn't have any bearing on the routes outside mercurius' context, so I just removed this test. `fastify-websocket` no longer supports the `handle` function during registration that creates the not found handler I believe.

mcollina (approved):
lgtm

PR #393, opened 2021-02-14
Call return on subscription resolver async iterator

### LINKED ISSUE #392: Closing out subscription does not execute return() on async iterator from subscription resolver
Hi

I am working on a new application and looking to use mercurius to serve up graphql subscriptions.  During development, I noticed that if I close the the subscription client either abruptly or gracefully by sending a `GQL_STOP` the underlying async iterator from the subscription resolver will keep running.  In my working program (unable to share this particular code as it is  private code from my work) - I am delegating to another remote graphql subscription running in a different service.  So in this case these delegated subscriptions are staying active within my mercurius application to the remote server even though the client has disconnected.  As I am unable to show that code, I have put together a small project which demonstrates this

https://github.com/moonkev/mercurius-sub-repro

If you run src/mercurius.js it will startup a server on port 4500.  
query - http://localhost:4500
subs - ws://localhost:4500

You can make a subscription using this very simple example (using any client, i.e. Altair, graphql-playground, etc).
```
subscription {
  messages { id }
}
```

If from your client you disconnect or try to unsubscribe gracefully, you will notice from the logging that the iterator continues to run.  I have included a counter-example using express-graphql.  You can start this by running src/express.js.  It will start on the same URLs as the mercurius example.  If you close the client connection here, you will see that the async iterators return method is called and the iterator will stop producing data.

The iterator itself is defined in src/schema.js.  Apologies if there is just something that I have misconfigured that can make this work.  I didn't see anything in the documentation about it.  Please let me know if any further information is needed, Thanks!

moonkev:
fixes #392

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -37,6 +37,7 @@ module.exports = class SubscriptionConnection {
     this.onConnect = onConnect
     this.onDisconnect = onDisconnect
     this.subscriptionContexts = new Map()
+    this.subscriptionIters = new Map()
     this.context = context
     this.isReady = false
     this.resolveContext = resolveContext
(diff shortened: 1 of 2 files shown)


moonkev:
will try to take a look at failing test tonight

mcollina (commented):

mcollina (commented):

mcollina (commented):

moonkev:
Have fixed broken tests and have also added a unit test

mcollina (approved):
lgtm

Issue #394, opened 2021-02-14
Tests intermittently failing


jonnydgreen:
Hi! As mentioned in this conversation https://github.com/mercurius-js/mercurius/pull/385#issuecomment-770115379, I've taken a closer look at the intermittently failing tests and identified a few potential fixes for the intermittent errors that I've come across:

 - For test [`subscription server works with fastify-websocket`](https://github.com/mercurius-js/mercurius/blob/master/test/subscription.js#L1719), there is a race condition where the test is ended before remaining checks can complete. Removing the `t.end()` should resolve this!
 - The test [`Polling schemas (subscriptions should be handled)`](https://github.com/mercurius-js/mercurius/blob/master/test/gateway/pollingInterval.js#L699) is not fully cleaned up in `t.tearDown`- should be fixed by adding this cleanup at the end of the test
 - For test [`gateway wsConnectionParams function is passed to SubscriptionClient`](https://github.com/mercurius-js/mercurius/blob/master/test/gateway/subscription.js#L337), the user service is closed before the gateway is closed. Switching this order in the `t.tearDown` should fix the intermittent errors in this case (and any other cases I can find in this test file)

With these proposed changes, running the following command `while npm run unit; do :; done` for ~15mins locally results in no errors!

I've already prepared the code for this so very happy to submit a PR for this if you are happy with the above approaches :)

jonnydgreen:
I have also noticed just now that the test [`https://github.com/mercurius-js/mercurius/blob/master/test/gateway/pollingInterval.js#L534`] also occasionally fails in the CI only with the following error:

```txt
# Subtest: Polling schemas (cache should be cleared)
        ok 1 - should be equivalent
        not ok 2 - should be equivalent
          ---
          diff: >
            --- expected
          
            +++ actual
          
            @@ -1,12 +1,14 @@
             Object {
            -  "data": null,
          
            +  "data": Object {
          
            +    "me": null,
          
            +  },
               "errors": Array [
                 Object {
                   "message": "Cannot query field \"me\" on type \"Query\". Did you mean \"me2\"?",
                   "locations": Array [
                     Object {
            -          "line": 3,
          
            -          "column": 11,
          
            +          "line": 2,
          
            +          "column": 3,
                     },
                   ],
                 },
```

Happy to take a look at this one as well :)

mcollina:
Amazing, could you send PRs?

jonnydgreen:
[...]

PR #395, opened 2021-02-14
Fix intermittent test failures

jonnydgreen:
Fixes: https://github.com/mercurius-js/mercurius/issues/394

### DIFF
--- a/test/gateway/hooks.js
+++ b/test/gateway/hooks.js
@@ -16,7 +16,7 @@ async function createTestService (t, schema, resolvers = {}) {
     federationMetadata: true
   })
   await service.listen(0)
-  return service
+  return [service, service.server.address().port]
 }
 
 const users = {
@@ -98,8 +98,7 @@ async function createTestGatewayServer (t) {
       }
     }
   }
-  const userService = await createTestService(t, userServiceSchema, userServiceResolvers)
-  const userServicePort = userService.server.address().port
+  const [userService, userServicePort] = await createTestService(t, userServiceSchema, userServiceResolvers)
 
   // Post service
   const postServiceSchema = `
@@ -137,8 +136,7 @@ async function createTestGatewayServer (t) {
       topPosts: (root, { count = 2 }) => Object.values(posts).slice(0, count)
     }
   }
-  const postService = await createTestService(t, postServiceSchema, postServiceResolvers)
-  const postServicePort = postService.server.address().port
+  const [postService, postServicePort] = await createTestService(t, postServiceSchema, postServiceResolvers)
 
   const gateway = Fastify()
   t.tearDown(async () => {
--- a/test/gateway/hooks.js
+++ b/test/gateway/hooks.js
@@ -2,7 +2,7 @@
 
 const { test } = require('tap')
 const Fastify = require('fastify')
-const { GraphQLSchema, parse } = require('graphql')
+const { GraphQLSchema } = require('graphql')
 const { promisify } = require('util')
 const GQL = require('../..')
 
@@ -178,7 +178,7 @@ test('gateway - hooks', async (t) => {
   app.graphql.addHook('preValidation', async function (schema, document, context) {
     await immediate()
     t.type(schema, GraphQLSchema)
-    t.deepEqual(document, parse(query))
+    t.type(document, 'object')
     t.type(context, 'object')
     t.ok('preValidation called')
   })
@@ -186,7 +186,7 @@ test('gateway - hooks', async (t) => {
   app.graphql.addHook('preExecution', async function (schema, document, context) {
     await immediate()
     t.type(schema, GraphQLSchema)
-    t.deepEqual(document, parse(query))
+    t.type(document, 'object')
     t.type(context, 'object')
     t.ok('preExecution called')
   })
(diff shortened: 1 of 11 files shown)


mcollina (approved):
lgtm

mcollina:
is this ready to go?

jonnydgreen:
The only thing this PR is missing is the failing test in the CI I just spotted now: https://github.com/mercurius-js/mercurius/issues/394#issuecomment-778858799

If you are happy for me to submit a solution for this test failure in a separate PR (I'm not sure how long it'll take to find the cause of the issue as it only seems to occur in the CI), this PR is good to go! :)

jonnydgreen:
Looking into it just now, I think it's due to not waiting enough for the pollingInterval changes to propagate (based on your solution in the same file for `Polling schemas (subscriptions should be handled)`):

Currently:

```js
await clock.tickAsync(2000)
```

Should be:

```js
await clock.tickAsync(10000)

// We need the event loop to actually spin twice to
// be able to propagate the change
await immediate()
await immediate()
```

I'll test this theory now/tomorrow!

jonnydgreen:
In my testing of the previous fixes, I found 3 more intermittent errors!
 -  Listener setup and stream writing causing timeouts in fastify-websocket subscription test
 - GraphQL parse sometimes timing out, so I have replaced this check with a simple type check as this doesn't affect the integrity of the tests
 - Error in subscriptions tests so I have made sure both ws clients are marked as ended in pollingInterval subscriptions

With these fixes, I have run the GitHub actions CI 10x with no more intermittent errors found on my fork, so I think this PR is ready to go :)

mcollina (approved):
lgtm

test/gateway/hooks.js
```diff
@@ -178,15 +178,15 @@ test('gateway - hooks', async (t) => {
   app.graphql.addHook('preValidation', async function (schema, document, context) {
     await immediate()
     t.type(schema, GraphQLSchema)
-    t.deepEqual(document, parse(query))
+    t.type(document, 'object')
```
mcollina, line 181: why this change? The previous assertion wanted to measure (deep) equality with the document and query.

jonnydgreen: I found that occasionally the GraphQL `parse` function times out when processing a query, leading to timeouts in the tests. This only occurred on the gateway code, so I changed this to a type assertion as we test this deep equality in other hook tests and therefore would be covered if anything changed in the hook runners.

As the query is a fixed string, I cannot think of a reason why this might be non-deterministic - other than issues in the `deepEqual` assertion or `parse` in some edge-case situations. I can take a deeper look at why the parse function is timing out occasionally, as I agree that ideally this should be retained if possible.

mcollina: that would be amazing, thanks

jonnydgreen: cool, no problem - looking into it now! :)

jonnydgreen: Good and bad news, I'm unable to replicate the hooks test error after ~15 CI runs so I will revert these changes :) However, I have found another test error in pollingInterval which occurs once in every 5 CI runs, so I will have a look at this as well!

jonnydgreen: Hi! Firstly, apologies for the delayed response, it's been quite a slow process debugging these intermittent tests as they only occur every 5 CI runs or so, which adds up to ~30min between each failed test which makes it hard to verify that my change has actually worked!

Quick update, I've got some code ready to push which tidies up some of the `t.tearDown` calls as a lot of them don't return the promiselike `.close()` function, e.g.

```js
t.tearDown(() => {
  app.close()
})
```
I've changed them in this file to handle the promise behaviour correctly by returning directly or awaiting in order within an async function. After implementing this change in just one of the test files, it seems to reduce the likelihood of the final intermittent error - I am fairly sure about this as I have just run the CI on my fork 20 times with no failures at all! 

For the pollingInterval test that is occasionally failing (this is the only problem test left), it seems like it is exiting the test early after the 8th assertion:

```txt
# Subtest: Polling schemas (subscriptions should be handled)
        ok 1 - should be equal
        ok 2 - should be equal
        ok 3 - should be equal
        ok 4 - should be equal
        ok 5 - should be equivalent
        ok 6 - should be equivalent
        ok 7 - should be equal
        ok 8 - should be equal
    # test count(5) != plan(null)
    # failed 1 of 5 tests
        # test count(8) != plan(null)
        # failed 1 of 8 tests
```

Somewhere after this [line](https://github.com/mercurius-js/mercurius/blob/master/test/gateway/pollingInterval.js#L947) and the next assertion I suspect one of the calls is exiting prematurely. If you have any suggestions as to why this particular test might be exiting every 20 CI runs or so on MacOS/Windows, I would be very grateful!

Finally, a quick scan of the rest of the tests shows a fair few are still not awaiting or returning the `app.close()` function, so wondered if you wanted me to sort this out as well (very happy to!)? It's worth noting that I think it would be a fairly sizeable change (39 results in 9 files) and maybe a bit of refactoring to make sure the gateways are closed before the federated services!

mcollina: The refactoring would be quite welcomed, thanks!

mcollina (approved):
lgtm

Amazing work!

jonnydgreen:
Thank you! :)

PR #396, opened 2021-02-15
fix test race condition, union types, & toc name

PabloSzx:
- Fix typescript types for unions are not following the real names in the code: https://github.com/mercurius-js/mercurius/blob/master/index.js#L294 https://github.com/mercurius-js/mercurius/blob/master/index.js#L319
- The TypeScript section is not in the main README TOC
- Subscriptions docs have a mistake in the TOC

### DIFF
--- a/docs/subscriptions.md
+++ b/docs/subscriptions.md
@@ -8,7 +8,7 @@
   - [Build a custom GraphQL context object for subscriptions](#build-a-custom-graphql-context-object-for-subscriptions)
   - [Subscription support (with redis)](#subscription-support-with-redis)
   - [Subscriptions with custom PubSub](#subscriptions-with-custom-pubsub)
-  - [Subscriptions with custom PubSub](#subscriptions-with-fastify-websocket)
+  - [Subscriptions with fastify-websocket](#subscriptions-with-fastify-websocket)
 
 ### Subscription support (simple)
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
Could you add/update a test for the type?

PabloSzx:
> Could you add/update a test for the type?

In practice it doesn't do much, since the code is not being run, and the issue was that it was not aligned, but if you really want to, sure

PabloSzx:
It turns out making tests for these types was a good thing, I found an interesting problem, the IFieldResolver type is conflicting with the __resolveType__ & __isTypeOf__, as the screenshots below show

 image

image

Honestly, I don't know how to fix it without making the types again / changing them considerably.

mcollina:
Me neither, maybe @fox1t can.

PabloSzx:
since the current fields can't be fixed without possible breaking changes, it's better to simply remove the misleading union types, since they were not working on the first place, and using [mercurius-codegen](https://github.com/mercurius-js/mercurius-typescript/tree/master/packages/mercurius-codegen) they are automatically fixed anyways.

I also fixed a race condition that was happening in the gateway/pollingInterval test

mcollina (approved):
lgtm

PR #399, opened 2021-02-18
fix: disabling cache breaks replaceSchema

### LINKED ISSUE #398: Disabling cache breaks replaceSchema
Passing `cache: false` as Plugin option makes `buildCache` return `null`, so LRUs instances are null.

But the `replaceSchema` method does not check if LRUs are null and throws
```
TypeError: Cannot read property 'clear' of null
```

It's necessary to wrap each LRU call with an if :
```
if (lru) {
  lru.clear()
}
```

Davide-Gheri:
Fixes #398

### DIFF
--- a/index.js
+++ b/index.js
@@ -256,9 +256,15 @@ const plugin = fp(async function (app, opts) {
 
     fastifyGraphQl.schema = s
 
-    lru.clear()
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #400, opened 2021-02-18
lower the keepalive timeout

mcollina:


aldis-ameriks:
@mcollina Could you share a short explanation about the motivation behind this change? I'd like to understand the consequences of using 1 minute vs 5s keepalive timeout in this context.

mcollina:
It reduces the chances of the socket being closed by the other side, and causing an error. This will create more sockets under low load and it's a great tradeoff.

PR #401, opened 2021-02-18
fix: custom entity cache id

### LINKED ISSUE #397: Unstable gateway entities resolution
I'm experiencing an instability from the gateway:

Once the gateway is started, some queries that need to resolve a shared entity (with the _entities query) fail with a `SocketError: other side closed`. The error is random and other queries that depends on the same service are working.
And there is no error on the service side (running on fastify/mercurius - healthchecks working) as the gateway doesn't send any request to the service. I also noticed that the CPU usage from the gateway skyrockets when this issue occurs.

Error:
```json
{
  "stack":"SocketError: other side closed\n    at Socket.onSocketEnd (/app/node_modules/mercurius/node_modules/undici/lib/core/client.js:773:22)\n    at Socket.emit (events.js:327:22)\n    at endReadableNT (internal/streams/readable.js:1327:12)\n    at processTicksAndRejections (internal/process/task_queues.js:80:21)",
  "msg":"other side closed"
}
```
It occurs after the gateway receives the main query response and when it needs to resolve entities, it produces one error for each entity it should resolve.

I'm unable to reproduce this error locally nor with a simple test case. I don't understand why it occurs and it doesn't seem related to a network issue between the gateway and the services because other queries are working properly and the issue occurs more and more often the longer the gateway runs (at startup it works as expected). The strangest part is that other queries that depends on the same entities type resolution works ...

I can not share the original query but here is a similar query structure:
```
query {
  resources {
    id
    user { // Reference to external type
      id
      username
    }
    configurations {
      id
      otherReference { // Reference to external type
         id
         name
         status
      }
    }
  }
}
```

Notable points that may help to pinpoint the issue:
- The query that fails needs to resolve two references from two different services
- The query works well on startup ( for ~1-2min) and locally (always)
- When the query fails other queries depending on the same service works, same for entities resolution
- it seems related to network fluctuations/delays because I'm unable to reproduce the error locally
- The error occurs randomly to one of the entity type that needs to be resolved, sometimes both
- gateway CPU goes crazy once the error is triggered (but not instantly)

I'm sorry if the issue doesn't have enough details yet, but I'm stuck and any help would be really appreciated ! I'll keep digging and try to reproduce the error locally to isolate a test case.

PacoDu:
may resolve #397, or at least the CPU usage issue

This PR changes the id used for the entity resolver caching. It uses the query path instead of the full query object which could be quite heavy to serialize.

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -314,7 +314,7 @@ async function buildGateway (gatewayOpts, app) {
       }))
 
       return result
-    })
+    }, query => query.id)
   }
 
   typeToServiceMap.Query = null
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -359,7 +359,8 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
 
     if (isReference && !parent[fieldName]) return null
 
-    const resolverKey = generatePathKey(info.path).join('.').replace(/\d/g, '_IDX_')
+    const queryId = generatePathKey(info.path).join('.')
+    const resolverKey = queryId.replace(/\d/g, '_IDX_')
     const { reply, __currentQuery, lruGatewayResolvers, pubsub } = context
 
     const cached = lruGatewayResolvers.get(`${__currentQuery}_${resolverKey}`)
@@ -466,7 +467,8 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
       query,
       variables,
       originalRequestHeaders: reply.request.headers,
-      context
+      context,
+      id: queryId
     })
 
     return transformData(response)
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

jonnydgreen:
Hi! Just been running the tests locally in order to fix a typo in the hooks documentation and am not getting 100% test coverage because this line in the `gateway.js` file: https://github.com/mercurius-js/mercurius/blob/master/lib/gateway.js#L317

I just wondered if you're also getting this? Am very happy to write the test myself for it, and if you'd like me to do it, I just wondered if you had any examples of queries for this fix! Thanks!

mcollina:
Go for it! I 'm getting 100% code coverage on my machine.

jonnydgreen:
Just ran the tests again after reinstalling the node_modules (I missed that there was a package.json update in this!) and getting 100% coverage now - apologies! :)

PR #405, opened 2021-02-21
fix(types): missing reply in context

### LINKED ISSUE #404: [types] reply object is missing in context
Hi.

In types `context` (`MercuriusContext`) doesn't contain `reply` object. Likely in js it's done here: https://github.com/mercurius-js/mercurius/blob/master/index.js#L242

Also in test `request` and `reply` are just declared in this interface https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts#L53, so they should be present in it?

So is missing `reply` a bug or it's done for a reason?

SkeLLLa:
Adds missing reply type to context. No need to pass it in `context` function, since it's passed already,

Closes #404

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -26,6 +26,7 @@ export interface PubSub {
 
 export interface MercuriusContext {
   app: FastifyInstance;
+  reply: FastifyReply;
   /**
    * __Caution__: Only available if `subscriptions` are enabled
    */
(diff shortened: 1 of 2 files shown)


PabloSzx (approved):

mcollina (approved):
lgtm

Issue #406, opened 2021-02-23
Add subscription hooks to Mercurius


jonnydgreen:
Hi! As discussed in the original [hooks PR](https://github.com/mercurius-js/mercurius/pull/385#issuecomment-771872613), we could add GraphQL subscription hooks to Mercurius so we can process data at different points in the subscription lifecycle.

## Proposed solution

The subscription hooks to add will be:

- `preSubscriptionParsing`: before subscriptions are extracted from the input websocket data. It will contain the following data:
        - `schema`
        - `source`
        - `context`
- `preSubscriptionExecution`: before subscriptions are executed on the server. It will contain the following data:
        - `schema`
        - `document`
        - `context`
- `preGatewaySubscriptionExecution`: before subscriptions are created with federated services in the gateway. It will contain the following data:
        - `schema`
        - `document`
        - `context`
- `onSubscriptionResolution`: every time subscription data is received in the connection. It will contain the following data:
	- `execution`
	- `context`

## Example lifecycle

```
Incoming GraphQL Websocket subscription data
  │
  └─▶ Routing
           │
  errors ◀─┴─▶ preSubscriptionParsing Hook
                  │
         errors ◀─┴─▶ Subscription Parsing
                        │
               errors ◀─┴─▶ preSubscriptionExecution Hook
                              │
                     errors ◀─┴─▶ Subscription Execution
                                    │
                                    │
                                    └─▶ preGatewaySubscriptionExecution Hook(s)
                                            │
                                   errors ◀─┴─▶ Gateway Subscription Execution(s)
                                                  │
                                      wait for subscription data
                                                  │
                   subscription closed on error ◀─┴─▶ Subscription Resolution (when subscription data is received)
                                                        │
                                                        └─▶ onSubscriptionResolution Hook
```

All new code with be 100% covered by tests, documentation will be provided and the new hooks will be correctly typed. I have started work on an implementation of this and can submit a PR for this once everyone is agreed on the solution! :)

mcollina:
Go for it!

PR #407, opened 2021-02-24
Add subscription hooks to Mercurius

jonnydgreen:
Closes: https://github.com/mercurius-js/mercurius/issues/406

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -212,7 +212,7 @@ module.exports = class SubscriptionConnection {
           await onSubscriptionResolutionHandler({ execution: value, context })
         } catch (error) {
           this.fastify.log.error(error)
-          this.handleConnectionClose()
+          return this.handleConnectionClose()
         }
       }
       this.sendMessage(GQL_DATA, data.id, value)
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -17,7 +17,7 @@ const {
   GQL_STOP
 } = require('./subscription-protocol')
 const { MER_ERR_GQL_SUBSCRIPTION_FORBIDDEN, MER_ERR_GQL_SUBSCRIPTION_UNKNOWN_EXTENSION } = require('./errors')
-const { preSubscriptionParsingHandler, onSubscriptionResolutionHandler, preSubscriptionExecutionHandler } = require('./handlers')
+const { preSubscriptionParsingHandler, onSubscriptionResolutionHandler, preSubscriptionExecutionHandler, onSubscriptionEndHandler } = require('./handlers')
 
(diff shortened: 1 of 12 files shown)


mcollina (commented):
Good work! Why did you add all the nextTick?

jonnydgreen:
Thanks! For `nextTick`, I followed the Node.js docs examples that used `once` as I wanted to guarantee that I was listening with `once` before the event was fired: https://nodejs.org/api/events.html#events_events_once_emitter_name_options

test/subscription-hooks.js
```diff
@@ -0,0 +1,335 @@
...
+      type: 'connection_init'
+    }))
+    client.write(JSON.stringify({
+      id: 1,
+      type: 'start',
+      payload: {
+        query
+      }
+    }))
+  })
```
mcollina, line 150: there is no need for this `nextTick`.

test/subscription-hooks.js
```diff
@@ -0,0 +1,335 @@
...
+
+  {
+    const [chunk] = await once(client, 'data')
+    const data = JSON.parse(chunk)
+    t.is(data.type, 'connection_ack')
+  }
+
+  process.nextTick(() => {
+    sendTestMutation(app)
+  })
```
mcollina, line 160: there is no need for this nextTick

mcollina (changes requested):
I think you can safely remove most (if not all) `nextTick`. When you write to a socket the messages go through a few full event loop cycles.

jonnydgreen:
Makes sense - I'll get on that now!

lib/subscription-connection.js
```diff
@@ -196,6 +206,15 @@ module.exports = class SubscriptionConnection {
 
     // TODO implement backpressure
     for await (const value of subIter) {
+      // Trigger onSubscriptionResolution hook
+      if (this.context.onSubscriptionResolution !== null) {
+        try {
+          await onSubscriptionResolutionHandler({ execution: value, context })
+        } catch (error) {
+          this.fastify.log.error(error)
+          this.handleConnectionClose()
```
mcollina, line 215: I think we should return after calling this function.

mcollina (changes requested):

mcollina (approved):
lgtm

PacoDu (commented):
lgtm ! Maybe we could also add a hook when the subscription ends ? (when `GQL_STOP` is received)

jonnydgreen:
For sure - I'll add that later along with the updated lifecycle! :)

PacoDu (approved):
lgtm !

mcollina (approved):
lgtm

PR #410, opened 2021-02-24
fix(types): mark 'schema' optional in 'MercuriusSchemaOptions'

mrazauskas:
At the moment arg 'schema' is not optional in type 'MercuriusSchemaOptions'. Because of this one gets and error if plugin is registered with 'schema' in options. An example of code which will throw Typescript error:

```
app.register(mercurius, { path: '/' })

app.register(async function (app) {
  app.graphql.extendSchema(schema)
})
```

You can also see the Typescript error in included test.

The code works as expected. Seems like typings are incorrect.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -271,7 +271,7 @@ export interface MercuriusSchemaOptions {
   /**
    * The GraphQL schema. String schema will be parsed
    */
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #411, opened 2021-02-27
docs: Link to Examples folder added

mrazauskas:
Perhaps it makes sense to delete 'See test.js for more examples…' and to add a link to Examples folder?

Also I am not so sure if we need two rather similar examples here. The one which I renamed 'Quick Start' might be just enough for a quick start. 'Executable Schema' could be moved to Examples folder. Less nesting in the structure:

- Install
- Quick Start
- More Examples
- Acknowledgements
- License

If this looks good, I can improve it.


mcollina:
Looks good! Keep working on it!

mrazauskas:
In Examples sections it would be nice to have some brief snippets which could illustrated Features mentioned on top of the page. Just to show the possibilities. Plus a link of full examples.

I moved Table of Contents up. Feels like it makes sense to go from Install to Quick Start.

It felt like some headline above TOC would look good. Table of Contents seemed to dry. Documentation is fine, but formal. What about Docs?

Also I noticed that some sections were visible only in the sidebar, but missing in TOC. Added these.

All headlines are in Title Case now.

mcollina:
> In Examples sections it would be nice to have some brief snippets which could illustrated Features mentioned on top of the page. Just to show the possibilities. Plus a link of full examples.

+1, would you work on those as well?
> 
> 
> I moved Table of Contents up. Feels like it makes sense to go from Install to Quick Start.
> 

+1
> 
> It felt like some headline above TOC would look good. Table of Contents seemed to dry. Documentation is fine, but formal. What about Docs?
> 

I think Documentation is fine.


> Also I noticed that some sections were visible only in the sidebar, but missing in TOC. Added these.

+1
> 
> 
> All headlines are in Title Case now.

+1

mrazauskas:
Sure. I will work code snippets. Will be back in few days with some draft.

mrazauskas:
Just uploaded logo files. Will continue to work on this PR.

Perhaps it makes sense to include logo in 'README.md' here too? Or better to send another PR?

Also while going through docs, I found few broken links. Seems like they were introduced after docs were spilt into several files. And there are few other things to clean up.

In this case it is better to make one larger clean-up PR, right? This isn’t any urgent bug to fix. Better to work longer and to make one solid PR.

mcollina (approved):
lgtm

mcollina:
Let's land this and iterrate!

PR #420, opened 2021-03-10
remove Window type

### LINKED ISSUE #419: mercurius references `Window` type that is not always available
Mercurius uses types from `Window`. That type is not available in typescript
unless the `lib` setting is set to `DOM` or is omitted.

See testcase here: https://github.com/runeh/mercurius-window-bug-repro

Steps to reproduce:

- `npm install`
- `npm run build`

Expected:

Compiles

Actual:

Typescript fails with

```
node_modules/mercurius/index.d.ts:545:33 - error TS2304: Cannot find name 'Window'.

545   playgroundHeaders?: ((window: Window) => object) | object;
                                    ~~~~~~
```

mcollina:
Fixes #419

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -2,7 +2,6 @@
   "compilerOptions": {
     "target": "es2018",
     "module": "commonjs",
-    "lib": ["DOM"],
     "noEmit": true,
     "strict": true,
     "esModuleInterop": true


SimenB:
@mcollina thoughts on removing it from your tsconfig? In that case I think it'd be an error.

https://github.com/mercurius-js/mercurius/blob/0e207b498921bde55e0ae2f6c5ac3bc127b5932e/tsconfig.json#L5

mcollina:
removed, good spot!

PR #421, opened 2021-03-14
Fix: Update docsify sidebar hash links

SimonJThompson:
- Minor fix for docsify site
- First group of links only used hashes without a path, meaning they broke on subpages


mcollina (approved):
lgtm

PR #425, opened 2021-03-16
fix: improve error messaging for gateway

jkirkpatrick24:
# Summary

Adds some additional error messaging for gateways that are configured to consume remote services. Previously, if a gateway loaded a remote service and that service had an invalid schema, the gateway would attempt to build a federated schema  and the result of that is a GQL parsing error: `Syntax Error: Unexpected <EOF>.`. 

This pull request adds an additional check before attempting to build the federated schema. If there are no valid SDLs that have been provided to the gateway, we inform the user that this is the case. 

I also added a couple of tests that check the error handling behaviour of a gateway that is consuming remote services.

# Caveats

This is probably more useful for development, but I though this messaging might help other developers locate similar issues faster than it took me!

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -171,9 +171,16 @@ async function buildGateway (gatewayOpts, app) {
   const { services, errorHandler = defaultErrorHandler } = gatewayOpts
 
   const serviceMap = await buildServiceMap(services, errorHandler)
-  const serviceSDLs = Object.values(serviceMap)
-    .map(service => service.schemaDefinition)
-    .filter(schemaDefinition => schemaDefinition !== '')
+
+  const serviceSDLs = Object.entries(serviceMap).reduce((acc, [name, value]) => {
+    const { schemaDefinition, error } = value
+
+    error !== null
+      ? app.log.warn(`Initializing service "${name}" failed with message: "${error.message}"`)
+      : acc.push(schemaDefinition)
+
+    return acc
+  }, [])
 
   if (serviceSDLs.length < 1) {
     throw new MER_ERR_GQL_GATEWAY_INIT('No valid service SDLs were provided')
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -17,7 +17,7 @@ const {
   createEntityReferenceResolverOperation,
   kEntityResolvers
 } = require('./gateway/make-resolver')
-const { MER_ERR_GQL_GATEWAY_REFRESH } = require('./errors')
+const { MER_ERR_GQL_GATEWAY_REFRESH, MER_ERR_GQL_GATEWAY_INIT } = require('./errors')
 const { preGatewayExecutionHandler } = require('./handlers')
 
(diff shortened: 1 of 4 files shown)


test/gateway/remote-services.js
```diff
@@ -0,0 +1,113 @@
+const { test } = require('tap')
```
mcollina, line 1: add use strict at the top

lib/gateway.js
```diff
@@ -171,8 +171,13 @@ async function buildGateway (gatewayOpts, app) {
   const { services, errorHandler = defaultErrorHandler } = gatewayOpts
 
   const serviceMap = await buildServiceMap(services, errorHandler)
+  const serviceSDLs = Object.values(serviceMap)
+    .map(service => service.schemaDefinition)
+    .filter(schemaDefinition => schemaDefinition !== '')
```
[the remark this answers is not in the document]
mcollina: I would use fastify.log.warn.

mcollina (changes requested):
Good work!

test/gateway/remote-services.js
jkirkpatrick24: 👍

jkirkpatrick24:
@PabloSzx and @mcollina thank you for the review! I've updated to add the warning when a service fails to initialize. Is that in line with what you were thinking?

test/gateway/remote-services.js
```diff
@@ -0,0 +1,115 @@
...
+  } catch (err) {
+    t.is(err.message, 'Unknown type "World".')
+  }
+})
+
+test('Does not error if at least one service schema is valid', async (t) => {
+  const [service, servicePort] = await createRemoteService(validSchema)
+  const [invalidService, invalidServicePort] = await createRemoteService(invalidSchema)
+
+  const gateway = Fastify()
```
PabloSzx, line 87:
```suggestion
  const gateway = Fastify({
    logger: true
  })

  let warnCalled = 0;
  gateway.log.warn = (message) => {
    warnCalled++;
    t.matchSnapshot(message)
  }

```

Testing that the warning is actually called

test/gateway/remote-services.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+      ]
+    }
+  })
+
+  try {
+    await gateway.ready()
+  } catch (err) {
+    t.error(err)
+  }
```
PabloSzx, line 114:
```suggestion
  }
  t.equal(warnCalled, 1, "Warning is called")
```

PabloSzx (changes requested):
After that minor test improvement, everything looks good, thanks 👍

mcollina (approved):
lgtm

jkirkpatrick24:
Thanks again @mcollina and @PabloSzx for the reviews. I've added the suggested changes. If theres anything else needed let me know 😄

PabloSzx (approved):
looks good 👍

PR #427, opened 2021-03-23
add __currentQuery to subscription context

oleksiipysanko:


lib/subscription-connection.js
```diff
@@ -187,6 +187,7 @@ module.exports = class SubscriptionConnection {
       {}, // rootValue
       {
         ...context,
+        __currentQuery: print(document),
```
mcollina, line 190: I think this would be too costly to do it all the time. Maybe we can do it as a getter instead?

mcollina (changes requested):

oleksiipysanko:
Hi @mcollina,

Yes, I didn't think it through yet - I'll add specific case a bit later. Also, it probably should be noted somewhere that in order to receive different context subscriptions need to have parameters, like: 

```
subscription {
 testEvent(id: ID!) {
  userId
 }
}
```

And by supplying different ids client will be able to obtain different data streams through gateway.

lib/subscription-connection.js
oleksiipysanko: I didn't dig into subscription protocols, but as far as I can understand, there can be some cases when query arrives as a DocumentNode straight away. My initial thought was just to ignore this case and set __currentQuery when query is string - can it work?

Don't quite understood about getter - where?

mcollina: > I didn't dig into subscription protocols, but as far as I can understand, there can be some cases when query arrives as a DocumentNode straight away. My initial thought was just to ignore this case and set __currentQuery when query is string - can it work?

To be honest I'm not sure. Maybe add some tests.

> Didn't quite understood about getter - where?

Instead of calling `print(document)` when the object is created, add a getter to said object and call `print()` when it is accessed. In this way, if no one access `__currentQuery`, there will not overhead.

oleksiipysanko: Understood, thank you, I'll change it to a getter.

oleksiipysanko:
We've encountered another problem with gateway subscriptions: we're using service types with federation to resolve events the same as with usual queries/mutations. 

Cache gets updated for each request to gateway (https://github.com/mercurius-js/mercurius/blob/f87a25ad3f712b0a894d844517590fab0564f878/lib/gateway.js#L195), but for subscriptions it is set only once on GQL start (https://github.com/mercurius-js/mercurius/blob/9647d2f049f854a11658e0aec04ff01c1379371e/lib/subscription.js#L45). 

Therefore, in subscriptions, first resolver execution will be cached and returned for any succeeding event afterwards. There is no way to alter context per subscription execution in graphql.js subscribe, so I've just added .create() call for subscription resolvers - it fixes the problem. I don't think that it can create any performance problems, but, please, let me know otherwise.

I will add tests for this case as well as soon as I'll have more time.

lib/subscription-connection.js
```diff
@@ -133,6 +133,10 @@ module.exports = class SubscriptionConnection {
 
     this.context._connectionInit = data.payload
 
+    const { payload: { headers = {} } = {} } = data
```
mcollina, line 136: please avoid such level of destructuring with default values.

mcollina (changes requested):

oleksiipysanko:
This next problem was also related to subscription federation requests. 

I guess it is more a feature than a bug - passing headers from subscription init payload to gateway requests invoked by this subscription to have access to a context (e.g. authorisation) in underlying services.

I have a bit tight schedule right now, but I'll add getter and unit tests asap, so this pr will be eligible for merge.

On a better side - we're successfully using mercurius gateway subscriptions with federation in our project now without any problems.

0x77dev:
@mcollina Unit tests for all cases added.

mcollina (commented):
Should we add some docs?

Seems good so far!

oleksiipysanko:
@mcollina maybe [this](https://github.com/mercurius-js/mercurius/pull/427#issuecomment-804818316) can be mentioned, but only in a context of multiple clients connecting with the same query (then they need to supply an argument as per comment). 

The headers passing is just an improvement for subscription -> federation scenario, not sure if it worth to mention, since this PR fixes an edge case.

lib/subscription-connection.js
mcollina: This needs work!

mcollina (approved):
lgtm

jonnydgreen (approved):
LGTM!

PR #452, opened 2021-04-06
feat(gateway): extend undici agent options

PacoDu:
This PR extends the options available for the gateway services clients (undici) and the documentation.


mcollina (approved):
lgtm

PR #454, opened 2021-04-08
Access context in rewriteHeaders

### LINKED ISSUE #431: Access context in rewriteHeaders
Hi, first of all, kudos for this awesome repo.

I was implementing the gateway but realised we cannot access context information in `rewriteHeaders`, it's quite a broad use case to pass  headers from context info to underlying services, Apollo allows it  in `RemoteGraphQLDataSource` by providing context in `willSendRequest`.

alex-parra:
Fix #431 

On gateway setups, query `context` was not being exposed to `rewriteHeaders`.
This PR fixes the issue by passing the `context` to every call to `serviceDefinition.sendRequest` which then exposes it to `rewriteHeaders` as an additional parameter.

### DIFF
--- a/docs/federation.md
+++ b/docs/federation.md
@@ -168,7 +168,7 @@ gateway.register(mercurius, {
       {
         name: 'user',
         url: 'http://localhost:4001/graphql',
-        rewriteHeaders: (headers, context = undefined) => {
+        rewriteHeaders: (headers, context) => {
           if (headers.authorization) {
             return {
               authorization: headers.authorization
--- a/docs/federation.md
+++ b/docs/federation.md
@@ -168,7 +168,7 @@ gateway.register(mercurius, {
       {
         name: 'user',
         url: 'http://localhost:4001/graphql',
-        rewriteHeaders: (headers, context) => {
+        rewriteHeaders: (headers, context = undefined) => {
           if (headers.authorization) {
             return {
               authorization: headers.authorization
--- a/docs/federation.md
+++ b/docs/federation.md
@@ -168,7 +168,7 @@ gateway.register(mercurius, {
       {
         name: 'user',
         url: 'http://localhost:4001/graphql',
-        rewriteHeaders: (headers) => {
+        rewriteHeaders: (headers, context) => {
           if (headers.authorization) {
             return {
               authorization: headers.authorization
--- a/lib/gateway/service-map.js
+++ b/lib/gateway/service-map.js
@@ -85,8 +85,7 @@ async function getRemoteSchemaDefinition (serviceConfig, initHeaders) {
         }
         `
     }),
-    headers,
-    context: {} // No context to pass at this stage
+    headers
   })
 
   const {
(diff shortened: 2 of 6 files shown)


lib/gateway/service-map.js
```diff
@@ -85,7 +85,8 @@ async function getRemoteSchemaDefinition (serviceConfig, initHeaders) {
         }
         `
     }),
-    headers
+    headers,
+    context: {} // No context to pass at this stage
```
simoneb, line 89: if there's no context, do we need to provide a fake one then or would leaving it undefined be a more sensible default than providing an empty object then?

test/gateway/rewrite-headers.js
```diff
@@ -0,0 +1,101 @@
+'use strict'
```
simoneb, line 1: maybe I'm missing it, but I don't see where this file tests that rewriteHeaders is actually called and, more importantly, that its return value is actually used.

alex-parra: Tests to `rewriteHeaders` are at line 64 and 65

mcollina (approved):
lgtm

docs/federation.md
```diff
@@ -168,7 +168,7 @@ gateway.register(mercurius, {
       {
         name: 'user',
         url: 'http://localhost:4001/graphql',
-        rewriteHeaders: (headers) => {
+        rewriteHeaders: (headers, context = undefined) => {
```
simoneb, line 171:
```suggestion
        rewriteHeaders: (headers, context) => {
```

simoneb (approved):
LGTM, see inline comment.

Santinell:
@alex-parra @mcollina 
I thought that in `rewriteHeaders` I will have the same context as generated in context method of gateway.
Like this:
```
 const rewriteHeaders = (headers, context) => {
    // I expect that here I will have the same context as in method `context` followed below
    return { context: JSON.stringify(context) };
  };
  app.register(mercurius, {
    context: async req => {
      const ctx = await authByHeaders(req.headers);
      return ctx;
    },
    gateway: {
      pollingInterval: 3000,
      services: [
        { name: 'host', url: 'http://localhost:3001/', rewriteHeaders },
        { name: 'dict', url: 'http://localhost:3002/', rewriteHeaders }
      ]
    }
  });
```

My use-case is auth on gateway side and sending context to services via `rewriteHeaders`

For now I have to implement auth on all services, even which don't use database for its methods. It's a huge overhead in packages and runtime, when one query using 3-5 services and every of them should auth.

Is it possible to get data from context method to `rewriteHeaders`?
Converting method `rewriteHeaders` to asynchronous is also can make it working.

mcollina:
I _think_ it should be the same context. Please open a new issue with a full reproducible example. Thanks.

Santinell:
I am sorry, problem was because sometimes context is undefined. I haven't expected such value and it lead to error on request its fields.
When I made it empty object by default this problem was resolved

PR #461, opened 2021-04-20
Add support for Mercurius auth plugin

jonnydgreen:
This PR adds a new application lifecycle hook as a result of this discussion in Mercurius Auth: https://github.com/mercurius-js/auth/pull/3#discussion_r615405536

It adds the following:
 - Add `onGatewayReplaceSchema` application lifecycle hook
 - Fixes an issue where the `preExecution` hook was not run when the query has been jit'd. I found this when improving the performance in the Mercurius Auth benchmarks

It also includes a fix for the slow `rewriteHeaders` tests which were taking ~8s to run. I've noticed that `pollingInterval` and `remote-services` tests also take a long time. Looking at the code, I think it's to do with the way the gateway is closed when it errors upon building the schema - however I need to confirm this. Very happy to take a long at this in a follow up PR when I have some spare time next week! :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -495,19 +495,10 @@ const plugin = fp(async function (app, opts) {
       }
     }
 
-    // minJit is 0 by default
-    if (cached && cached.count++ === minJit) {
-      cached.jit = compileQuery(fastifyGraphQl.schema, document, operationName)
-    }
-
-    if (cached && cached.jit !== null) {
-      const execution = await cached.jit.query(root, context, variables || {})
-
-      return maybeFormatErrors(execution, context)
-    }
+    const shouldCompileJit = cached && cached.count++ === minJit
 
     // Validate variables
-    if (variables !== undefined) {
+    if (variables !== undefined && !shouldCompileJit) {
       const executionContext = buildExecutionContext(fastifyGraphQl.schema, document, root, context, variables, operationName)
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -179,9 +179,18 @@ const plugin = fp(async function (app, opts) {
...
+            const context = assignApplicationLifecycleHooksToContext(fastifyGraphQl[kHooks], {})
+            const schema = await gateway.refresh()
+            if (schema !== null) {
+              // Trigger onGatewayReplaceSchema hook
+              if (context.onGatewayReplaceSchema !== null) {
+                await onGatewayReplaceSchemaHandler(context, { instance: app, schema })
+              }
+              fastifyGraphQl.replaceSchema(schema)
+            }
+          } catch (error) {
```
jonnydgreen, line 192: Drawing attention to this. I added the catch and subsequent log because when testing hooks error handling, I was getting an unhandled Promise rejection because a promise is passed to setInterval and this situation was previously untested (we would get the same problem if `gateway.refresh` errored).

jonnydgreen:
For the failed tests, seems like an npm package can't be found in the `npm install` - maybe a rerun might fix this (I just tested on my machine and it grabbed it fine)?

mcollina (approved):
lgtm

PR #468, opened 2021-04-26
fix #467

### LINKED ISSUE #467: [Bug] onResolution hook extensions do nothing.
While trying to add extensions to my GraphQL return data as mentioned in #466, I tried using mercurius `onResoluton` hook to do so, as it has a `exec.extensions` property, but when I get my results back, I don't have any extensions.

What I expect is for there to be extensions data.

A minimum code example:

```ts
import fastify, { FastifyInstance } from 'fastify'
import mercurius from 'mercurius'

const server: FastifyInstance = fastify()

await server.register(mercurius, {
	schema,
	graphiql: true,	
})

await server.graphql.addHook('onResolution', async (exec, context) => {
	exec.extensions = {
		test: 'Test Extension',
	}

	await Promise.resolve()
})
```

Sven65:
fix #467 

Fixes the issue where extensions data wouldn't follow along in the GraphQL response.

The issue had to do with the `responseProperties` object in `routes.js` not having a field for the extensions, causing it to be omitted.

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -292,9 +292,7 @@ module.exports = async function (app, opts) {
       ))
     } else {
       // Regular query
-      const data = execute(request.body, request, reply)
-
-      return data
+      return execute(request.body, request, reply)
     }
   })
 
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -40,12 +40,17 @@ const responseProperties = {
         extensions: {
           type: 'object',
           properties: {
+
             code: { type: 'string' }
           },
           additionalProperties: true
         }
       }
     }
+  },
+  extensions: {
+    type: 'object',
+    additionalProperties: true
   }
 }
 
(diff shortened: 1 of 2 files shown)


lib/routes.js
```diff
@@ -287,7 +292,9 @@ module.exports = async function (app, opts) {
       ))
     } else {
       // Regular query
-      return execute(request.body, request, reply)
+      const data = execute(request.body, request, reply)
+
+      return data
```
mcollina, line 297: I think this change is not needed.

mcollina (commented):
Good work! The solution was actually very simple.

lib/routes.js
Sven65: Sorry, missed restoring this from my testing, fixed it now!

mcollina (approved):
lgtm

PabloSzx (approved):

Sven65:
Are the checks supposed to fail like that?

simone-sanfratello (approved):

PR #472, opened 2021-04-28
fix mergeInfo type conflict

### LINKED ISSUE #471: Subscriptions and withFilter Typings error
## TL;DR
Typings error having to do with subscriptions when using `withFilter`.  Not sure if I'm doing something wrong or the generated typings are wrong.  Am willing open to open PR if necessary for fix.  Am somewhat new to graphql.
### Problem
Given the following code
```ts
export const resolvers: Resolvers = {
   Subscription: {
          somethingChanged: {
              subscribe: withFilter(
                  (parent, args, ctx) => ctx.pubsub.subscribe('somethingChanged'),
                  (payload, args, ctx) => Something.isRelevant(args.id, payload)
              ),
          },
      },
}
```

I'm getting the following typing error on the subscribe field
```
Types of parameters 'info' and 'info' are incompatible.
    Type 'GraphQLResolveInfo' is not assignable to type 'GraphQLResolveInfo & { mergeInfo: MergeInfo; }'.
      Property 'mergeInfo' is missing in type 'GraphQLResolveInfo' but required in type '{ mergeInfo: MergeInfo; }'.ts(2322)
```
### Research
I looked in the [typings file](https://github.com/mercurius-js/mercurius/blob/master/index.d.ts#L659) and found that the `withFilter` function exports the info parameter type as follows
```ts
  const withFilter: <
   /** omitted for brevity **/
  >(
    /** omitted for brevity **/,
    filterFn: (
        /** omitted for brevity **/
      info: GraphQLResolveInfo & {
        mergeInfo: MergeInfo
      }
    ) => boolean | Promise<boolean>
  ) => (
    /** omitted for brevity **/,
    info: GraphQLResolveInfo & {
      mergeInfo: MergeInfo
    }
  ) => AsyncGenerator<TPayload>
```
The above shows that when using the `withFilter` function, the function itself is required to be called with the `info` parameter having the  `mergeInfo` field.  

I looked at the [implementation](https://github.com/mercurius-js/mercurius/blob/master/lib/subscriber.js#L78) of the `withFilter` function and it doesn't seem to modify the info parameter that is originally passed into the subscriber itself.  The subscriber itself takes in an info parameter that is purely of type `GraphQLResolveInfo` at least from looking at the generated typings when using `mercurius-codegen`

```ts
export type SubscriptionSubscribeFn<TResult, TParent, TContext, TArgs> = (
    /** omitted for brevity **/
    info: GraphQLResolveInfo
) => AsyncIterator<TResult> | Promise<AsyncIterator<TResult>>;

export type SubscriptionResolveFn<TResult, TParent, TContext, TArgs> = (
   /** omitted for brevity **/
    info: GraphQLResolveInfo
) => TResult | Promise<TResult>;
```
### Question

How do I go about solving this typing issue?  Did I make a mistake myself or is there an issue in the generated typings?  I'm more than happy to open a PR fixing an issue if that is required to fix the issue. 

For reference, I'm a little new to graphql (been tinkering with it for a few weeks now) and am using this library because I need subscription support and apollo's subscription support looked a bit lacking.

PabloSzx:
closes #471

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -668,7 +668,7 @@ declare namespace mercurius {
       args: TArgs,
       context: TContext,
       info: GraphQLResolveInfo & {
-        mergeInfo: MergeInfo
+        mergeInfo?: MergeInfo
       }
     ) => boolean | Promise<boolean>
   ) => (
@@ -676,7 +676,7 @@ declare namespace mercurius {
     args: TArgs,
     context: TContext,
     info: GraphQLResolveInfo & {
-      mergeInfo: MergeInfo
+      mergeInfo?: MergeInfo
     }
   ) => AsyncGenerator<TPayload>
 }
@@ -744,7 +744,7 @@ export interface IFieldResolver<TSource, TContext = MercuriusContext, TArgs = Re
     args: TArgs,
     context: TContext,
     info: GraphQLResolveInfo & {
-      mergeInfo: MergeInfo;
+      mergeInfo?: MergeInfo;
     }
   ): any;
 }


mcollina (commented):
Could you add a test for this?

PabloSzx:
sure, I could add this

```ts
subscribe: mercurius.withFilter<{
          name: string
          breed?: string
        }>(
          (_root, _args, { pubsub }, info) => {
            info.mergeInfo
            return pubsub.subscribe('new_dog')
          },
          (payload) => {
            return payload.breed === 'retriever'
          }
        )
```

but to be honest, that doesn't do much, and I am not able to find where is the equivalent JavaScript code that's adding that "mergeInfo" in runtime, I didn't add it to the types either.

mcollina:
I'm a little puzzled but ok, let's land this.

Issue #481, opened 2021-05-08
Cached validation error because of variables error return errors on subsequent correct queries


Davide-Gheri:
This is a particular case, but happened to me a couple of times already.

Let's say we have validation rule that takes into account variables like `graphql-query-complexity`:
```js
validationRules: ({ variables, operationName }) => [
    createQueryComplexityValidator({
        maximumComplexity: 100,
        variables,
        operationName,
        estimators: [fieldExtensionsEstimator(), simpleEstimator({ defaultComplexity: 1 })],
     }),
],
```

and a query like this:
```graphql
query SomeQuery($id: String!) {
   someField(id: $id)
}
```

If a client makes a request with an incorrect `$id` variable (a number, or even `undefined`) the server will respond with an error. Since the validation rule checks variables, this error is thrown by the validation rule, not the graphql query execution.

If then the client makes another request with the same query but correct variables, the server will keep responding with the same validation error as before, even if the validation rule will pass.

This is because of the LRU cache used on:
https://github.com/mercurius-js/mercurius/blob/3d261827c7cba55cca1d79f87e1a59ae835ea114/index.js#L423-L429

does not take into account variables and the fact that the same source can be called with different variables, and that a validation rule could use those variables to validate the document

Disabling the cache obviously fixes the problem but it's not a real fix.
A quick and dirty fix could be generate a cache key concatenating the `source` and stringified `variables`:
```js
const cacheKey = source + JSON.stringify(variables);
...
lru.get(cacheKey);
...
lruErrors.get(cacheKey);
```

mcollina:
I think you have found a really interesting bug, good spot!

I think the solution is a bit more complex, because adding variables to the cache key completely defeat the purpose of caching the query (it's very close to disabling it).

I think the only solution is to move the `validationRules` outside of that caching logic, as those cannot be cached.

Would you like to send a PR?

Davide-Gheri:
@mcollina You are suggesting moving calls to `preValidationHandler` and `validate` after the caching logic, so basically caching only the document parsing (and maybe we can keep caching queryDepth check).

With that we could remove entirely `lruErrors` since it is used only there, do you agree?

I'll start working on a PR the next days

Davide-Gheri:
I don't think it will fix, the keys don't change if the query source don't change, the values however are what causes the problem and can change without affecting the query source string

mcollina:
Could this happen only with the custom rule you mentioned?

[...]

PR #485, opened 2021-05-11
Prevent query caching with validationRules function

mcollina:
Closes https://github.com/mercurius-js/mercurius/issues/481


PacoDu (approved):
lgtm

Issue #476, opened 2021-05-02
Improve Mercurius test stability


jonnydgreen:
As suggested in this conversation https://github.com/mercurius-js/mercurius/pull/470#issuecomment-828245464, I'm opening this issue to discuss improvements to the Mercurius test stability before refactoring them to be more stable.

## Observed issues

| Test                                                                                                                                                                                                                  | Environment(s)                 |
| --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------- |
| [`gateway/pollingInterval.js: Polling schemas (subscriptions should be handled)`](#polling-schemas-subscriptions-should-be-handled)                                                                                   | `[windows-latest]-[14.x.x]` |
| [`gateway/remote-services.js: All tests`](#remote-services.js)                                                                                                                                                        | All tests                   |
| [`gateway/pollingInterval.js: Polling schemas (should properly regenerate the schema when a downstream service restarts)`](#polling-schemas-should-properly-regenerate-the-schema-when-a-downstream-service-restarts) | All tests                   |

### Polling schemas (subscriptions should be handled)

This is easily seen in the CI if you rewrite the tests to repeat until failure (I have tested this on a test branch on my fork). From my initial experiments, this test fails ~33% of the time. I am yet to see this on a local machine (MacOS), however I am loading mercurius onto a Windows PC asap to replicate the exact environment as closely as possible.

The point of failure is [this line here](https://github.com/mercurius-js/mercurius/blob/master/test/gateway/pollingInterval.js#L1179). None of the assertions after this line are reached, and from the logs, the test is exited prematurely (causing a test plan error). I need to dig deeper to find the exact cause, but my initial thoughts are that since it is intermittent, it is some race condition.

### remote-services.js

The tests in this file take about ~8s to run which suggests that everything is not being cleaned up. From an initial look at the code, I think it might be to do with undici connections not being cleaned up when `buildFederation` fails. If this is correct, this is a fairly simple fix, but tricky to test!

### Polling schemas (should properly regenerate the schema when a downstream service restarts)

This test takes ~4s to run which again suggests that everything is not being cleaned up properly. Similarly to the `remote-services.js` being slow, this looks like it may be to do with the undici clients not being closed in the gateway when a downstream service is closed, however, I need to dig deeper to confirm the exact cause!

[...]

PR #491, opened 2021-05-15
Improve test stability

jonnydgreen:
Fixes: https://github.com/mercurius-js/mercurius/issues/476

The intermittent `pollingInterval.js` failures include two fixes:
 - One to prevent a race condition on Windows when writing a `connection_init` and `start` at the same time.
 - One to prevent a race condition on Windows where the mutation was called before the subscription was created.

For reducing the time taken for the `remote-services.js` test to run, I was wasn't able to work out a test case for this without a large code refactor. I've added a timeout for this test to capture the fact that the service connections are now being cleaned up correctly - let me know if you have any ideas on this! :)

I've run the tests on repeat ~60x and found no failures, so am fairly confident this catches the intermittent failures.

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -187,6 +187,9 @@ async function buildGateway (gatewayOpts, app) {
   }, [])
 
   if (serviceSDLs.length < 1) {
+    for (const service of Object.values(serviceMap)) {
+      service.close()
+    }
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm, good work!!

Issue #499, opened 2021-05-31
Value Types


steinroe:
[According to the Apollo Federation Docs](https://www.apollographql.com/docs/federation/value-types/), it is possible to share types, so-called value types, among subgraphs. This is e.g. useful when using a relay-based graph, where all connections share a `PageInfo` type. These value types don't "originate" in a particular subgraph. Instead, all of the subgraphs that define a value type share ownership.

An example
```
// Subgraph 1
extend type Query {
  posts: PostConnection!
}

type PostConnection {
  pageInfo: PageInfo!
  edges: [PostEdge!]!
}

type PageInfo {
  hasNextPage: Boolean
}

type PostEdge {
  node: Post!
}

type Post @key(fields: "id") {
  id: ID!
}
```

```
// Subgraph 2
extend type Query {
  users: UserConnection!
}

type UserConnection {
  pageInfo: PageInfo!
  edges: [UserEdge!]!
}

type PageInfo {
  hasNextPage: Boolean
}

type UserEdge {
  node: User!
}

type User @key(fields: "id") {
  id: ID!
}
```

When querying a gateway for `users`, in which the user service is listed before the post service, it throws `Missing @key directive in PageInfo type`. I debugged this issue and it seems like that the `PageInfo` type is assigned to the post service and it is not checked whether a type is a value type.

I am happy to provide a PR to "search for" value types and remove the assignment to a service.

steinroe:
Should we keep this open to implement the todo I mentioned? Otherwise feel free to close. :)

mcollina:
> Should we keep this open to implement the todo I mentioned? Otherwise feel free to close. :)

Can you please document them here?

steinroe:
Sure!

As of now, we search for duplicates based on the type names and decide whether it is a value type by filtering out all entities (determined by the key directive) and types that are neither object, input nor interface.

[According to the Apollo Federation docs on Value Types](https://www.apollographql.com/docs/federation/value-types/) however, for types with field definitions, all fields and their types must be identical (including nullability). Further, for enums and unions, all possible values must match. Otherwise, the schema composition should fail.

Hence, summarised:
- Include enums and unions in the search for value types
- Compare the value type candidates to ensure that they are the same and if not, schema composition should fail
  - if object, interface or inputs, all fields and their types must be identical 
  - if enum or union, all possible values must match

PR #502, opened 2021-05-31
feat(gateway): handle value types

steinroe:
Implements value type handling as discussed in #499

Leaves an open todo as described in find-value-types.js:
```
/**
 * TODO: it should be checked whether all types with the same name have the exact same fields,
 * incl. types, subtypes and nullability
 * If that is not the case, I would argue that throwing would be the best approach
 */
```

### DIFF
--- a/lib/gateway/find-value-types.js
+++ b/lib/gateway/find-value-types.js
@@ -1,22 +1,23 @@
+'use strict'
 const {
   isInterfaceType,
   isObjectType,
   isInputType
 } = require('graphql')
 
-function duplicates (array) {
-  const uniq = array
-    .map((item) => {
-      return {
-        count: 1,
-        item
-      }
-    })
-    .reduce((a, b) => {
-      a[b.item] = (a[b.item] || 0) + b.count
-      return a
-    }, {})
-  return Object.keys(uniq).filter((a) => uniq[a] > 1)
+function getDuplicateTypes (types) {
+  const occurrences = {}
+  const duplicates = []
+  for (const type of types) {
+    /* istanbul ignore else */
+    if (!occurrences[type.name]) {
+      occurrences[type.name] = 1
+    } else if (occurrences[type.name] === 1) {
+      duplicates.push(type)
+      occurrences[type.name] = 2
+    }
+  }
+  return duplicates
 }
 
 /**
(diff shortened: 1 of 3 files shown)


test/gateway/value-types.js
```diff
@@ -0,0 +1,410 @@
...
+  })
+
+  t.same(JSON.parse(postsRes2.body), {
+    data: {
+      hello: 'world'
+    }
+  })
+})
+
+test('Should not register value types if field ', async (t) => {})
```
mcollina, line 410: can you fill this in?

test/gateway/value-types.js
```diff
@@ -0,0 +1,410 @@
...
+    headers: {
+      'content-type': 'application/json'
+    },
+    url: '/graphql',
+    body: JSON.stringify({
+      query: usersQuery
+    })
+  })
+
+  t.same(JSON.parse(usersRes.body), usersData)
```
mcollina, line 263:
```suggestion
  t.same(usersRes.json(), usersData)
```

and everywhere else.

lib/gateway/find-value-types.js
```diff
@@ -0,0 +1,54 @@
...
+        count: 1,
+        item
+      }
+    })
+    .reduce((a, b) => {
+      a[b.item] = (a[b.item] || 0) + b.count
+      return a
+    }, {})
+  return Object.keys(uniq).filter((a) => uniq[a] > 1)
+}
```
mcollina, line 20: Why can't we just put all names in a set instead?

lib/gateway/find-value-types.js
```diff
@@ -0,0 +1,54 @@
...
+ * Find all value types according to the definition given here:
+ * https://www.apollographql.com/docs/federation/value-types/
+ *
+ * TODO: it should be checked whether all types with the same name have the exact same fields,
+ * incl. types, subtypes and nullability
+ * If that is not the case, I would argue that throwing would be the best approach
+ */
+function findValueTypes (allTypes) {
+  // Iterate all service defs and find all types by name that are used by more than one service
+  const duplicateNames = duplicates(allTypes.map(type => type.name))
```
mcollina, line 32: Delete this map

test/gateway/value-types.js
steinroe: This seems to be a leftover. Or do you miss a test?

mcollina: looks like it is, just drop it!

steinroe:
Thanks for your quick feedback!

I changed the implementation for find-value-types into the following, but the test coverage for functions got reduced to 80% for that file and I cannot figure out why. Do you have an idea?

```js
const {
  isInterfaceType,
  isObjectType,
  isInputType
} = require('graphql')

function getDuplicateTypes (types) {
  const occurrences = {}
  const duplicates = []
  for (const type of types) {
    if (occurrences[type.name] === 1) {
      duplicates.push(type)
      occurrences[type.name] = 2
    } else {
      occurrences[type.name] = 1
    }
  }
  return duplicates
}

/**
 * Find all value types according to the definition given here:
 * https://www.apollographql.com/docs/federation/value-types/
 *
 * TODO: it should be checked whether all types with the same name have the exact same fields,
 * incl. types, subtypes and nullability
 * If that is not the case, I would argue that throwing would be the best approach
 */
function findValueTypes (allTypes) {
  // Is exposed by more than one service
  return getDuplicateTypes(allTypes)
    .filter(type => {
      // Has to be object, interface or input
      const isEligibleType = (isObjectType(type) || isInterfaceType(type) || isInputType(type))
      // Is not an entity
      const isNotEntity = !type.astNode.directives.find(directive => directive.name.value === 'key')
      return isEligibleType && isNotEntity
    })
    .map(type => type.name)
}

module.exports = {
  findValueTypes
}

```

lib/gateway/find-value-types.js
```diff
@@ -0,0 +1,54 @@
+const {
```
mcollina, line 1: use strict is missing

mcollina:
I don't know, I'm sorry.

I would change `getDuplicateTypes` to:

```js
function getDuplicateTypes (types) {
  const unique = new Set()
  const duplicates = new Set()
  for (const type of types) {
    const name = type.name
    if (unique.has(name))
      duplicates.add(name)
    else
      unique.add(name)
  }
  return duplicates // no need to convert to array, we can use it with a `for...of` later
}
```

steinroe:
Thanks for your help. The coverage was missing because it never ran into `isNotEntity` in `return isEligibleType && isNotEntity`. I added a test for that and cleaned up.

lib/gateway/find-value-types.js
steinroe: Because `allTypes` is an array of `GraphQLObjectType` and not strings. Using the `GraphQLObjectType` instead of strings is required for `isObjectType` etc. to work properly.

mcollina (approved):
lgtm

PR #507, opened 2021-06-08
refactor(gateway): defineResolvers now properly handles value types

steinroe:
Hi,

I realized that the current valueTypes implementation is fundamentally flawed, for example, "top-level" value types were not handled properly. When a mutation/query/subscription returns a value type, it throws `Cannot return null for Mutation.myMutation`.

Hence, I refactored the way the `defineResolvers` function handles non-scalar types.  I added the documentation in-line. It basically uses `serviceForFieldType` and `serviceForType` to derive whether the field should be resolved from the parent, through a query / subscription / mutation or through an entity query.

#499

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -151,7 +151,7 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
              * If there is no service for the type, it is a query, mutation or subscription
              */
             let service = serviceMap[serviceForFieldType]
-            if (typeof service === 'undefined') {
+            if (!service) {
               /**
                * If the value type is the return type of a query, subscription or mutation, its service is undefined, e.g. for
                * extend type Query {
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -164,14 +164,14 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
             }
             if (type.name === 'Subscription') {
               field.subscribe = makeResolver({
-                service: service,
+                service,
                 createOperation: createQueryOperation,
                 isQuery: true,
                 isSubscription: true
               })
             } else {
               field.resolve = makeResolver({
-                service: service,
+                service,
                 createOperation: createQueryOperation,
                 transformData: response => response.json.data[fieldName],
                 isQuery: true
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -37,23 +37,41 @@ function isDefaultType (type) {
 /**
  * The gateway resolver methods are responsible to delegate certain parts of the query to a service.
  *
- * For every type in the schema - defined by services - it checks each field to decide what kind of resolver it needs.
- * This is done by getting the service name of the current type and comparing it to the return type of the field when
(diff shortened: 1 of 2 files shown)


mcollina:
Does this fix https://github.com/mercurius-js/mercurius/issues/499 completely?

mcollina (approved):
lgtm

steinroe:
The todos described refer to the recognition of value types and this PR only fixes the resolvers, so we cannot close the issue yet.

lib/gateway.js
```diff
@@ -90,58 +90,82 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
...
+              field.resolve = makeResolver({
+                service: service,
+                createOperation: createQueryOperation,
+                transformData: response => response.json.data[fieldName],
+                isQuery: true
+              })
+            }
+          } else if (serviceForFieldType !== null && serviceForFieldType !== serviceForType) {
+            /**
+             * If there is no service for the field type but a service for the type and it is not the same,
```
steinroe, line 138: this should be `If there is a service for the field type`

mcollina: can you update it?

steinroe: Yes, I also saw that the docs on function level are not required anymore. Should I remove them or rather change it to reflect the new design?

mcollina: Updated them to the new design

lib/gateway.js
```diff
@@ -90,58 +90,82 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
...
+               *    userServiceInfo: SomeReturnType!
+               * }
+               * where SomeReturnType is a value type.
+               * In these cases, we get the service from the typeFieldsToService map.
+               */
+              service = serviceMap[typeFieldsToService[`${type}-${fieldName}`]]
+            }
+            if (type.name === 'Subscription') {
+              field.subscribe = makeResolver({
+                service: service,
```
jonnydgreen, line 123:
```suggestion
                service,
```

lib/gateway.js
```diff
@@ -90,58 +90,82 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
...
+                createOperation: createQueryOperation,
+                isQuery: true,
+                isSubscription: true
+              })
+            } else {
+              field.resolve = makeResolver({
+                service: service,
```
jonnydgreen, line 130:
```suggestion
                service,
```

jonnydgreen (commented):
Looks great! I really like the comments you've added 🚀

lib/gateway.js
steinroe: I updated the docs :)

simoneb:
Just dropping a line to say I've been working right today on a similar area to make subscriptions work in gateway with scalar types https://github.com/mercurius-js/mercurius/pull/508. Not sure if the two are related but I thought I'd mention this anyway. What I found is that handling of types needs to be consistent when done through a federation gateway, or you'll end up with things which work when in a classic scenario but not when in gateway mode

Issue #511, opened 2021-06-12
Loaders cause subscriptions to error


kidqueb:
It looks like loaders are undefined for subscriptions and will return "Cannot read property 'X' of undefined". I made a repro but havent been able to really dig into it yet. 

What I've gather so far is this line here https://github.com/mercurius-js/mercurius/blob/master/index.js#L367 is where it originates and we can at least allow the payload to return what data it has passed by making sure we have a fn to call.

```
if (reply[kLoaders] && reply[kLoaders][name]) {
  return reply[kLoaders][name]({ obj, params })
}
```

For my use case I'm actually sending all the data the subscription needs to populate nested records, but this error prevents that data from making its way over.

I get some L&D time at work so I can look into it a bit more this week, but I also have a hunch theres more to it fully working than passing in the loaders to subscriptions. Any tips on the mechanisms at work before diving in further would be appreciated.

https://github.com/kidqueb/mercurius-sub-loader-repro

jonnydgreen:
Good catch! As I understand it, the problem is the fact that loaders are `undefined` on the reply object when used in subscriptions.

From a quick glance at the codebase, it seems the loaders are only created in a GraphQL request:

```js
app.decorateReply('graphql', function (source, context, variables, operationName) {
  context = Object.assign({ reply: this, app }, context)
  if (app[kFactory]) {
    this[kLoaders] = factory.create(context)
  }

  return app.graphql(source, context, variables, operationName)
})
```

The `reply` object for subscriptions is constructed here: https://github.com/mercurius-js/mercurius/blob/master/lib/subscription-connection.js#L192, so if we take a similar approach, I think that should solve the issue! The only difficulty is how we get hold of the factory instance, which we can maybe put on the fastify graphql instance object perhaps

kidqueb:
Currently the way I understand the way loaders work is by creating a cache for each request, leveraging that cache to minimize db access, and then after the reply is sent it gets garbage collected.

Does that sound about right? Is it safe to assume this cache would be problematic on a connection? 

If so, maybe the simple solution is to abstract away the creation of loaders and create the subscription loaders specifically with `{ cache: false }` forced into the options.

mcollina:
> Currently the way I understand the way loaders work is by creating a cache for each request, leveraging that cache to minimize db access, and then after the reply is sent it gets garbage collected.
> 
> 
> 
> Does that sound about right? Is it safe to assume this cache would be problematic on a connection? 
> 
> 
> 
[...]

PR #512, opened 2021-06-12
Check for loader existence before calling it

kidqueb:
https://github.com/mercurius-js/mercurius/issues/511

Made an interim PR to patch the error preventing the payload from sending over what it can incase this is at all helpful. Will follow up on that issue when I can get back to this in the upcoming week.


mcollina (changes requested):

kidqueb (commented):
Modified to account for existing data instead of returning null and added a test 👍

Edit: Oops, commented as a review ha

mcollina (approved):
lgtm

jonnydgreen (approved):

PR #528, opened 2021-07-06
chore: upgrade undici to v4 and deps

simone-sanfratello:
notes:
- **drop node `v10` support** because `undici` does not support it, that should lead to a major release
- can't upgrade `p-map` to `v5` because it is an ESM module now


mcollina:
Coverage decreased a bit.

simone-sanfratello:
probably was related to the node v10 failing test?

mcollina (approved):
lgtm

PR #538, opened 2021-07-20
Add GraphiQL request headers tab

puchm:
I was missing the request headers tab so I enabled it again - I think this is in favour of everyone.

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -32,7 +32,11 @@ function render () {
   }
 
   ReactDOM.render(
-    React.createElement(GraphiQL, { fetcher }),
+    React.createElement(GraphiQL, {
+      fetcher,
+      headerEditorEnabled: true,
+      shouldPersistHeaders: true
+    }),
     document.getElementById('main')
   )
 }


mcollina (approved):
lgtm, this is great!

puchm:
Just realised this doesn't work - sorry I expected that to be all but since the fetcher function is being overwritten as well we have to modify that to send the headers as well.
Current behaviour is that it shows the headers tab but doesnt send them.

puchm:
Ok so...
Currently, the fetcher is defined as follows:
```
async function fetcher (params) {
    const res = await fetch(window.GRAPHQL_ENDPOINT, {
      method: 'post',
      headers: {
        Accept: 'application/json',
        'Content-Type': 'application/json'
      },
      body: JSON.stringify(params),
      credentials: 'include'
    })
    return res.json()
  }
```

What I don't get about this is the `credentials` option. Does it have any use? If not, we could just use the easiest way of creating the fetcher which is as follows:
```
const fetcher = GraphiQL.createFetcher({ url: window.GRAPHQL_ENDPOINT})
```

The headers from the headers tab work after changing it to this. If the `credentials` option has an actual use we need to insert the headers from the headers tab into the current fetcher - I could however not figure out how to get the headers from the input box.

What do you think?

mcollina:
You can see an example implementation at https://github.com/graphql/graphiql/blob/65aa64811ccd40c5e8ba99edec825a91115a3c8b/packages/graphiql-toolkit/src/create-fetcher/lib.ts#L51-L68.

puchm:
Yes, but does the `credentials` option have any use in the current code? The sample implementation you sent does not contain that and the solution I proposed also doesn't. I am wondering if we need it or not.

mcollina:
I'm not sure what `credentials` you are referring to.

puchm:
It is in the current implementation of `fetcher` we are using. It's an option for JavaScript's `fetch`.

I pasted the current implementation in a previous comment, this one: https://github.com/mercurius-js/mercurius/pull/538#issuecomment-885162237

mcollina:
Could you send a PR to fix this problem? Using the `createFetcher` should be ok.

Issue #343, opened 2020-11-04
Federation gateway authentication and authorization


adamovittorio:
I would like to handle authentication and authorization per GraphQL operation at the Gateway level.

This is quite a complex feature, and I don't have a clear solution in mind; I would like to start the discussion for a possible implementation method.

## General considerations

- Pluggable RBAC method
- The services are responsible for defining the RBAC at schema level to inform the Gateway

```graphql
# Directives
directive @auth(
  requires: Role = USER,
) on OBJECT | FIELD_DEFINITION

enum Role {
  ADMIN
  USER
  UNKNOWN
}

# Service 1
extend Query {
  me: User @auth(role: USER)
  userById(id: ID!): User @auth(role: ADMIN)
}

type User @key(fields: "id") {
  id: ID!
  name: String!
}

# Service 2
type Posts @key(fields: "id") {
  id: ID!
  title: String!
  secretValue: String! @auth(role: ADMIN)
}

extend type User @key(fields: "id") {
  id: ID! @external
  posts: [Post]
}
```

## Prerequisites

- Implement authentication and authorization directives
- Decorate service schemas with the authentication directives

## Flow

- The user sends a GraphQL operation with credentials
- The Gateway verifies the user credentials, parses the user RBAC, and add them with the user credentials to the context
- The Gateway parses the operation document and compare it with the schema definitions, the result of this step is to have an RBAC policy to compare with the user's one
- The Gateway compares the policies and, in case of successful authorization, delegates the GraphQL execution to the underlying service

mcollina:
I've been doing some research on the topic. Most of this revolves around https://www.openpolicyagent.org/ and https://casbin.org/. 

If'd like to work on this, I'll be happy to create a new repo in the org... I think this one has way too many features already.

I _think_ we might need to define how the auth token is modelled - or maybe you are already assuming that we'd use JWT. In the case, we developed https://github.com/nearform/fast-jwt for a very similar purpose - it's a similar technique to the one used in the AWS API Gateway.

adamovittorio:
I like the idea of having different auth strategies that the user can choose. 

We could start with the most common-auth method to verify the user token and to decorate the request context with the credentials. Export the GraphQL directives to use in the schema definition, and as you suggested using something like open policy or casbin to define and check the RBAC.

All of that could live in the new auth repo, but regarding the user RBAC's comparison with the requested operation, *I still haven't properly understood how/where to do it!?*

[...]

PR #559, opened 2021-08-20
Fix duplicate custom directives in the gateway schema

jonnydgreen:
This fixes the issue report here: https://github.com/mercurius-js/mercurius/issues/343#issuecomment-900593479

I found another issue with the GraphQL server not closing properly if the federated schema throws for any reason. I'll resolve this in a follow up PR as the change is out of scope for this PR :)


mcollina (approved):
lgtm

PR #564, opened 2021-09-06
move graphql into peerDependencies

capaj:
When user installs graphql into their own package.json alongside `mercurius`, they get:
```
Ensure that there is only one instance of "graphql" in the node_modules
directory. If different versions of "graphql" are the dependencies of other
relied on modules, use "resolutions" to ensure only one version is installed.

https://yarnpkg.com/en/docs/selective-version-resolutions

Duplicate "graphql" modules cannot be used at the same time since different
versions may have different capabilities and behavior. The data from one
version used in the function from another could produce confusing and
spurious results.
```

This is a common problem with graphql package. It was previously discussed here for example: https://github.com/MichalLytek/type-graphql/issues/144 

Beware-this is a breaking change. Consumers of mercurius will need to go in and add graphql as direct dependency when this gets released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,9 @@ Features:
 ## Install
 
 ```bash
-npm i fastify mercurius
+npm i fastify mercurius graphql
+# or
+yarn add fastify mercurius graphql
 ```
 
 The previous name of this module was [fastify-gql](http://npm.im/fastify-gql) (< 6.0.0).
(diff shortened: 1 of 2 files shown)


PabloSzx (approved):

mcollina:
`peerDependencies` are really complex to manage across multiple package managers. I'm sorry it came to this.

mcollina (approved):
lgtm

jonnydgreen (approved):
lgtm

PR #565, opened 2021-09-07
Added graphql both as a peerDependency and as a dependency

mcollina:


mcollina:
cc @simone-sanfratello check if this works for you

PR #579, opened 2021-09-22
docs(typescript): fix mercurius-codegen example

FdezRomero:
The current example in the docs passes the `mercurius` instance to `mercuriusCodegen` instead of the Fastify instance. This PR fixes the example.


mcollina (approved):
lgtm

PR #589, opened 2021-09-28
Add mercurius-validation plugin docs

jonnydgreen:


jonnydgreen:
I can add an entry for `mercurius-cache` as well while I'm at it if you would like?

mcollina (approved):
lgtm

mcollina:
> I can add an entry for `mercurius-cache` as well while I'm at it if you would like?

thanks!

PR #591, opened 2021-10-01
fix(typescript): Fixed error handling types.

ShogunPanda:
Hello!
This PR fixes wrong TypeScript specification for the `errorHandler` option.
By digging in the code I noticed that handler is used in fastify's `setErrorHandler` which does not allow a return type.
The response must be sent to the client using `reply.send`. The only difference with fastify is that the error type can optionally (and most of the times will) include a `errors` array, so I've created a specific exported type.

I've also fixed the example for the federated use-case.

### DIFF
--- a/docs/federation.md
+++ b/docs/federation.md
@@ -374,14 +374,14 @@ server.register(mercurius, {
         name: 'company',
         url: 'http://localhost:3001/graphql'
       }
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #593, opened 2021-10-04
feat: add support to graqphql_transport_ws

### LINKED ISSUE #574: Subscription Protocols: graphql-transport-ws
Hi,

we are currently moving away from Apollo client and using urql now (highly recommend!). Urqls recommended subscription client is graphql-ws, [which is using](https://github.com/enisdenjo/graphql-ws/discussions/237#discussioncomment-1337652) the `graphql-transport-ws` protocol and not the `graphql-ws` protocol that is used by the [deprecated subscriptions-transport-ws](https://github.com/apollographql/subscriptions-transport-ws/blob/231c1b0d25b20090033140b83c8a39871d4ed88c/src/protocol.ts#L1). (Took me some time to get over the protocol and library names...)

It seems like mercurius is supporting the  `graphql-ws` protocol / `subscriptions-transport-ws` library only 
https://github.com/mercurius-js/mercurius/blob/037cc1f35c3e4901d3dc219a9ef88b43ac017b0e/lib/subscription.js#L12-L17

Are there plans to move to the new protocol? Or maybe enable both? I am too inexperienced with subscription protocols to tackle this issue on my own, but with some help I am happy to create a PR.

brainrepo:
Added the support to graqphql_transport_ws.
closes #574

### DIFF
--- a/examples/gateway-subscription.js
+++ b/examples/gateway-subscription.js
@@ -267,7 +267,10 @@ async function start () {
       }, {
         name: 'comment',
         url: 'http://localhost:4003/graphql',
-        wsUrl: 'ws://localhost:4003/graphql'
+        wsUrl: 'ws://localhost:4003/graphql',
+        wsConnectionParams: {
+          protocols: ['graphql-transport-ws'] // optional, if not set, will use the default protocol (graphql-ws)
+        }
       }]
     }
   })
--- a/lib/subscription-client.js
+++ b/lib/subscription-client.js
@@ -7,7 +7,7 @@ const {
   GRAPHQL_WS,
   getProtocolByName
 } = require('./subscription-protocol')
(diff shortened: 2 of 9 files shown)


brainrepo:
This version still uses graphql_ws for the gateway feature.
Should we update it to graqphql_transport_ws, or keep it unchanged?

mcollina:
> Should we update it to graqphql_transport_ws, or keep it unchanged?

Ideally we should have a config flag to pick which one.

mcollina (approved):
lgtm

mcollina:
after this land, could you prepare a PR to "flip" the default? I'll land it in the next major.

jonnydgreen (approved):
LGTM - awesome!

puchm:
If we want to support both, we need GraphiQL to adapt to that config setting too I think.

mcollina:
@puchm could you update that too?

PR #594, opened 2021-10-06
feat: switch default subscription protocol to graphql-transport-ws

brainrepo:
according with #593 
here a PR to switch the default subscription protocol to graphql-transport-ws


mcollina (approved):
lgtm

PR #595, opened 2021-10-06
feat: add status code support to ErrorWithProps

### LINKED ISSUE #592: Add a statusCode optional parameter to ErrorWithProps
Instead of

```js
const err = new mercurius.ErrorWithProps('UNAUTHORIZED');
err.statusCode = 200
throw err
```

We should be doing

```js
throw new mercurius.ErrorWithProps('UNAUTHORIZED', 200);
```

Note that we should also add some documents on how errorFormatter works as there has been a few issues about it. There might be something more to change under the hood - we might want to return a `200` by default in case of errors for example (sigh).

_Originally posted by @mcollina in https://github.com/mercurius-js/auth/issues/32#issuecomment-932356247_

brainrepo:
closes #592

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -592,11 +592,12 @@ declare namespace mercurius {
    * Extended errors for adding additional information in error responses
    */
   class ErrorWithProps extends Error {
-    constructor(message: string, extensions?: object);
+    constructor(message: string, extensions?: object, statusCode?: number);
     /**
      * Custom additional properties of this error
      */
     extensions?: object;
+    statusCode?: number;
   }
 
   /**
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -7,7 +7,7 @@ class ErrorWithProps extends Error {
   constructor (message, extensions, statusCode) {
     super(message)
     this.extensions = extensions
-    this.statusCode = statusCode || 200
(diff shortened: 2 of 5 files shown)


index.d.ts
```diff
@@ -592,11 +592,12 @@ declare namespace mercurius {
    * Extended errors for adding additional information in error responses
    */
   class ErrorWithProps extends Error {
-    constructor(message: string, extensions?: object);
+    constructor(message: string, extensions?: object, statusCode?: number);
```
jonnydgreen, line 595: could we have a type test for this?

lib/errors.js
```diff
@@ -4,9 +4,10 @@ const { formatError, GraphQLError } = require('graphql')
 const createError = require('fastify-error')
 
 class ErrorWithProps extends Error {
-  constructor (message, extensions) {
+  constructor (message, extensions, statusCode) {
```
jonnydgreen, line 7: could we update to the documentation for this?

jonnydgreen (commented):
Functionally looks good - just a couple of extra things!

mcollina (approved):
lgtm

mcollina:
@jonnydgreen you good with this?

PR #609, opened 2021-10-19
match graphiql resource paths

wolfram-fm:
As defined in `routes.js`:

```js
app.get('/graphiql/main.js', (req, reply) => {
  // ...
})

app.get('/graphiql/config.js', (req, reply) => {
  // ...
})
```

and so when we reference these resources in the html, it should match the paths so they start from root, not the current path.

### DIFF
--- a/static/graphiql.html
+++ b/static/graphiql.html
@@ -17,7 +17,7 @@
 </head>
 <body>
   <div id="main"></div>
-  <script src="./graphiql/config.js"></script>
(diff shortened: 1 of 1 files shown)


mcollina:
I tend to prefer relative resources because if we expose graphiql behind a prefix, everything would still work as we expect to. By using root-path resources, we would lose this.

What benefits would this bring? Maybe should we make this prefix-aware and/or configurable?

PR #618, opened 2021-10-30
Documentation improvements for type-graphql example

giacomorebonato:
- adding 2 missing imports. It's more important for `buildSchema` since it's exported by many libraries a function with same name, but with different implementation
- the example doesn't run out of the box because class Recipe has a `specification` field. I am proposing to omit the value in the returned type


mcollina (approved):
lgtm, thanks!

PR #630, opened 2021-11-08
Add info object to loaders

### LINKED ISSUE #627: Pass the resolvers info parameter to loaders
The documentation states that

> Each loader function has the signature loader(queries, context). queries is an array of objects defined as { obj, params } where obj is the current object and params are the GraphQL params (those are the first two parameters of a normal resolver).

This is implemented in https://github.com/mercurius-js/mercurius/blob/b671de31178395ea53f2ea39c19cc3f8d80be8eb/index.js#L371-L376

I'm suggesting to also pass the "info" param (it's the fourth one) from a normal resolver.

sameer-coder:
This PR adds support for the `info` object being passed to the loaders.

Each loader function has the signature loader(queries, context).  `queries` is an array of objects and is now defined as `{ obj, params, info }`

Closes #627

### DIFF
--- a/docs/loaders.md
+++ b/docs/loaders.md
@@ -9,10 +9,11 @@ also cache the results, so that other parts of the GraphQL do not have
 to fetch the same data.
 
 Each loader function has the signature `loader(queries, context)`.
-`queries` is an array of objects defined as `{ obj, params }` where
-`obj` is the current object and `params` are the GraphQL params (those
-are the first two parameters of a normal resolver). The `context` is the
-GraphQL context, and it includes a `reply` object.
+`queries` is an array of objects defined as `{ obj, params, info }` where
+`obj` is the current object, `params` are the GraphQL params (those
+are the first two parameters of a normal resolver) and `info` contains
+additional information about the query and execution. The `context` is 
+the GraphQL context, and it includes a `reply` object.
 
 Example:
 
--- a/docs/loaders.md
+++ b/docs/loaders.md
@@ -12,8 +12,9 @@ Each loader function has the signature `loader(queries, context)`.
 `queries` is an array of objects defined as `{ obj, params, info }` where
 `obj` is the current object, `params` are the GraphQL params (those
 are the first two parameters of a normal resolver) and `info` contains
-additional information about the query and execution. The `context` is 
-the GraphQL context, and it includes a `reply` object.
+additional information about the query and execution. `info` object is
+only available in the loader if the cache is set to `false`. The `context`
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

mcollina:
Ci is failing

sameer-coder:
@mcollina yes, it was failing because of some loader-related tests. Should Pass now

test/loaders.js
```diff
@@ -86,6 +87,11 @@ test('loaders create batching resolvers', async (t) => {
             name: 'Buddy'
           },
           params: {}
+        }, {
+          obj: {
+            name: 'Max'
+          },
+          params: {}
```
mcollina, line 94: This should not happen and it was what I feared about this implementation. By making `info` part of of the object, we are passing it through https://github.com/mcollina/single-user-cache/blob/master/index.js#L66 and the fetching of this entity is now duplicated.

As you can see, it has now the same output when the cache is disabled. __I think we should add the `info` object only if the cache is disabled__.

mcollina (changes requested):

PacoDu:
There was a similar PR about passing the info field to loaders: https://github.com/mercurius-js/mercurius/pull/334

mcollina:
This is the relevant comment: https://github.com/mercurius-js/mercurius/pull/334#issuecomment-715139247

docs/loaders.md
```diff
@@ -9,10 +9,12 @@ also cache the results, so that other parts of the GraphQL do not have
...
-`queries` is an array of objects defined as `{ obj, params }` where
-`obj` is the current object and `params` are the GraphQL params (those
-are the first two parameters of a normal resolver). The `context` is the
-GraphQL context, and it includes a `reply` object.
+`queries` is an array of objects defined as `{ obj, params, info }` where
+`obj` is the current object, `params` are the GraphQL params (those
+are the first two parameters of a normal resolver) and `info` contains
+additional information about the query and execution. `info` object is
+only available in the loader if the cache is set to `false`. The `context`
+is the GraphQL context, and it includes a `reply` object.
```
mcollina, line 17: the docs must be updated

sameer-coder: I did update it to add a line about when the `info` object will be available
> `info` contains additional information about the query and execution. `info` object is only available in the loader if the cache is set to `false`

Should I remove the info object and change it back to `{ obj, params, info }` in the docs?

mcollina: Ah I missed it, it's ok now.

mcollina (approved):
lgtm

PR #633, opened 2021-11-09
[bugfix]: "Cannot read property 'create' of undefined" with encapsulated app where parent object has loaders defined

tobocop2:
In an effort to create a new graphql endpoint, I was encountering the following error when registering the following plugin code:
NOTE: the parent app object has loaders defined in a separate module. 

```javascript

'use strict'

const fp = require('fastify-plugin')
const GQL = require('mercurius')

const schema = `
  type Query {
    msg: String
  }
`

const resolvers = {
  Query: {
    msg
  }
}

async function msg(_, args, context) {
  return "hello world!"
}


async function plugin(server) {


  server.register(async (app) => {
    app.register(GQL, {
      schema,
      defineMutation: true,
      resolvers,
      path: '/auth'
    })
  })
}

module.exports = fp(plugin, {
  name: 'auth',
  dependencies: ['db', 'redis']
})
```

Upon issuing the following request:

```bash
curl 'localhost:5000/auth' -X POST -H 'Content-Type: application/json' --data-raw '{"query":"query { msg }","variables":null}' | jq
```

I am receiving the following response:

```json
{
  "data": null,
  "errors": [
    {
      "message": "Cannot read property 'create' of undefined"
    }
  ]
}
```

With this bug fix, this problem is now resolved. the factory object is otherwise not in scope when registering the new route this way.

I absolutely need multiple graphql routes for the problem I am trying to solve and this fix has resolved my issue.

### DIFF
--- a/index.js
+++ b/index.js
@@ -258,7 +258,7 @@ const plugin = fp(async function (app, opts) {
 
     context = Object.assign(context, { reply: this, app })
     if (app[kFactory]) {
-      this[kLoaders] = factory.create(context)
+      this[kLoaders] = app[kFactory].create(context)
     }
 
     return app.graphql(source, context, variables, operationName)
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):
Good spot! Can you please add a unit test?

tobocop2:
@mcollina tests added, see here: 
https://github.com/mercurius-js/mercurius/pull/633/commits/8292ef4095da7bd3409d8f40bd78d569ff1f6e8b

I also updated the description of the bug fix as it occurs only when defineLoaders is called.

mcollina (approved):
lgtm

PR #637, opened 2021-11-09
Fix: Graphiql not working without sw

### LINKED ISSUE #625: GraphiQL IDE not loading without service worker support
When using GraphiQL IDE in a browser that doesn't support service workers the editor doesn't load and users can only see a blank screen.

The code rendering GraphiQL IDE has some level of service worker support detection in https://github.com/mercurius-js/mercurius/blob/master/static/main.js#L42-L62
When service worker is not supported, the code tries to render the app without first downloading dependencies, resulting in a javascript error.

It's worth noting that this is not limited to outdated browsers: service workers are only supported when using https.

sameer-coder:
This PR fixes an issue with graphiql when we try to load it without service workers.

Closes #625

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -39,25 +39,28 @@ function render () {
   )
 }
 
+function importDependencies () {
+  const link = document.createElement('link')
+  link.href = 'https://unpkg.com/graphiql@1.4.2/graphiql.css'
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

Issue #629, opened 2021-11-08
getProtocolByName has the wrong mapping for GQL_STOP?


coder-se:
When using `graphql-transport-ws` for subscriptions, `GQL_STOP` is mapped to `stop` but according to the protocol it seems like it should be `complete` both for Server > Client and Client > Server.

https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#complete

https://github.com/mercurius-js/mercurius/blob/813e940dc2941efa3d85551bc45224e73a19d0ea/lib/subscription-protocol.js#L18-L19

mcollina:
That document you are referring to is the reference for the `graphql-ws` protocol, so it's ok that things are different.

It's hard to debate protocol compliance over just an issue. Have you got an example to reproduce your problem?

coder-se:
I will try to put together a standalone example to reproduce the problem. What we're seeing is that our client (from the graphql-ws package) is sending `complete` message and our mercurius backend is responding with `Invalid payload type`. 

Regarding the document, maybe it's me being confused by the name change of the package, https://www.npmjs.com/package/graphql-transport-ws

But at the top of the document it says, 
> The WebSocket sub-protocol for this specification is: `graphql-transport-ws`.

And at the bottom it specifies following for streaming operations;
```
6. - _Client_ stops the subscription by dispatching a `Complete` message
    - _Server_ completes the source stream
      _or_
   - _Server_ dispatches the `Complete` message indicating that the source stream has completed
   - _Client_ completes the stream observer
```

coder-se:
Here's a standalone example of the issue. If you inspect the websocket messages in your browser then there should be a `Invalid payload type` error message visible.

https://gist.github.com/coder-se/f1d445b50f0efe26b303d373af792bde


![image](https://user-images.githubusercontent.com/5502726/140895569-2cbfce30-5377-4cfd-8377-56ea607f88c4.png)

PR #638, opened 2021-11-09
Fix incorrect message type for GQL_STOP when using graphql-transport-ws

coder-se:
Closes https://github.com/mercurius-js/mercurius/issues/629

Let me know if the unit-tests should be structured differently. Or if you want me to extend the PR and add tests for all the message types of `graphql-transport-ws`.


mcollina (approved):
lgtm

PR #641, opened 2021-11-09
feat: add modifiedSchema on preExecution hooks

Eomm:
Opening this PR to get some feedback.

I'm working on this issue: https://github.com/mercurius-js/auth/issues/43
The target of that issue is to modify the output schema for inspection queries like so

```js
app.graphql.addHook('preExecution', async function filterHook (schema, document, context) {

  // removes those objects that the user has not access to
  const filteredSchema = auth.filterDirectives(schema, authSchema)

  return {
    schema: filteredSchema
  }
})
```

The inspection query is a standard query that is being executed on the `app.graphql.schema` object.

To let the user (and in this case the mercurius/auth plugin) change the schema where the query is executed and affect the output results, the `preExecution` hook may manage a schema replacing.

Note that I did not use `app.graphql.transformSchema` because it would replace the initial/original schema: this is unwanted: the target here is to replace the schema only for a single request execution.

Doing so, every request may change the schema accordingly to its data (such as client's role&permissions).

This PR lacks documentation, but I would like to know if you are OK with this solution before add it.
Feel free to suggest some additional tests too

Thanks

### DIFF
--- a/index.js
+++ b/index.js
@@ -515,7 +515,6 @@ const plugin = fp(async function (app, opts) {
     }
 
     const shouldCompileJit = cached && cached.count++ === minJit
-
     // Validate variables
     if (variables !== undefined && !shouldCompileJit) {
       const executionContext = buildExecutionContext(fastifyGraphQl.schema, document, root, context, variables, operationName)
@@ -534,13 +533,17 @@ const plugin = fp(async function (app, opts) {
     }
 
     // minJit is 0 by default
-    if (shouldCompileJit && !modifiedSchema) {
-      cached.jit = compileQuery(fastifyGraphQl.schema, modifiedDocument || document, operationName)
+    if (shouldCompileJit) {
+      if (!modifiedSchema) {
+        cached.jit = compileQuery(fastifyGraphQl.schema, modifiedDocument || document, operationName)
+      } else {
+        // the must decrease the counter to ignore the query
+        cached && cached.count--
+      }
     }
 
     if (cached && cached.jit !== null) {
       const execution = await cached.jit.query(root, context, variables || {})
-
       return maybeFormatErrors(execution, context)
     }
 
--- a/index.js
+++ b/index.js
@@ -164,9 +164,9 @@ const plugin = fp(async function (app, opts) {
       }),
       mutation: opts.defineMutation
         ? new GraphQLObjectType({
-          name: 'Mutation',
-          fields: {}
-        })
+            name: 'Mutation',
+            fields: {}
+          })
         : undefined
     })
   }
--- a/index.js
+++ b/index.js
@@ -534,11 +534,8 @@ const plugin = fp(async function (app, opts) {
     }
 
     // minJit is 0 by default
-    if (shouldCompileJit) {
-      cached.jit = compileQuery(
-        modifiedSchema || fastifyGraphQl.schema,
-        modifiedDocument || document, operationName
-      )
+    if (shouldCompileJit && !modifiedSchema) {
+      cached.jit = compileQuery(fastifyGraphQl.schema, modifiedDocument || document, operationName)
     }
 
     if (cached && cached.jit !== null) {
--- a/index.js
+++ b/index.js
@@ -527,14 +527,18 @@ const plugin = fp(async function (app, opts) {
     }
 
     // Trigger preExecution hook
+    let modifiedSchema
     let modifiedDocument
     if (context.preExecution !== null) {
-      ({ modifiedDocument } = await preExecutionHandler({ schema: fastifyGraphQl.schema, document, context }))
+      ({ modifiedSchema, modifiedDocument } = await preExecutionHandler({ schema: fastifyGraphQl.schema, document, context }))
     }
 
     // minJit is 0 by default
     if (shouldCompileJit) {
-      cached.jit = compileQuery(fastifyGraphQl.schema, modifiedDocument || document, operationName)
+      cached.jit = compileQuery(
+        modifiedSchema || fastifyGraphQl.schema,
+        modifiedDocument || document, operationName
+      )
     }
 
     if (cached && cached.jit !== null) {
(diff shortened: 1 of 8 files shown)


mcollina:
> To let the user (and in this case the mercurius/auth plugin) change the schema where the query is executed and affect the output results, the preExecution hook may manage a schema replacing.

This seems a really bad idea. What about concurrency issues?

Eomm:
> What about concurrency issues?

I don't expect those issues since the "temporary-schema" is used on the request's scope and it must not affect the global schema that remains the default.

I can arrange a test for it as well

Eomm:
> This seems a really bad idea.

I did not find a way to implement the desired logic using the actual API, such as the `onResolution` hook, mainly because there is not a reference to the (auth) resolver to execute to decide if a Graph FIELD or OBJECT should be returned or not.

I agree that this could be a dangerous feature to use with caution - the risk is a broken API.
Could we protect the user from misusing this feature?

mcollina (commented):
Can you add docs?

mcollina (approved):
lgtm

index.js
```diff
@@ -527,14 +527,18 @@ const plugin = fp(async function (app, opts) {
...
+      ({ modifiedSchema, modifiedDocument } = await preExecutionHandler({ schema: fastifyGraphQl.schema, document, context }))
     }
 
     // minJit is 0 by default
     if (shouldCompileJit) {
-      cached.jit = compileQuery(fastifyGraphQl.schema, modifiedDocument || document, operationName)
+      cached.jit = compileQuery(
+        modifiedSchema || fastifyGraphQl.schema,
+        modifiedDocument || document, operationName
+      )
```
mcollina, line 541: I don't think we should cache or compile queries where the schema has been dynamically modified.

There is definitely a race here where a schema is modified and then cached while a subsequent query would change it in a different way (maybe on user permission).

mcollina (changes requested):

index.js
```diff
@@ -527,13 +527,14 @@ const plugin = fp(async function (app, opts) {
...
+    let modifiedSchema
     let modifiedDocument
     if (context.preExecution !== null) {
-      ({ modifiedDocument } = await preExecutionHandler({ schema: fastifyGraphQl.schema, document, context }))
+      ({ modifiedSchema, modifiedDocument } = await preExecutionHandler({ schema: fastifyGraphQl.schema, document, context }))
     }
 
     // minJit is 0 by default
-    if (shouldCompileJit) {
+    if (shouldCompileJit && !modifiedSchema) {
```
mcollina, line 537: This should be documented. I think the same problem might happen for modifiedDocument too - we might end up fixing a bug.

lib/handlers.js
```diff
@@ -29,15 +29,19 @@ async function preValidationHandler (request) {
...
   if (errors.length > 0) {
     addErrorsToContext(request.context, errors)
   }
-  if (typeof modifiedDocument !== 'undefined') {
-    return { modifiedDocument, modifiedQuery: print(request.document) }
+  if (typeof modifiedDocument !== 'undefined' || typeof modifiedSchema !== 'undefined') {
+    return {
+      modifiedSchema,
+      modifiedDocument,
+      modifiedQuery: print(request.document)
```
mcollina, line 43: are we using modifiedQuery anywhere?

mcollina (commented):
Please add a few test with jit enabled.

lib/handlers.js
Eomm: It is used on the gateway implementation

index.js
Eomm: Fixed

> I think the same problem might happen for modifiedDocument

I think the `modifiedDocument` is worth being jit'ed it is a consistent update.
For example, a developer is renaming a field from `old` to `new`

index.js
```diff
@@ -534,13 +533,17 @@ const plugin = fp(async function (app, opts) {
     }
 
     // minJit is 0 by default
-    if (shouldCompileJit && !modifiedSchema) {
-      cached.jit = compileQuery(fastifyGraphQl.schema, modifiedDocument || document, operationName)
+    if (shouldCompileJit) {
+      if (!modifiedSchema) {
```
mcollina, line 537:
```suggestion
      if (!modifiedSchema || !modifiedDocument) {
```

Eomm: If we change the `schema` but not the `document` the jit is executed against the "origin" schema.
Is it useful creating the jit when the document will be executed using the `modifiedSchema`?

mcollina: I don't understand.

> I think the modifiedDocument is worth being jit'ed it is a consistent update.
For example, a developer is renaming a field from old to new

What makes me itchy in jiting this is that the hook takes the `context` as an argument, therefore those changes could be based on user credentials. Compiling those in is a potential security hazard.

What we really need is another field to defer this to the user. Having a `cacheable` or `jit` property would put the developer in charge of what is happening.

Eomm: I get the point, I will try to arrange a security test for this case.

The current PR's code blocks the jit execution when the schema is modified by the hook.
The suggested code enables the jit when the user changes the `modifiedSchema` property that seems unwanted.

> What we really need is another field to defer this to the user. Having a cacheable or jit property would put the developer in charge of what is happening.

Clear, I would add it in a follow-up PR because there are a few cases that must be discussed or would you like to upgrade this PR instead?

The doubts concern:

- should we provide the cache counter to the hook function using the context?
- does the `cacheable` property have priority over the `jit` option?

mcollina (approved):
lgtm

index.js
mcollina: > should we provide the cache counter to the hook function using the context?

Unsure about this

> does the cacheable property have priority over the jit option?

Yes

lib/handlers.js
mcollina: Maybe make it a getter?

jonnydgreen (commented):
Looks good! As well as the getter added on the `modifiedQuery` could we also update the typings and the typing tests? https://github.com/mercurius-js/mercurius/blob/master/index.d.ts#L788

lib/handlers.js
Eomm: It was unused.
Removed

The same name is used on the `preGatewayExecutionHandler`

jonnydgreen (approved):
LGTM

Issue #643, opened 2021-11-10
Gateway mode - Validation of federationSchema fails - should ignore multiple _service queries 


MoritzGruber:
Hi together, 
first i have to thank you for all the work you put into this amazing project! 

I currently got the problem that i have running 
3 graphql services that got a _service Query exposed (as the federation spec)

but then at runtime i can not start my gateway because i get this error twice:
`GraphQLError [Object]: Field "Query._service" can only be defined once.`

which has its root in

this: 
`lib/federation.js`
```
  // instead of relying on extendSchema internal validations
  // we run validations in our code so that we can use slightly different rules
  // as extendSchema internal rules are meant for regular usage
  // and federated schemas have different constraints
  const errors = validateSDL(extensionsDocument, federationSchema, compositionRules)
  if (errors.length === 1) {
    throw errors[0]
  } else if (errors.length > 1) {
    const err = new MER_ERR_GQL_INVALID_SCHEMA()
    err.errors = errors
    throw err
  }
  
  ```

My current quick-fix is to comment out the validation but I see why the validation makes sense. I would not remove this validation. In fact it would be much better to add some ignore for the _service query if we collect the service SDL in gateway mode from the services. Do you think that makes sense?

**Comment by mcollina** [comment]
Sure thing! 
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

**Comment by MoritzGruber** [comment]
Sure. Amazing response time, thanks!  😮  

While digging into it. I am releasing the problem can also be solved by the **gateway** **or** the **service**. Apollo solving this in the service. 

In our use case we use gqlgen(golang) https://github.com/99designs/gqlgen
Which does not follow with what apollo states. They include the _service in the Query by default in the SDL. 

 I was reading trough the apollo docs again:
```
The _service resolver should return the _Service type which has a single field called sdl. This SDL (schema definition language) is a printed version of the service's schema including the annotations of federation directives. This SDL does not include the additions of the federation spec above.
```
https://www.apollographql.com/docs/federation/federation-spec/#fetch-service-capabilities

On the one hand, it is great to be compatible with the current gqlgen and other gql servers that have the same problem with federation. On the other hand, if gqlgen updated its default behaviour to match what Apollo suggests in the documentation, we wouldn't have the problem. 

Personally, I still think it's a "nice to have" for mercurius - even if it goes beyond the Apollo specification. As a maintainer, what do you say to this @mcollina? 

[...]

PR #645, opened 2021-11-10
compatibility: schema validation in gateway mode 

MoritzGruber:
Added a filter on the query extensions in the schema create. 
This filter removes all the _service fields in the extend queries. 
This will help to be compatible with multiple services in gateway mode that do not remove the _service field in the SDL while still supporting federation.

see:
https://github.com/mercurius-js/mercurius/issues/643


mcollina (approved):
lgtm

Issue #616, opened 2021-10-29
`errorHandler` question


wiktor-obrebski:
Currently `errorHandler` [plugin option](https://mercurius.dev/#/docs/api/options?id=plugin-options) function sometimes is triggered, but sometimes is ignored. 

In `index.js` file `errorFormatter` is used to format some errors, without calling `errorHandler`.
https://github.com/mercurius-js/mercurius/blob/6c523d96f57eb08a33a3a2b7b2e1d48c4e12e869/index.js#L558-L563

But in routes, `errorHandler` handle the error and possibly call `errorFormatter`.
https://github.com/mercurius-js/mercurius/blob/6c523d96f57eb08a33a3a2b7b2e1d48c4e12e869/lib/routes.js#L132-L140

It gives strange situation when some errors are handled in custom way I want them too, but some not. I can not use `errorFormatter` for everything, because it do not get `request` instance, so I have no contextual information e.g. to log things correctly.

To be more specific, the `errorHandler` is not used when the error happen in one of my schema executors I using (in request context).

My question: is this expected situation? I found it quite confusing. Should we try to use `errorHandler` in both error flows in some way?

wiktor-obrebski:
@mcollina any chance you can take a look? I would like to confirm if this is a real issue, or I do not understand something correctly. If I'll get confirmation I will try to provide a pull request to fix this.

mcollina:
It's ok the flow is split. If a resolver rejects/throw, that error is going to end inside errorFormatter: these errors will be part of the "usual" GraphQL response. 
The Fastify `errorHandler` is being called if there is an error _outside_ of the GraphQL flow. These are outside of the GraphQL responses.

Can you make an example that makes this confusing?

This might be better documented, would you like to send a PR?

wiktor-obrebski:
>  If a resolver rejects/throw, that error is going to end inside errorFormatter: these errors will be part of the "usual" GraphQL response. (...) Can you make an example that makes this confusing?

I need to add logic to hande resolvers errors by myself. I need to log information about this (contextually, with request info etc.) to stdout. Additionally, I want to report it to Sentry. On both cases, I want add few contextual information (like e.g. `reqId` added automatically by fastify).
But `errorFormatter` has been designed to format errors, not handle them. So there is not access to current `req` object in the function. So I can not get the contextual information I need. I guess there can be more reasons to handle the errors in custom way.

Generally I found it very confusing that some errors go thru `errorHandler` and some not. 

[...]

PR #653, opened 2021-11-17
Polishing/error formatter context

wiktor-obrebski:
This is an implementation of the idea discussed in issue #616 .
According to types in `index.d.ts` custom error formatter should be always called with request context: https://github.com/mercurius-js/mercurius/blob/master/index.d.ts#L473

In reality, it sometimes was called with context, but sometimes without it.
This PR goal is to standardize this behavior and provide request context every time the error formatter is called.

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -63,18 +63,13 @@ function toGraphQLError (err) {
 
 function defaultErrorFormatter (err, ctx) {
   let errors = [{ message: err.message }]
-  let log
-
-  if (ctx) {
-    // There is always app if there is a context
-    log = ctx.reply ? ctx.reply.log : ctx.app.log
-  }
+  // There is always app if there is a context
+  const log = ctx.reply ? ctx.reply.log : ctx.app.log
 
   if (err.errors) {
     errors = err.errors.map((error, idx) => {
-      if (log) {
-        log.error({ err: error }, error.message)
-      }
+      log.error({ err: error }, error.message)
+
       // parses, converts & combines errors if they are the result of a federated request
       if (error.message === FEDERATED_ERROR.toString()) {
         return error.extensions.errors.map(err => formatError(toGraphQLError(err)))
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -136,13 +136,17 @@ module.exports = async function (app, opts) {
     app.setErrorHandler(opts.errorHandler)
   } else if (opts.errorHandler === true || opts.errorHandler === undefined) {
     app.setErrorHandler((error, request, reply) => {
-      const context = request[kRequestContext]
-      const { statusCode, response } = errorFormatter(error, context)
+      const { statusCode, response } = errorFormatter(
+        error,
+        request[kRequestContext]
+      )
       reply.code(statusCode).send(response)
     })
   }
   const contextFn = opts.context
   const { subscriptionContextFn } = opts
+
+  app.decorateRequest(kRequestContext)
 
   const {
     path: graphqlPath = '/graphql',
@@ -173,10 +177,16 @@ module.exports = async function (app, opts) {
       return new MER_ERR_GQL_PERSISTED_QUERY_NOT_FOUND('Unknown query')
     }
 
-    const context = request[kRequestContext]
-
     // Handle the query, throwing an error if required
-    return reply.graphql(query, Object.assign(context, { pubsub: subscriber, __currentQuery: query }), variables, operationName)
+    return reply.graphql(
+      query,
+      Object.assign(
+        request[kRequestContext],
+        { pubsub: subscriber, __currentQuery: query }
+      ),
+      variables,
+      operationName
+    )
   }
 
   function executeRegularQuery (body, request, reply) {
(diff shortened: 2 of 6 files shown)


lib/routes.js
```diff
@@ -279,14 +283,20 @@ module.exports = async function (app, opts) {
     schema: postSchema(allowBatchedQueries),
     attachValidation: true
   }, async function (request, reply) {
+    // Generate the context for this request
+    const context = contextFn ? await contextFn(request, reply) : {}
+
+    request[kRequestContext] = context
+    Object.assign(request[kRequestContext], { reply, app })
```
mcollina, line 290: If `contextFn` is not set, this `Object.assign()` could be avoided.

test/routes.js
```diff
@@ -1655,7 +1698,11 @@ test('error thrown from onDisconnect is logged', t => {
   const error = new Error('error')
 
   const app = Fastify()
+
+  // override `app.log` to avoid polluting other tests
+  app.log = Object.create(app.log)
```
mcollina, line 1703: I don't think this is needed

test/routes.js
```diff
@@ -1702,6 +1749,9 @@ test('promise rejection from onDisconnect is logged', t => {
...
```
mcollina, line 1754: I don't think this is needed

mcollina (commented):
I think this change should be documented somewhere.

test/routes.js
wiktor-obrebski: I believe it is. when we do not provide `logger` instance (or config) to `Fastify` factory, it use [abstract-logging](https://www.npmjs.com/package/abstract-logging) library. The important part is that it is the same logger object instance for every fastify instance.

When this test override `app.log.error` it do it for all other future tests that use `Fastify()` factory. It was not a problem since now because of tests order and behavior, but it started to be with current code.

lib/routes.js
wiktor-obrebski: [defaultErrorFormatter](https://github.com/wiktor-obrebski/mercurius/blob/polishing/error-formatter-context/lib/errors.js#L64) assumes that if there is any context (even empty object) there is `reply` or `app` property available on it.

before it was not a problem, as the `errorFormatter` just did not have the context available in some cases. but now it have it always.

lib/routes.js
```diff
@@ -134,8 +135,9 @@ module.exports = async function (app, opts) {
   if (typeof opts.errorHandler === 'function') {
     app.setErrorHandler(opts.errorHandler)
   } else if (opts.errorHandler === true || opts.errorHandler === undefined) {
-    app.setErrorHandler((error, _, reply) => {
-      const { statusCode, response } = errorFormatter(error)
+    app.setErrorHandler((error, request, reply) => {
+      const context = request[kRequestContext]
```
mcollina, line 139: please add an `app.decorateRequest(kRequestContext)`

wiktor-obrebski:
@mcollina updated according to your advice

mcollina (commented):
Good work! Could you adjust the docs to cover this?

wiktor-obrebski:
@mcollina let me know if current docs upgrade is enough or you meant something more.

test/routes.js
mcollina: Ah awesome ok!

mcollina (approved):
lgtm

wiktor-obrebski:
@mcollina  when we could release new version? https://mercurius.dev/#/docs/api/options?id=plugin-options already provide new documentation that is not in line with last released version.

lib/errors.js
```diff
@@ -63,18 +63,13 @@ function toGraphQLError (err) {
 
 function defaultErrorFormatter (err, ctx) {
   let errors = [{ message: err.message }]
-  let log
-
-  if (ctx) {
```
SimenB, line 68: why is this condition removed? We're seeing

```
TypeError: Cannot read properties of undefined (reading 'reply')
    at defaultErrorFormatter (/app/node_modules/mercurius/lib/errors.js:67:19)
    at Object.<anonymous> (/app/node_modules/mercurius/lib/routes.js:139:40)
    at handleError (/app/node_modules/fastify/lib/reply.js:555:20)
    at onErrorHook (/app/node_modules/fastify/lib/reply.js:524:5)
    at _Reply.Reply.send (/app/node_modules/fastify/lib/reply.js:128:5)
    at done (/app/node_modules/fastify/lib/contentTypeParser.js:167:13)
    at Parser.defaultJsonParser [as fn] (/app/node_modules/fastify/lib/contentTypeParser.js:262:14)
    at IncomingMessage.onEnd (/app/node_modules/fastify/lib/contentTypeParser.js:243:27)
    at AsyncResource.runInAsyncScope (node:async_hooks:199:9)
    at IncomingMessage.emit (node:events:402:35)
```

in production after updating

mcollina: That's not supposed to happen due other constraints in this PR. That's definitely a bug but we will need a repro for it.

mcollina: I think I've found the problem, good spot. I'll open an issue

PR #656, opened 2021-11-18
Add new option useSecureParse, move secure parsing under it and add normal parsing as default

### LINKED ISSUE #650: Remove secure-json-parse from Gateway
I think we could safely consider the target services as trusted and remove the secure-json-parse call in https://github.com/mercurius-js/mercurius/blob/25500ddf95f3cfafaab23d03248ac6c52740142c/lib/gateway/request.js#L92.

Even better, we could put it behind an option.

radomird:
Closes #650

### DIFF
--- a/lib/gateway/request.js
+++ b/lib/gateway/request.js
@@ -89,12 +89,7 @@ function sendRequest (request, url, useSecureParse) {
           }
 
           try {
-            let json
-            if (useSecureParse) {
-              json = sJSON.parse(data.toString())
-            } else {
-              json = JSON.parse(data.toString())
-            }
+            const json = (useSecureParse ? sJSON : JSON).parse(data.toString())
 
             if (json.errors && json.errors.length) {
               // return a `FederatedError` instance to keep `graphql` happy
--- a/lib/gateway/request.js
+++ b/lib/gateway/request.js
@@ -58,7 +58,7 @@ function buildRequest (opts) {
   }
 }
 
-function sendRequest (request, url) {
+function sendRequest (request, url, useSecureParse) {
   return function (opts) {
     return new Promise((resolve, reject) => {
       request({
@@ -89,7 +89,12 @@ function sendRequest (request, url) {
           }
 
           try {
-            const json = sJSON.parse(data.toString())
+            let json
+            if (useSecureParse) {
+              json = sJSON.parse(data.toString())
+            } else {
+              json = JSON.parse(data.toString())
+            }
 
             if (json.errors && json.errors.length) {
               // return a `FederatedError` instance to keep `graphql` happy
@@ -102,7 +107,7 @@ function sendRequest (request, url) {
               json
             })
           } catch (e) {
-            reject(e)
+            reject(new FederatedError(e))
           }
         })
       })
--- a/test/gateway/request.js
+++ b/test/gateway/request.js
@@ -84,3 +84,82 @@ test('sendRequest method rejects when response contains errors', async (t) => {
 
   t.end()
 })
+
+test('sendRequest method should accept useSecureParse flag and parse the response securely', async (t) => {
+  const app = fastify()
+  app.post('/', async (request, reply) => {
+    return '{" __proto__": { "foo": "bar" } }'
+  })
+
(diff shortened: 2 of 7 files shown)


mcollina:
Can you please add a unit test? You can also use proxyquire.

radomird:
> Can you please add a unit test? You can also use proxyquire.

Yes of course - still in progress :)

radomird:
Ready for review.

lib/gateway/request.js
```diff
@@ -89,7 +89,12 @@ function sendRequest (request, url) {
           }
 
           try {
-            const json = sJSON.parse(data.toString())
+            let json
+            if (useSecureParse) {
+              json = sJSON.parse(data.toString())
+            } else {
+              json = JSON.parse(data.toString())
+            }
```
simoneb, line 97: nit

```suggestion
            const json ? (useSecureParse ? sJSON : JSON).parse(data.toString())
```

test/gateway/request.js
```diff
@@ -84,3 +84,82 @@ test('sendRequest method rejects when response contains errors', async (t) => {
...
+    query ServiceInfo {
+      _service {
+        sdl
+      }
+    }
+    `
+    })
+  })
+
+  // cecking for prototype leakage: https://github.com/fastify/secure-json-parse#introduction
```
simoneb, line 120:
```suggestion
  // checking for prototype leakage: https://github.com/fastify/secure-json-parse#introduction
```

test/gateway/request.js
```diff
@@ -84,3 +84,82 @@ test('sendRequest method rejects when response contains errors', async (t) => {
...
+  )({
+    method: 'POST',
+    body: JSON.stringify({
+      query: `
+    query ServiceInfo {
+      _service {
+        sdl
+      }
+    }
+    `
```
simoneb, line 116: I'm not sure I understand why we're sending a request body. what are we doing with it?

radomird: Actually nothing, I copied the previous test and forgot to remove it. Will do it now.

lib/gateway/request.js
radomird: Did you mean this maybe:
```js
const json = (useSecureParse ? sJSON : JSON).parse(data.toString())
```

test/gateway/request.js
simoneb: ok but they I'm wondering, what are we testing here? the behavior you're testing is on a plain route, not a graphql route. I'm confused

lib/gateway/request.js
simoneb: yes

test/gateway/request.js
radomird: We're testing if the response is parsed correctly with useSecureParse flag set to true.
If you look at the response you'll see it's a stringified object containing __proto__ the goal is to make sure that the secure parsing still works.

simoneb: what I don't understand is why this behavior is affecting plain HTTP requests. this is mercurius, why are we implementing behaviors which affect requests that should not be affected outside of what mercurius does?

radomird: The documentation is not very clear about this, but this article sheds a bit more light on the topic: https://www.nearform.com/blog/implementing-a-graphql-gateway-with-fastify/

The idea is that `mercurius` can work in a so called *"gateway mode"*. In that mode the server is connecting to different services, after it gets the  results of those services, it combines the output of all of them into one final schema. Now, part of the gateway is the `request.js` , that Matteo asked to be changed, and the reasoning behind is I think clear from the ticket.

simoneb: thanks, makes sense now!

simoneb (approved):
LGTM

simoneb:
@radomird doesn't the new option need to be documented?

radomird:
> @radomird doesn't the new option need to be documented?

I does, will updated the docs.

simoneb (approved):
Perfect 🚀

mcollina (approved):
lgtm

Issue #649, opened 2021-11-12
Optimize Gateway


mcollina:
It's possible to severely optimize the Gateway.

https://github.com/mercurius-js/mercurius/blob/master/lib/gateway.js#L384-L425

The logic _could_ batch the entity queries in a single GraphQL request to be sent to a target peer, fully leveraging the dataloader pattern.

simoneb:
This issue can be closed as the corresponding PR has been merged

PR #667, opened 2021-11-25
feat: optimize gateway by batching queries

giacomorebonato:
https://github.com/mercurius-js/mercurius/issues/649

* Current implementation doesn't break any tests and seems to work also when enabling batched queries.
* At the moment I am explicitly telling to the gateway instance of Fastify which ones are the services that support query batching:

```js
gateway.register(mercurius, {
  gateway: {
    services: [
      {
        name: 'add',
        allowBatchedQueries: true
      }
    ]
  }
})
```

* I copy pasted some unit test and enabled batched queries on them just for quick feeback

To do:

- [x] verify why in the existing implementation each query is picking its own context and request-headers (shouldn't it be the same) - Left a comment in `lib/gateway.js` about this
- [x] verify if error handling is consistent to not batched implementation - `test/gateway/errors-with-batching.js`
- [x] verify if Fastify gateway instance can implicitly know which services have batched queries enabled - This is not possible as I am aware
- [x] verify that a Gateway supports batched queries itself - `test/gateway/batching-on-both-gateway-and-services.js`


lib/gateway/get-query-result.js
```diff
@@ -0,0 +1,142 @@
...
+}
+
+const getBactchedResult = async ({ mergeQueriesResult, queries, serviceDefinition, service }) => {
+  const { mergedQueries, resultIndexes } = mergeQueriesResult
+  const context = queries[0].context
+  const originalRequestHeaders = queries[0].originalRequestHeaders
+  const batchedQueries = []
+
+  for (const [query, variables] of Object.entries(mergedQueries)) {
+    const document = parse(query)
```
mcollina, line 39: I'm pretty sure we have this parsed already., could we avoid it? I see we have this done in the previous code as well... so if it's too heard to fix here let's open a separate issue to remove it.

lib/gateway/get-query-result.js
```diff
@@ -0,0 +1,142 @@
...
+
+  return result
+}
+
+const getResult = async ({ mergeQueriesResult, serviceDefinition, queries, service }) => {
+  const { mergedQueries, resultIndexes } = mergeQueriesResult
+  const jsons = []
+  const queriesEntries = Object.entries(mergedQueries)
+
+  for await (const [queryIndex, [query, variables]] of queriesEntries.entries()) {
```
mcollina, line 99: These could be done in parallel, not sequentially.

mcollina:
Good work!

lib/gateway/get-query-result.js
giacomorebonato: Found! :)

mcollina:
Why a service would not support batched queries? I would enable that by default.

giacomorebonato:
> I would enable that by default

I agree with you @mcollina and I'll apply this configurable default on the Gateway, as I see it as well as a best practice.

Your question made me think though: I think that maybe a service could have batching disabled because it's not aware of being used as a service from a gateway instance.

mcollina:
> Your question made me think though: I think that maybe a service could have batching disabled because it's not aware of being used as a service from a gateway instance.

How could you disable batching in a service? I think that's part of the GraphQL spec.

giacomorebonato:
My thought goes to a Service made with Fastify + Mercurius and [batched queries disabled](https://mercurius.dev/#/docs/batched-queries):

```
app.register(mercurius, {
  schema,
  resolvers,
  graphiql: true,
  allowBatchedQueries: false // false is default value when the property is omitted
})
```

Screenshot 2021-11-26 at 12 05 05

Do you suggest that batched queries should be enabled by default in Mercurius, regardless if it's about building a service for a gateway?

I am a very new contributor to the project, so I might be missing something.

lib/gateway/get-query-result.js
```diff
@@ -0,0 +1,140 @@
+'use strict'
+
+const { preGatewayExecutionHandler } = require('../handlers')
+
+const mergeQueries = (queries) => {
```
Eomm, line 5: I would add some comments on the code to understand the context/logic

lib/gateway/get-query-result.js
```diff
@@ -0,0 +1,140 @@
...
+      }
+
+      entityIndex++
+    }
+  }
+
+  return result
+}
+
+const getResult = async ({ mergeQueriesResult, serviceDefinition, context, service }) => {
```
Eomm, line 91: Using named function could help in case of errors being thrown

```suggestion
async function getResult ({ mergeQueriesResult, serviceDefinition, context, service }) {
```

mcollina:
Thanks @giacomorebonato, I know understand this better.

When I wrote the issue originally, I thought we would use queries like the following:

```graphql
query user1 {
  ...
}

query user2 {
  ...
}
```

Instead, you implemented this using https://mercurius.dev/#/docs/batched-queries - which is a feature I even forgot we supported.

As things are, I think you implemented this correctly and it should be enabled on demand and not by default as it is off-standard.

Note that there is currently no way to execute multiple queries in parallel in GraphQL by the spec: https://github.com/graphql/graphql-spec/issues/375.

Note that batching is currently at rfc stage at: https://github.com/graphql/graphql-over-http/blob/main/rfcs/Batching.md.

Could you please set the default to back `false? I'm sorry for the confusion.

docs/federation.md
```diff
@@ -351,6 +355,40 @@ server.register(mercurius, {
...
+const server = Fastify()
+
+server.register(mercurius, {
+  graphiql: true,
+  gateway: {
+    services: [
+      {
+        name: 'user',
+        url: 'http://localhost:3000/graphql'  // queries will be batched into one request
+                                              // same as setting allowBatchedQueries: true
```
mcollina, line 377: please uncomment this

docs/federation.md
```diff
@@ -351,6 +355,40 @@ server.register(mercurius, {
 server.listen(3002)
 ```
 
+#### Batched Queries to services
+
+To fully leverage the DataLoader pattern a Gateway assumes that can send a request with batched queries to its services.  
+This configuration can be disabled if the service doesn't support batched queries instead.
```
mcollina, line 361: This does not match the implementation anymore.

mcollina (approved):
lgtm

PR #672, opened 2021-11-30
fix: typo in fetchBatchedResult

MoritzLoewenstein:
Hate to be this guy, but I just saw this typo while reading through the code. 
As far as I can tell this function is not exposed, so should be fine to change.


mcollina (approved):
lgtm

PR #676, opened 2021-12-01
Added graphql-jit compiler options

### LINKED ISSUE #390: Allow setting graphql-jit's customJSONSerializer / fast-json-stringify
I'd like to use graphql-jit's `customJSONSerializer` option, to use `fast-json-stringify` instead of `JSON.stringify`, but it looks like that isn't possible:

https://github.com/mercurius-js/mercurius/blob/361e9534f8c5e0dd693e16c7943b74b0d4c20e1e/index.js#L483

Could this be supported by mercurius's API?

I'm also a little curious if `customJSONSerializer` should just always be passed true, i.e. if its faster, just have that be the default. I had actually assumed it was, and was surprised that neither graphql-jit nor mercurius default it to on.

I guess `fast-json-stringify` does have the disclaimer that its slower on large output (...or at least "not as dramatically faster"), so maybe that is why. 

Kind of random/tangential, I also noticed the `fast-json-stringify` disclaimer about it dove-tailing really well with `flatstr`, but I don't see `flatstr` used in either mercurius or graphql-jit; is that a missed opportunity for a performance win?

sameer-coder:
This PR adds new options in mercurius which can be passed to the graphql-jit compiler.

Closes #390

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -97,10 +97,7 @@
   - `notSupportedError?: string`: An error message to return when a query matches `isPersistedQuery`, but returns no valid hash from `getHash`. Defaults to `Bad Request`.
 - `allowBatchedQueries`: Boolean. Flag to control whether to allow batched queries. When `true`, the server supports recieving an array of queries and returns an array of results.
 
-- `compilerOptions`: Object. Configurable options for the graphql-jit compiler.
-  - `disableLeafSerialization`: Boolean (Default: false). Disables leaf node serializers. The serializers validate the content of the field at runtime so this option should only be set to true if there are strong assurances that the values are valid.
-  - `customSerializers`: Object as Map (Default: {}) - Replace serializer functions for specific types. Can be used as a safer alternative for overly expensive serializers.
-  - `customJSONSerializer`: Boolean (Default: false) - Whether to produce also a JSON serializer function using fast-json-stringify. The default stringifier function is JSON.stringify
+- `compilerOptions`: Object. Configurable options for the graphql-jit compiler. For more details check https://github.com/zalando-incubator/graphql-jit
 
 #### queryDepth example
 
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -97,6 +97,11 @@
   - `notSupportedError?: string`: An error message to return when a query matches `isPersistedQuery`, but returns no valid hash from `getHash`. Defaults to `Bad Request`.
 - `allowBatchedQueries`: Boolean. Flag to control whether to allow batched queries. When `true`, the server supports recieving an array of queries and returns an array of results.
 
+- `compilerOptions`: Object. Configurable options for the graphql-jit compiler.
+  - `disableLeafSerialization`: Boolean (Default: false). Disables leaf node serializers. The serializers validate the content of the field at runtime so this option should only be set to true if there are strong assurances that the values are valid.
+  - `customSerializers`: Object as Map (Default: {}) - Replace serializer functions for specific types. Can be used as a safer alternative for overly expensive serializers.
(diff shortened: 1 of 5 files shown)


docs/api/options.md
```diff
@@ -97,6 +97,11 @@
   - `notSupportedError?: string`: An error message to return when a query matches `isPersistedQuery`, but returns no valid hash from `getHash`. Defaults to `Bad Request`.
 - `allowBatchedQueries`: Boolean. Flag to control whether to allow batched queries. When `true`, the server supports recieving an array of queries and returns an array of results.
 
+- `compilerOptions`: Object. Configurable options for the graphql-jit compiler.
+  - `disableLeafSerialization`: Boolean (Default: false). Disables leaf node serializers. The serializers validate the content of the field at runtime so this option should only be set to true if there are strong assurances that the values are valid.
+  - `customSerializers`: Object as Map (Default: {}) - Replace serializer functions for specific types. Can be used as a safer alternative for overly expensive serializers.
+  - `customJSONSerializer`: Boolean (Default: false) - Whether to produce also a JSON serializer function using fast-json-stringify. The default stringifier function is JSON.stringify
```
mcollina, line 103: Are there any performance benefits for using this?

mcollina (approved):
lgtm

index.js
```diff
@@ -554,7 +554,11 @@ const plugin = fp(async function (app, opts) {
     if (shouldCompileJit) {
       if (!modifiedSchema && !modifiedDocument) {
         // can compile only when the schema and document are not modified
-        cached.jit = compileQuery(fastifyGraphQl.schema, document, operationName)
+        if (opts.compilerOptions) {
+          cached.jit = compileQuery(fastifyGraphQl.schema, document, operationName, opts.compilerOptions)
+        } else {
+          cached.jit = compileQuery(fastifyGraphQl.schema, document, operationName)
+        }
```
simoneb, line 561: how is this different from:

```suggestion
        cached.jit = compileQuery(fastifyGraphQl.schema, document, operationName, opts.compilerOptions)
```

?

sameer-coder: you are right! Simplified version is much better

docs/api/options.md
sameer-coder: I haven't run any benchmarks right now. I will try to compare the performance and report back

docs/api/options.md
```diff
@@ -97,6 +97,11 @@
   - `notSupportedError?: string`: An error message to return when a query matches `isPersistedQuery`, but returns no valid hash from `getHash`. Defaults to `Bad Request`.
 - `allowBatchedQueries`: Boolean. Flag to control whether to allow batched queries. When `true`, the server supports recieving an array of queries and returns an array of results.
 
+- `compilerOptions`: Object. Configurable options for the graphql-jit compiler.
```
mcollina, line 100: Can you add a link instead of listing them all?

sameer-coder: Removed the sub options and added a link

test/cache.js
```diff
@@ -34,7 +34,14 @@ test('cache skipped when the GQL Schema has been changed', async t => {
...
   t.teardown(() => app.close())
 
-  await app.register(GQL, { schema, resolvers, jit: 1 })
+  await app.register(GQL, {
+    schema,
+    resolvers,
+    jit: 1,
+    compilerOptions: {
+      customJSONSerializer: true
+    }
```
mcollina, line 43: This test is not verifying anything. You'd need to use proxyquire to verify you are passing the parameters correctly.

sameer-coder: I have added a new test which checks if we are passing the `compilerOptions` correctly to the function. Please let me know if the test needs any changes.

sameer-coder:
Working on fixing the CI

mcollina (approved):
lgtm

PR #678, opened 2021-12-01
Fix regression in handling badly formed JSON

mcollina:
Fixes https://github.com/mercurius-js/mercurius/issues/677

PR #679, opened 2021-12-02
Extend regression fix in handling bad json to include custom context

wiktor-obrebski:
This PR extend #678 to include custom context in error formatter.

@mcollina I am not sure if context (custom and/or default) is needed in case when custom error handler has been used (as in is such case error formatter is not called at all with my understanding?)

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -133,22 +133,33 @@ function tryJSONParse (request, value) {
 
 module.exports = async function (app, opts) {
   const errorFormatter = typeof opts.errorFormatter === 'function' ? opts.errorFormatter : defaultErrorFormatter
+  const contextFn = opts.context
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm, good work

mcollina:
> @mcollina I am not sure if context (custom and/or default) is needed in case when custom error handler has been used (as in is such case error formatter is not called at all with my understanding?)

I think it's needed. Folks would expect it to be there.

mcollina (approved):
lgtm

wiktor-obrebski:
@mcollina can you release this, please?

mcollina:
I'm on vacation for a week or so.

wiktor-obrebski:
ok, sorry for bothering you. hava a good time ;)

PR #680, opened 2021-12-07
Clarify `preValidation` documentation

wiktor-obrebski:
The `preValidation` hook is not called when queries has been cached in `lru`. I find this quite confusing, maybe this info will help others to avoid some mistakes.


jonnydgreen (approved):
LGTM

PR #686, opened 2021-12-16
Upgrade to graphql 16

### LINKED ISSUE #628: Doesn't work with graphql v16+
Hi,
just wanted to update the graphql package to version 16 but fastify/mercurius doesn't seem to like it. 

CleanShot 2021-11-08 at 14 38 09

`No overload matches this call.
  Overload 1 of 3, '(plugin: FastifyPluginCallback<MercuriusOptions, Server>, opts?: FastifyRegisterOptions<MercuriusOptions> | undefined): FastifyInstance<...> & PromiseLike<...>', gave the following error.
    Type 'NexusGraphQLSchema' is not assignable to type 'string | GraphQLSchema | string[] | undefined'.
      Property 'isPossibleType' is missing in type 'NexusGraphQLSchema' but required in type 'GraphQLSchema'.
  Overload 2 of 3, '(plugin: FastifyPluginAsync<MercuriusOptions, Server>, opts?: FastifyRegisterOptions<MercuriusOptions> | undefined): FastifyInstance<...> & PromiseLike<...>', gave the following error.
    Argument of type 'typeof mercurius' is not assignable to parameter of type 'FastifyPluginAsync<MercuriusOptions, Server>'.
      Type 'void' is not assignable to type 'Promise<void>'.
  Overload 3 of 3, '(plugin: FastifyPluginCallback<MercuriusOptions, Server> | FastifyPluginAsync<MercuriusOptions, Server> | Promise<...> | Promise<...>, opts?: FastifyRegisterOptions<...> | undefined): FastifyInstance<...> & PromiseLike<...>', gave the following error.
    Type 'NexusGraphQLSchema' is not assignable to type 'string | GraphQLSchema | string[] | undefined'.ts(2769)
schema.d.ts(83, 3): 'isPossibleType' is declared here.`

Thanks for your help!

gdorsi:
fixes #628


test/subscription-connection.js
```diff
@@ -311,7 +311,7 @@ test('subscription connection handles GQL_START message correctly, when payload.
 
   sc.isReady = true
 
-  await sc.handleMessage(JSON.stringify({
+  sc.handleMessage(JSON.stringify({
```
gdorsi, line 314: I've removed the async/await because the message was sent after the promise resolution, causing the test to end before all expectations were met. 

Should I dig more into this?

mcollina: It's ok

lib/errors.js
```diff
@@ -26,39 +26,20 @@ class FederatedError extends Error {
...
-    positions: {
-      value: err.positions || undefined
-    },
-    originalError: {
-      value: err.originalError || undefined
-    },
-    extensions: {
-      value: err.extensions || undefined
-    }
-  })
```
mcollina, line 61: Why this change?

mcollina (approved):
Good work!

LGTM

lib/errors.js
gdorsi: I got this error `error.toJSON is not a function`

That's because they changed the `formatError` impl to simply do `return error.toJSON();`

I thought it was safer to move to `new GraphQLError`

Should I go back to `Object.create`?

Doing `Object.create(GraphQLError.prototype,` would also solve the issue

examples/graphql-shield-authorization.js
```diff
@@ -1,126 +0,0 @@
-'use strict'
```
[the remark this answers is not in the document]
gdorsi: it's dependencies are not yet compatible with graphql 16

Discussion here --> https://github.com/mercurius-js/mercurius/issues/628#issuecomment-967035794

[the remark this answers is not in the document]
nuragic: Cool thanks!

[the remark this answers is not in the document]
simoneb: rather than delete, isn't this something we can upgrade to work with graphql@16? In any case we should think about a better way other than deleting. Even if they are not compatible, I assume they will be at some point, meaning that we may want to keep the example around in some way

lib/federation.js
```diff
@@ -129,6 +129,7 @@ function getStubTypes (schemaDefinitions, isGateway) {
 
 function gatherDirectives (type) {
   let directives = []
+  /* istanbul ignore else seems that extensionASTNodes is always provided in graphql16+, we are keeping this check for safety */
```
simoneb, line 132: what is this comment doing?

test/subscription-connection.js
simoneb: not sure why this behavior would change though because of the graphql bump. do you have an explanation for why it changes @gdorsi ?

lib/federation.js
gdorsi: In graphql 15 `extensionASTNodes` could be either `null` or `Array`

Now it is only `Array` but I thought that was safer to keep the check.

I've added this comment to get the 100% coverage and explain why I didn't get rid of it.

simoneb: ok so this comment mixes an istanbul command with human-readable information.
I would recommend we keep the two separated, meaning, 2 comments

test/subscription-connection.js
gdorsi: The reason is this change https://github.com/graphql/graphql-js/commit/39581581acc184301cfbb0ee348d0f4b6a21a99f

The have added an await to the function that throws the validation error.

gdorsi: We could fix it with [this](https://github.com/gdorsi/mercurius/commit/1b05dfb0c5a707fbe4e4a586380542d837d1d36d) change to handleMessage but I'm not sure how this would affect the performance.

PR #690, opened 2021-12-21
fix: specify type better for context arg in defaultErrorFormatter

capaj:
When context is null for example, it crashes with a type error
```
TypeError: Cannot read property 'reply' of null
    at AsyncFunction.defaultErrorFormatter (/home/capaj/work-repos/authier-repos/authier/node_modules/mercurius/lib/errors.js:67:19)
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -614,7 +614,7 @@ declare namespace mercurius {
    */
   const defaultErrorFormatter: (
     execution: ExecutionResult,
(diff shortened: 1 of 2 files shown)


mcollina (commented):
Thanks! Could you add an assertion for this in our types tests.

mcollina (approved):
lgtm

PR #691, opened 2021-12-22
Implement GQL_CONNECTION_KEEP_ALIVE on subscription server

stefanosala:
Implements the [GQL_CONNECTION_KEEP_ALIVE](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_keep_alive) protocol message in order to keep the socket alive (e.g. useful on Heroku, to prevent H15 errors).

Happy to tweak test cases in case we need more coverage on the interval.


mcollina (approved):
lgtm

Issue #532, opened 2021-07-10
How to accept queries and mutations over websocket? (client-server)


xxSkyy:
I've got problem with accepting queries and mutations over websocket in mercurius and I can't find solution how to handle them. I'm implementing chat and social functions to my project and don't want every message sent over http, its overkill.
I did add to split if query/mutation has websocket property it will be sent over it than http but despite subscriptions goes well queries don't want to and I dont know way to handle it.

I've got exact same error as #195 
```
/* SENT */ {"type":"connection_init","payload":{}}
/* RECV */ {"type":"connection_ack"}
/* SENT */ {"id":"1","type":"start","payload":{"variables":{"message":"TEST!"},"extensions":{},"operationName":"InsertNotificationOverWS","query":"mutation InsertNotificationOverWS($message: String!) {\n  insertNotification(message: $message)\n}\n"}}
/* RECV */ {"type":"error","id":"1","payload":"Subscription field must return Async Iterable. Received: undefined"}
/* SENT */ {"id":"1","type":"stop"}
```

[Here](https://github.com/mercurius-js/mercurius/blob/0f101149239c2544f0dd3b912a74ef41a46e7787/docs/protocol-extension.md) is written only about server to server case when in reference are about client-server also.

How can I achieve mercurius to read queries over websocket?

mcollina:
That's not currently supported..
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

xxSkyy:
> That's not currently supported..
> Would you like to send a Pull Request to address this issue? Remember to add unit tests.

I figured it out for my project [in this fork](https://github.com/xxSkyy/mercurius) if someone wants to use it or implement fell free, I don't have enough time right now to make full PR

indatawetrust:
do you have a pull request plan for this topic?

PR #705, opened 2022-01-14
Queries and Mutations via WebSocket

### LINKED ISSUE #195: How to allow queries and mutations over a WebSocket connection?
Hello! 

So I use `fastify` and `fastify-gql` for my API gateway. I have some subscription enabled via the option flag `subscription: true`. The react client uses `import { ApolloClient, HttpLink, InMemoryCache, split } from "@apollo/client"` and `import { WebSocketLink } from "@apollo/link-ws"`. And everything works.

A part of my app is a real-time text editor and pushing operations using POST is just not resulting in a great experience. However, using a `WebSocket connection` works great.

On the server, is there a way to allow the same WebSocket connection (from enabling Subscriptions) to be used to receive and perform a `query` or `mutation` without having two different WebSocket servers side by side?

Right now, the client is setup as follows:

```typescript
import { ApolloClient, HttpLink, InMemoryCache, split } from "@apollo/client"
import { getMainDefinition } from "@apollo/client/utilities"
import { WebSocketLink } from "@apollo/link-ws"

const wsLink = new WebSocketLink({
	uri: `ws://localhost:5000/graphql`,
	options: {
		reconnect: true,
	},
})

const httpLink = new HttpLink({
	uri: "http://localhost:5000/graphql",
})

// The split function takes three parameters:
//
// * A function that's called for each operation to execute
// * The Link to use for an operation if the function returns a "truthy" value
// * The Link to use for an operation if the function returns a "falsy" value

const splitLink = split(
	(op: any) => {
		const { query, operationName } = op
		const definition = getMainDefinition(query)
		return (
			(
				// any mutation's name I want to send via WS ends with "OverWS"
				typeof operationName === `string` && operationName.endsWith(`OverWS`)
			) 
				||
			(
				definition.kind === "OperationDefinition" &&
				definition.operation === "subscription"
			)
		)
	},
	wsLink,
	httpLink
)

export const client = new ApolloClient({
	cache: new InMemoryCache(),
	link: splitLink,
})

```

```typescript
// And logs (Chrome Dev)

/* SENT */ {"type":"connection_init","payload":{}}
/* RECV */ {"type":"connection_ack"}
/* SENT */ {"id":"1","type":"start","payload":{"variables":{"message":"TEST!"},"extensions":{},"operationName":"InsertNotificationOverWS","query":"mutation InsertNotificationOverWS($message: String!) {\n  insertNotification(message: $message)\n}\n"}}
/* RECV */ {"type":"error","id":"1","payload":"Subscription field must return Async Iterable. Received: undefined"}
/* SENT */ {"id":"1","type":"stop"}
```

The server is as in the "Simple Subscription example":

```typescript
import fastify from "fastify"
import cors from "fastify-cors"
import GQL from "fastify-gql"
import { buildSchema } from "graphql"
import { PubSub } from 'graphql-subscriptions'

export const pubsub = new PubSub();

const schemaAsString = `
	type Notification {
		id: ID!
		message: String
	}
   
	type Query {
		notifications: [Notification]
	}
   
	type Mutation {
		insertNotification(message: String): Notification
		addNotification(message: String): Notification
	}
   
	type Subscription {
		notificationAdded: Notification
	}
	 `
	 
const schema = buildSchema(schemaAsString)

let idCount = 1
const notifications = [
	{
		id: idCount,
		message: "Notification message",
	},
]

export const createServer = async (port: number = 5000) => {

	const app = fastify()

	app.register(cors, { origin: true, })

	app.register(GQL, {
		graphiql: true,
		schema,
		resolvers: {
			Query: {
				notifications: () => notifications,
			},
			Mutation: {
				insertNotification: async (_, { message }) => {
					const id = idCount++
					const notification = {
						id,
						message,
					}

					notifications.push(notification)

					pubsub.publish(`NOTIFICATION_ADDED`, { 
						notificationAdded: notification 
					});

					return notification
				}
			},
			Subscription: {
				notificationAdded: {
					subscribe: () => pubsub.asyncIterator(`NOTIFICATION_ADDED`),
				},
			},
		},
		subscription: true,
	})

	app.listen(port)

	console.log(`Listening on port ${port}!`)

	return app
}
```

Thank you! And very appreciative of any pointer!

nuragic:
This proposal adds a new flag, `subscription.fullWsTransport`: when set to `true` we also allow `Query` and `Mutation` root operations via the subscription channel (WebSocket).

There're also some minor changes (docs, etc). _See commits_.

Resolves #195 (and more related I guess, e.g. #532).


nuragic:
i've added an [example server](https://github.com/mercurius-js/mercurius/pull/705/files#diff-77373300cf7c26999175d334503d26584909b271c78cc7bea0fe761b565b8f0a), it can be tested with this client app: https://github.com/nuragic/graphql-react-app

nuragic:
I've skipped 2 tests, I put a comment on each one, I think they're not valid because they're violating the spec protocol.

test/routes.js
```diff
@@ -2037,3 +2037,28 @@ test('if operationName is null, it should work fine', async (t) => {
...
+  app.register(GQL, {
+    schema,
+    fullWsTransport: true
+  })
+
+  const resGet = await app.inject({
+    method: 'GET',
+    url: '/graphiql'
+  })
+  t.equal(resGet.statusCode, 404)
```
mcollina, line 2057: This should not be the case.

nuragic: @mcollina Yeah this is a typo, thanks! I meant the `/graphql` endpoint... the idea behind this one is that if you set the `fullWsTransport: true`  maybe you want to _force_ every operation via WS, so you don't want the HTTP endpoints to be accessible... but maybe it doesn't make sense?

mcollina: I would not do that.

docs/protocol-extension.md
```diff
@@ -1,6 +1,8 @@
 # GraphQL over WebSocket Protocol extension
 
-The GraphQL over WebSocket Protocol used by `mercurius` follows apollo's protocol defined [here](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md). In order to achieve gateway-to-service communication and handle `connection_init` per client, an extension is used on the protocol for server-to-server communication.
+The GraphQL over WebSocket Protocol `mercurius` uses by default is [graphql-transport-ws](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md). It also supports the protocol used by Apollo's `subscriptions-transport-ws`; however that library is **unmantained** so it's not recommended to use.
```
nuragic, line 3: If possible, I'd deprecate the old graphql-ws protocol in the next major, seems not worth to keep supporting it because the `subscriptions-transport-ws` states it's unmaintained.

docs/protocol-extension.md
```diff
@@ -1,6 +1,8 @@
...
+
+In order to achieve gateway-to-service communication and handle `connection_init` per client, an extension is used on the protocol for server-to-server communication. See https://github.com/mercurius-js/mercurius/issues/268 for more details about the original issue and the actual implemented solution.
```
mcollina, line 5: Which clients support WS as a complete transport? We should list compatible clients here.

docs/api/options.md
```diff
@@ -60,6 +60,7 @@
   - `subscription.onConnect`: `Function` A function which can be used to validate the `connection_init` payload. If defined it should return a truthy value to authorize the connection. If it returns an object the subscription context will be extended with the returned object.
   - `subscription.onDisconnect`: `Function` A function which is called with the subscription context of the connection after the connection gets disconnected.
   - `subscription.keepAlive`: `Integer` Optional interval in ms to send the `GQL_CONNECTION_KEEP_ALIVE` message.
+  - `subscription.fullWsTransport`: `Boolean` Enable sending every operation via WS.
```
mcollina, line 63: I think some more heavy-handed changes to our docs are needed. I think we should flip the config and  have a `websocket` top level option.

package.json
```diff
@@ -6,6 +6,7 @@
   "types": "index.d.ts",
   "scripts": {
     "docs": "docsify serve",
+    "tap": "tap",
```
mcollina, line 9: why this change?

mcollina (changes requested):
What clients support this protocol?

package.json
nuragic: Ah sorry, I'll undo this (added to run a single test e.g. `npm run tap test/subscription-connection.js`)

docs/api/options.md
nuragic: Ok, make sense! But do we want to assume a breaking change? How would you structure the new config?

docs/protocol-extension.md
nuragic: Well, as far as I know: apollo, urql, react-query, graphql-ws... For example, graphql-hooks doesn't support this use case, but I was thinking to make a PR to add support for it after this PR would land 😄 

Anyways, this document refers to a specific use case for the `extension` field, and was added on #271: my changes are not related to this PR in reality (I just fixed/improved it a little bit because it was outdated/unclear).

nuragic: Ok I'll convert this doc into "GraphQL over WebSocket" so we can add more related sections (extensions, supported clients, etc).

docs/api/options.md
nuragic: I think this is the only point left so far, but I'm not sure about how to introduce those changes into the config/docs right now.

mcollina: Let's not do a breaking change but support both the new and old options.

docs/graphql-over-websocket.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Protocol
+
+The GraphQL over WebSocket Protocol (i.e. the WebSocket sub-protocol) `mercurius` uses by default is [graphql-transport-ws](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md). It also supports the protocol used by Apollo's `subscriptions-transport-ws`; however that library is **unmaintained** so it's not recommended to use.
+
+## Extensions
+
+The `extensions` field is reserved for things that implementors want to add on top of the spec.
+
+### Message structure
```
mcollina, line 17: This document does not specify enough information to implement this protocol.

nuragic: This document was already here, I just added a little bit info to it, It was added with this PR: https://github.com/mercurius-js/mercurius/pull/271

Anyways, I don't mind adding any missing stuff... the problem is, I don't understand what is missing 😅 

If you're referring to the "extensions" or "message structure" section, I haven't touched anything about it (part of the PR linked above).

If you're referring to the GraphQL over WS protocol itself, it is documented here:
> The GraphQL over WebSocket Protocol (i.e. the WebSocket sub-protocol) `mercurius` uses by default is [graphql-transport-ws](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md).

mcollina: https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#single-result-operation ah it's here. Let's just directly link this?

nuragic: > https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#single-result-operation ah it's here. Let's just directly link this?

I'll add a new paragraph with that reference as well if you think it's worth! But it's just a section of the already linked protocol 😄

docs/api/options.md
nuragic: Thinking about adding a `websocket` option on the config top-level, not sure it would be better... if you mean to add it to group related configs, I think that  `subscription` is actually a better choice, taking in mind that subscriptions could also be implemented with SSE (there's already an open discussion about that), and also that doesn't make sense to setup any websocket if you don't use subscriptions... 🤔

docs/graphql-over-websocket.md
nuragic: Pushed some changes on the docs, I hope it's more clear now 🧹

nuragic:
Once this lands, we can automatically ship https://github.com/nearform/graphql-hooks/pull/767 which adds support for using this feature, there's also an example app on there.

mcollina (changes requested):
The linting has been significantly changed in some files, making the review of this kind of impossible. Could you please restore the previous linting and apply your changes?

index.js
```diff
@@ -40,9 +40,24 @@ const {
   MER_ERR_METHOD_NOT_ALLOWED,
   MER_ERR_INVALID_METHOD
 } = require('./lib/errors')
-const { Hooks, assignLifeCycleHooksToContext, assignApplicationLifecycleHooksToContext } = require('./lib/hooks')
-const { kLoaders, kFactory, kSubscriptionFactory, kHooks } = require('./lib/symbols')
-const { preParsingHandler, preValidationHandler, preExecutionHandler, onResolutionHandler, onGatewayReplaceSchemaHandler } = require('./lib/handlers')
```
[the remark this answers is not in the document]
nuragic: @mcollina If I run `npm run lint:fix` nothing changes... in fact, the format changes you mention were made by the configured linter (standard). 🤷‍♂️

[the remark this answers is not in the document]
mcollina: > If I run npm run lint:fix nothing changes...

These lines pass linting. However they make reviewing this PR impossible.

> in fact, the format changes you mention were made by the configured linter (standard). 🤷‍♂️

This is not correct as running standard does not make those changes.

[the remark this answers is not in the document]
nuragic: I don't want to change anything outside the scope of this PR, indeed... maybe those lines changed because of my prettier extension for VS code, but, to be fair, by running `npm run lint:fix` it should undo those format changes... right? Anyways, if standard can't undo them I'll try to fix this manually. Sorry for the inconvenience.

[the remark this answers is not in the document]
nuragic: @mcollina I've fixed the formatting issues

docs/api/options.md
mcollina: let's go with this option and we could refactor later

lib/subscription-connection.js
```diff
@@ -171,73 +171,98 @@ module.exports = class SubscriptionConnection {
...
     }
 
-    let subscriptionLoaders
-    if (this.fastify && this.fastify[kSubscriptionFactory]) {
-      subscriptionLoaders = this.fastify[kSubscriptionFactory]
-    }
+    const operationAST = getOperationAST(document, operationName)
+
+    if (operationAST.operation === 'subscription') {
+      // execute Subscription
```
mcollina, line 183: Can you extract the content of this if in its own function? This function is becoming unreadable.

lib/subscription-connection.js
```diff
@@ -171,73 +171,98 @@ module.exports = class SubscriptionConnection {
...
+      this.fullWsTransport === true && (operationAST.operation === 'query' || operationAST.operation === 'mutation')) {
+      // execute Query or Mutation (errors are still caught and handled by the caller of this function)
+      const response = await this.fastify.graphql(
+        query,
+        context,
+        variables,
+        operationName
+      )
+      this.sendMessage(this.protocolMessageTypes.GQL_DATA, data.id, response)
+      this.sendMessage(this.protocolMessageTypes.GQL_COMPLETE, data.id, null)
```
mcollina, line 262: Can you please extract the content of this if into its own function? Thanks

mcollina (approved):
lgtm

lib/subscription-connection.js
```diff
@@ -171,18 +171,50 @@ module.exports = class SubscriptionConnection {
       }
     }
 
+    const document = typeof query !== 'string' ? query : parse(query)
+
+    if (!document) {
+      throw new Error('Must provide document.')
```
jonnydgreen, line 177: We should probably define a specific Mercurius error here. Also, is this covered in the tests? I don't see a test case for it here, unless I'm missing something?

jonnydgreen (commented):
Looks good - couple of minor comments!

index.d.ts
```diff
@@ -518,6 +518,7 @@ export interface MercuriusCommonOptions {
         }) => Record<string, any> | Promise<Record<string, any>>;
         onDisconnect?: (context: MercuriusContext) => void | Promise<void>;
         keepAlive?: number,
+        fullWsTransport?: boolean,
```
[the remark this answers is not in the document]
nuragic: Hey thanks for the review! What do you mean by type test? 😄  Sorry but I have like 0 experience with TypeScript.

lib/subscription-connection.js
nuragic: Well, seems it's covered by tests right now, 'Must provide document.' is expected in 1 test. Anyway, this is the exact same error message thrown by the graphql.js lib itself... https://github.com/graphql/graphql-js/search?q=%22Must+provide+document%22

I don't know, maybe we can define a specific one for Mercurius, if you have any specific suggestion... otherwise I would need to analyse the flow a bit more.

index.d.ts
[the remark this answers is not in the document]
jonnydgreen: Hi, no problem at all!

We have type tests to verify that our TypeScript type definitions match what we want. A good analogy is the dream query in GraphQL. I.e. one can specify a dream query and then we verify that the dream query satisfies the GraphQL schema. This is exactly what we're doing with the type tests.

You can find the type tests in this file: https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts

Essentially, all you need to do is to define a registration with `fullWsTransport`, run `npm run typescript` and you're good to go. See the following as an example: https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts#L327

Hope that helps?

lib/subscription-connection.js
jonnydgreen: Cool, if it's covered by existing tests then all good for me - I must've missed it in the file diff

index.d.ts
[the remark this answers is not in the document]
nuragic: Many thanks for the explanation, got it! Will do it as soon as possible.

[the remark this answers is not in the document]
nuragic: Just pushed the change. I get no output after running `npm run typescript`, is that ok?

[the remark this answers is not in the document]
jonnydgreen: yep, that's expected, as long as it doesn't error :)

jonnydgreen (approved):
LGTM

nuragic:
Thanks folks!!! 🤟

PR #706, opened 2022-01-17
Retry failed services

### LINKED ISSUE #520: How to handle downstream/network errors on gateway initialization?
I have encountered the scenario that federated services may not be up yet when the gateway has initialized. In this case, the gateway will not retry polling those services in question. It's a different behavior as opposed to when everything (all services, the gateway) is up and running and, say, one of the federated services goes down. In this case, the gateway will retry polling that service according to the set 'polling'-interval. This is what I  would also expect on errors with non-ready federated services on gateway startup.

Is this a scenario overlooked in the polling mechanism or is it my task to handle this error? If the latter, how would I ask the gateway to retry connection/polling to the service in question in the errorHandler? The example in the docs only shows simple error logging but how could I actually address the problem?

The workaround right now is to restart the gateway when all federated services are up and running. This is fine for development but I would rather make sure the gateway always retries reaching out to the configured services on gateway startup errors.

Greets
-act

sameer-coder:
This PR adds functionality to retry any services that might have failed or are slow to start when mercurius starts up in gateway mode. At the moment I have hardcoded the timer to be 1 second but we can change it to a whatever seems appropriate.

Open to any feedback to improve the implementation.

Closes #520

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -227,7 +227,7 @@ interface ServiceConfig {
 }
 
 interface Gateway {
-  refresh: (isRetry: boolean) => Promise<GraphQLSchema | null>;
+  refresh: (isRetry?: boolean) => Promise<GraphQLSchema | null>;
   serviceMap: Record<string, ServiceConfig>;
 }
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -227,7 +227,7 @@ interface ServiceConfig {
 }
 
 interface Gateway {
-  refresh: () => Promise<GraphQLSchema | null>;
+  refresh: (isRetry: boolean) => Promise<GraphQLSchema | null>;
   serviceMap: Record<string, ServiceConfig>;
 }
 
--- a/index.js
+++ b/index.js
@@ -188,9 +188,8 @@ const plugin = fp(async function (app, opts) {
 
         const context = assignApplicationLifecycleHooksToContext({}, fastifyGraphQl[kHooks])
         const serviceInfo = await gateway.refresh(isRetry)
-        if (!serviceInfo) return
 
-        const { schema = null, failedMandatoryServices = [] } = serviceInfo
+        const { schema, failedMandatoryServices } = serviceInfo
         if (!failedMandatoryServices.length) {
           clearInterval(gatewayRetryIntervalTimer)
         }
@@ -233,9 +232,9 @@ const plugin = fp(async function (app, opts) {
             if (serviceInfo && serviceInfo.schema) {
               // Trigger onGatewayReplaceSchema hook
               if (context.onGatewayReplaceSchema !== null) {
-                await onGatewayReplaceSchemaHandler(context, { instance: app, schema })
+                await onGatewayReplaceSchemaHandler(context, { instance: app, schema: serviceInfo.schema })
               }
-              fastifyGraphQl.replaceSchema(schema)
+              fastifyGraphQl.replaceSchema(serviceInfo.schema)
             }
           } catch (error) {
             app.log.error(error)
--- a/index.js
+++ b/index.js
@@ -176,6 +176,7 @@ const plugin = fp(async function (app, opts) {
 
   let entityResolversFactory
   let gatewayRetryIntervalTimer = null
+  const retryServicesCount = gateway && gateway.retryServicesCount ? gateway.retryServicesCount : 10
 
   const retryServices = (interval) => {
     let retryCount = 0
@@ -183,11 +184,15 @@ const plugin = fp(async function (app, opts) {
 
     return setInterval(async () => {
       try {
-        if (retryCount === 10) isRetry = false
+        if (retryCount === retryServicesCount) {
+          clearInterval(gatewayRetryIntervalTimer)
+          isRetry = false
+        }
         retryCount++
 
         const context = assignApplicationLifecycleHooksToContext({}, fastifyGraphQl[kHooks])
         const serviceInfo = await gateway.refresh(isRetry)
+        if (!serviceInfo) return
 
         const { schema, failedMandatoryServices } = serviceInfo
         if (!failedMandatoryServices.length) {
@@ -207,12 +212,12 @@ const plugin = fp(async function (app, opts) {
   }
 
   if (gateway) {
+    const retryInterval = gateway.retryServicesInterval || 3000
     gateway = await buildGateway(gateway, app)
 
     const serviceMap = Object.values(gateway.serviceMap)
     const failedMandatoryServices = serviceMap.filter(service => !!service.error && service.mandatory)
 
-    const retryInterval = 3000
     if (failedMandatoryServices.length) {
       gatewayRetryIntervalTimer = retryServices(retryInterval)
       gatewayRetryIntervalTimer.unref()
--- a/index.js
+++ b/index.js
@@ -175,18 +175,23 @@ const plugin = fp(async function (app, opts) {
   }
 
   let entityResolversFactory
-  let gatewayInterval = null
   let gatewayRetryIntervalTimer = null
 
   const retryServices = (interval) => {
-    return setInterval(async () => {
-      try {
+    let retryCount = 0
+    let shouldRetry = true
+
+    try {
+      return setInterval(async () => {
+        if (retryCount === 10) shouldRetry = false
+        retryCount++
+
         const context = assignApplicationLifecycleHooksToContext({}, fastifyGraphQl[kHooks])
-        const serviceInfo = await gateway.refresh()
+        const serviceInfo = await gateway.refresh(shouldRetry)
         if (!serviceInfo) return
 
-        const { schema, failedServices } = serviceInfo
-        if (!failedServices.length) {
+        const { schema = null, failedMandatoryServices = [] } = serviceInfo
+        if (!failedMandatoryServices.length) {
           clearInterval(gatewayRetryIntervalTimer)
         }
         if (schema !== null) {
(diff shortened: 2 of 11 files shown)


mcollina (changes requested):
I think we should have a limit on how many times a failed service could be retried.
I would prefer the logic to be separated from the `gatewayInterval` timer because I do not want the gateway to "become ready" if there are failed services that are `mandatory`.

If there are services that are not `mandatory` then we can go ahead without them (and not retry, there is the polling interval for those).

sameer-coder:
The CI is failing due to coverage being less than 100. I will am working on updating the tests.

index.js
```diff
@@ -175,9 +175,50 @@ const plugin = fp(async function (app, opts) {
...
+          // Trigger onGatewayReplaceSchema hook
+          if (context.onGatewayReplaceSchema !== null) {
+            await onGatewayReplaceSchemaHandler(context, { instance: app, schema })
+          }
+          fastifyGraphQl.replaceSchema(schema)
+        }
+      }, interval)
+    } catch (error) {
+      app.log.error(error)
+    }
```
mcollina, line 207: This creates a memory leak in case of an error. Move the try catch inside the setInterval function.

lib/gateway/service-map.js
```diff
@@ -201,7 +201,7 @@ async function buildServiceMap (services, errorHandler) {
       serviceConfig = await serviceMap[service.name].init()
     } catch (err) {
       serviceConfigErr = err
-      errorHandler(err, service)
+      if (!service.mandatory) errorHandler(err, service)
```
mcollina, line 204: we should probably log if it's not mandatory

test/gateway/schema.js
```diff
@@ -2385,7 +2385,7 @@ test('Update the schema', async (t) => {
   gateway.graphql.gateway.serviceMap.working.setSchema(fullSchema)
   const newSchema = await gateway.graphql.gateway.refresh()
 
-  gateway.graphql.replaceSchema(newSchema)
+  gateway.graphql.replaceSchema(newSchema.schema)
```
mcollina, line 2388: why this change?

lib/gateway.js
```diff
@@ -379,19 +379,27 @@ async function buildGateway (gatewayOpts, app) {
     serviceMap,
     entityResolversFactory: factory,
     pollingInterval: gatewayOpts.pollingInterval,
-    async refresh () {
+    async refresh (shouldRetry) {
```
mcollina, line 382: I don't understand the naming of this parameter

mcollina (changes requested):

test/gateway/schema.js
sameer-coder: the `gateway.refresh()` function now returns an object `{ schema, failedServices }` instead of only the `schema` which is why I had to change this.

lib/gateway.js
sameer-coder: It's a boolean parameter which specified if the retries should be done or not. should I change it to something like `isRetry`?

lib/gateway/service-map.js
sameer-coder: Ok I will add a logline before calling the error handler for non mandatory services

test/gateway/schema.js
mcollina: Unfortunately this would be a semver-major change that I would like to avoid. Can you rearrange the API differently?

mcollina (changes requested):

sameer-coder:
@mcollina There are quite a few polling related tests failing. I am trying to find what the issue is and will update the PR soon.

index.js
```diff
@@ -175,9 +175,49 @@ const plugin = fp(async function (app, opts) {
...
   let entityResolversFactory
+  let gatewayRetryIntervalTimer = null
+
+  const retryServices = (interval) => {
+    let retryCount = 0
+    let isRetry = true
+
+    return setInterval(async () => {
+      try {
+        if (retryCount === 10) isRetry = false
```
jonnydgreen, line 186: should we make this configurable?

index.js
```diff
@@ -175,9 +175,49 @@ const plugin = fp(async function (app, opts) {
...
+    }, interval)
+  }
+
   if (gateway) {
     gateway = await buildGateway(gateway, app)
 
+    const serviceMap = Object.values(gateway.serviceMap)
+    const failedMandatoryServices = serviceMap.filter(service => !!service.error && service.mandatory)
+
+    const retryInterval = 3000
```
jonnydgreen, line 215: should we make this configurable?

jonnydgreen (commented):
It would be great to include some documentation on this retry flow as well

index.js
mcollina: I think it would be a good idea

index.js
mcollina: I think it would be a good idea

index.js
sameer-coder: cool. I will make it configurable.

sameer-coder:
When I run the tests on my local machine they are all passing. I am not sure what's wrong but I think one of the timer related tests  fails 1 out of 5 times.

test/gateway/retry-failed-services.js
```diff
@@ -0,0 +1,513 @@
...
+      {
+        message: 'Cannot query field "posts" on type "User".',
+        locations: [{ line: 6, column: 9 }]
+      }
+    ],
+    data: null
+  })
+})
+
+test('gateway - stop retrying after no. of retries exceeded', async (t) => {
```
sameer-coder, line 400: I guess this is the problematic test which is failing the CI at times

jonnydgreen: From past experience, it could be related to the hardcoding of the port to 500(1|2) in the tests if it's failing intermittently in the CI. I would suggest passing `0` to `.listen()` if you can. If you need access to the port for use in the gateway config, you could listen straightaway and trigger errors dynamically using hooks (there are alternative methods, for sure! :) ). As an example:

```js
let shouldError = true
const service2 = await createTestService(postService.schema, postService.resolvers)
service2.addHook('onRequest', async () => {
  if (shouldError) {
    throw new Error('kaboom')
  }
})
await service2.listen(0)

// some work ...

shouldError = false
```

jonnydgreen: Any luck on this? I'm available to help debug this, this week, if you need :)

sameer-coder: Thanks a lot @jonnydgreen  No luck so far. I tried the way you suggested but the tests are still failing. Seems the problem lies elsewhere. I am working on this today and will dig deeper. I will provide an update soon. Thanks!

jonnydgreen: Awesome, sounds good!

mcollina:
@jonnydgreen can you do a final check?

sameer-coder:
@jonnydgreen Would appreciate if you could run the tests once just to make sure that the issue with tests has been resolved. Thanks.

lib/gateway.js
```diff
@@ -446,7 +460,7 @@ async function buildGateway (gatewayOpts, app) {
 
       defineResolvers(schema, typeToServiceMap, serviceMap, typeFieldsToService)
 
-      return schema
+      return { schema, failedMandatoryServices }
```
jonnydgreen, line 463: I'm guessing this will be a semver major change? We should update the typings (with type tests) as well: https://github.com/mercurius-js/mercurius/blob/6943453acd1f0bc1787d1b23695230227e890c58/index.d.ts#L230

jonnydgreen (commented):
Functionally all good - just a comment on the typings to be updated for the `.refresh` return value

lib/gateway.js
```diff
@@ -379,19 +379,33 @@ async function buildGateway (gatewayOpts, app) {
...
+            // failedMandatoryServices so it can be returned for retrying
+            if (service.mandatory) {
+              failedMandatoryServices.push(service)
+            }
           })
         )
       )
 
+      if (failedMandatoryServices.length > 0) {
+        return { schema: null, failedMandatoryServices }
```
mcollina, line 406: Can you please avoid this breaking change? Instead of chancing the return type, you can throw an error here.

lib/gateway.js
mcollina: It's better to avoid the breaking change.

mcollina (changes requested):

sameer-coder:
I've updated the code so that it's no longer a breaking change

index.d.ts
```diff
@@ -227,7 +227,7 @@ interface ServiceConfig {
 }
 
 interface Gateway {
-  refresh: () => Promise<GraphQLSchema | null>;
+  refresh: (isRetry: boolean) => Promise<GraphQLSchema | null>;
```
jonnydgreen, line 230: I would say this is still a breaking change from a type PoV since the current type tests won't pass with this change (we can no longer call `.refresh()` without compilation errors in TypeScript). Can we make the input optional?

sameer-coder: Sorry, I didn't notice that. The input is optional. I will change it to `refresh: (isRetry?: boolean)`

sameer-coder: I have changed it. I think the PR should be ready now.

lib/gateway.js
```diff
@@ -379,19 +379,33 @@ async function buildGateway (gatewayOpts, app) {
...
+            // failedMandatoryServices so it can be returned for retrying
+            if (service.mandatory) {
+              failedMandatoryServices.push(service)
+            }
           })
         )
       )
 
+      if (failedMandatoryServices.length > 0) {
+        throw new MER_ERR_SERVICE_RETRY_FAILED()
```
jonnydgreen, line 406: Do we have enough info to infer which mandatory service failed?

jonnydgreen (commented):
LGTM - just one minor comment about when mandatory services fail

lib/gateway.js
sameer-coder: yes we do have the service name and other info. Do you suggest we add the service name to the error message?

mcollina: Let's do that, I would also attach them to a property of the error

mcollina (approved):
Good work @sameer-coder!!

jonnydgreen (approved):
LGTM - can't wait to use this!

PR #710, opened 2022-01-31
Changed error handling for non-GraphQL errors

### LINKED ISSUE #543: Gateway Service HTTP Errors
It looks like the Mercurius Gateway service expects "errors" to exist in the response, however, if I have a fastify plugin that throws an error before it ever hits GraphQL, then the gateway emits an error.

```json
{
  "errors": [
    {
      "message": "Cannot read property 'profile' of undefined",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "profile"
      ]
    }
  ],
  "data": null
}
```

It might be nice if Mercurius handled common HTTP errors (Unauthorized, Forbidden, etc) so that it just emits that error on those types of exceptions, or if each gateway service could configure it's response to format the error if it exists.

sameer-coder:
Closes #543

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -80,13 +80,19 @@ function defaultErrorFormatter (err, ctx) {
     log.info({ err }, err.message)
   }
 
-  return {
-    statusCode,
-    response: {
-      data: err.data || null,
-      errors
-    }
-  }
+  // If error is not a GraphQLError, then send error message only
+  const response = err.errors
+    ? {
+        data: err.data || null,
+        errors
+      }
+    : {
+        statusCode,
+        error: 'Internal Server Error',
+        message: err.message
+      }
+
+  return { statusCode, response }
 }
 
 function addErrorsToExecutionResult (execution, errors) {


mcollina (commented):

sameer-coder:
Added a new test case and updated some existing ones. Please review

simoneb:
I don't think this is doing what we want to do. Don't we want to keep returning an array of errors but simply provide a more informative error about the inner error?

sameer-coder:
> 

The inner error message is already being set as the `message` property on the error. I am not sure if we have any more information about the error other than the error message.

mcollina (approved):
lgtm

mcollina:
@jonnydgreen could you take a look?

jonnydgreen (commented):
Functionally looks good, I think it would be good to add some documentation about this behaviour?

sameer-coder:
@jonnydgreen I have updated the docs. Can you please take a look? Thanks

jonnydgreen (approved):
LGTM

jonnydgreen:
> @jonnydgreen this seems a semver-major change. wdyt?

Yeah definitely - especially we will now be returning a different error responses in certain situations

mcollina:
right! sounds like a plan!

mcollina:
I had to revert this change because of #860, we must use 200 everywhere.

simoneb:
Amazing memory @mcollina !

PR #719, opened 2022-02-03
Fixes issue with batchedQueries overriding loader

flovilmart:
When using `allowBatchedQueries` the loader would get overwritten for each query being processed
which would ultimately prevent the loaders being reuses for the whole batch.

The fix is pretty native, and only overwrittes the reply's loader if not provided if in allowBatchedQueries mode.

The behaviour for non `allowBatchedQueries` should be preserved.


mcollina (approved):
lgtm, good spot!

flovilmart:
🎉

PR #720, opened 2022-02-04
fix: move `graphql` from deps to dev deps

SimenB:
Since it's defined in `peerDependencies`, you should use the version provided by your dependents, not bring your own.


mcollina:
See the discussion in #720. We had to do this to support multiple versions of npm, yarn and pnpm.

mcollina:
Does this cause any actual problem?

SimenB:
#720 is this PR 😅

SimenB:
No, not really. But it makes the peer deep not as useful as it should. Happy to take a look at some underlying issue, sounds like a hoisting bug in som older pacakge manager

mcollina:
Ah sorry: https://github.com/mercurius-js/mercurius/pull/711.

mcollina:
> No, not really. But it makes the peer deep not as useful as it should. Happy to take a look at some underlying issue, sounds like a hoisting bug in some older pacakge manager

The idea for this library is that it needs GraphQL v16. It won't work on multiple GraphQL versions. The peerDependency is there to tell the package manager to hoist it.

SimenB:
Right, but `peerDependency` is to allow the user to inject a version, and ensure there aren't duplicates in the dep tree. If you have it as `dependency` you'll always get that anyways, and the peer dep serves no real purpose. (except warn if a user uses something else, but that doesn't matter since you bring your own anyways)

But yeah, #711 seems like a good place to discuss. And from https://github.com/mercurius-js/mercurius/pull/711#issuecomment-1027606539 the solution is actually https://github.com/zalando-incubator/graphql-jit/pull/156 🙂

PR #727, opened 2022-02-08
[Gateway] Correctly resolve references returned from top-level resolvers

### LINKED ISSUE #726: Federated node can't return external Type
Hi,
I'm facing this issue: when a query returns an external type, the right node is not called.
By calling `meDirect` the Gateway tries to fetch the `Node 1` instead of the `Node 2`

### Schema Node 1:

```gql
extend type Query {
  me: User
}

type User @key(fields: "id") {
  id: ID!
  name: String
  username: String
}
```

### Schema Node 2:

```gql
extend type Query {
  meWrap: Wrap
  meDirect: User
}

type Wrap {
  user: User
}

  id: ID! @external
}
```

|query|results|
|---|---|
`meWrap { user { name } }` | works
`meDirect { id name }` | 🚫 error

The error:

```
{
  "data": {
    "meDirect": null,
  },
  "errors": [
    {
      "message": "Cannot query field \"meDirect\" on type \"Query\".",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ]
}
```

Here is a test that replicates the error:

https://github.com/Eomm/mercurius/blob/bug-external-type/test/federation-external.js

----

Edit:

we found that switching these two lines remove the error output but the `@external` types are not resolved.

```diff
git diff
diff --git a/lib/gateway.js b/lib/gateway.js
index e6d4fbc..a4296ba 100644
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -148,7 +148,7 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
             /**
              * If there is no service for the type, it is a query, mutation or subscription
              */
-            let service = serviceMap[serviceForFieldType]
+            let service = serviceMap[typeFieldsToService[`${type}-${fieldName}`]]
             if (!service) {
               /**
                * If the return type of a query, subscription or mutation is a value type, its service is undefined or null, e.g. for
@@ -158,7 +158,7 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
                * where SomeReturnType is a value type.
                * In these cases, we get the service from the typeFieldsToService map.
                */
-              service = serviceMap[typeFieldsToService[`${type}-${fieldName}`]]
+              service = serviceMap[serviceForFieldType]
             }
             if (!service) {
               /**
```

mcollina:
Fixes #726

### DIFF
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -382,7 +382,7 @@ function generatePathKey (path) {
  *  - Field entity resolver: when the field was added through type extension in the service of the field's type
  *
  */
-function makeResolver ({ service, createOperation, transformData, isQuery, isReference, isSubscription, typeToServiceMap, serviceMap }) {
+function makeResolver ({ service, createOperation, transformData, isQuery, isReference, isSubscription, typeToServiceMap, serviceMap, name }) {
   return async function (parent, args, context, info) {
     const {
       fieldNodes,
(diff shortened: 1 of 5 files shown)


lib/gateway/make-resolver.js
```diff
@@ -503,11 +505,46 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
...
+            operation = createEntityReferenceResolverOperation({
+              returnType: transformedTypeName,
+              selections,
+              variableDefinitions: []
+            })
+
+            query = print(operation)
+
+            variables.representations = [{
+              ...removeNonIdProperties(transformed, schema.getType(transformedTypeName)),
```
Eomm, line 530: If the `removeNonIdProperties` result is an empty json, could we skip the entity resolver resolution?

For example, does the following query need to be fetched in any case to let the `__resolveReference` does the process?

```gql
meDirect { id }
```

mcollina: Now that I look into it I'm not 100% convinced that removeNonIdProperties is the actual way to implement that behavior. However it would totally be useful and I will open up a follow-up issue to discuss it.

codeflyer:
I've found a bug when a list of types i returned.

The value of the elements is equal to the last one for each element.

Eg.

in the `meList` resolver in the test instead of returning a single element return more than one.
```
    meList: () => {
        return [{ id: '1', __typename: 'User' }, { id: '2', __typename: 'User' }]
      }
```

The expected result is 

```
{"data":{"meList":[{"id":"1","name":"John"},{"id":"2","name":"Jane"}]}}
```

instead is returned

```
{"data":{"meList":[{"id":"2","name":"Jane"},{"id":"2","name":"Jane"}]}}
```

mcollina:
Fixed

PR #730, opened 2022-02-11
test: add failing test for type redefinition

codeflyer:
ref https://github.com/mercurius-js/mercurius/issues/731


mcollina:
this was fixed and released

PR #733, opened 2022-02-11
fix: gateway empty types

Eomm:
Follow up #726 

When the resolvers returns empty states (`null` or `[]`) an error is thrown


mcollina (approved):
lgtm

PR #736, opened 2022-02-15
fix: update graphql-jit

SimenB:
This resolves a peer dependency warning for graphql@16.

(I realise it's in semver range, but forcing the update so the dependency tree is guaranteed to be valid also after aggressive deduplication makes sense)


mcollina (approved):
lgtm

PR #739, opened 2022-02-18
fix: gateway fragment usage

Eomm:
The federated node was not receiving the `Fragment` declaration from the gateway, so it was triggering an `Unknown fragment` error


mcollina (approved):
lgtm

Eomm:
The ci is failed due a `not ok 6 - Polling schemas` a bit of unluckiness

PR #741, opened 2022-02-19
add missing types for retry services options

codker:
Hello! I started looking into using Mercurius as a gateway and I noticed that the typescript interface was missing two fields that were recently added.

This PR adds `retryServicesCount` and `retryServicesInterval` in the MercuriusGatewayOptions interface.


mcollina (changes requested):
We use tsd to test our types, could you add a few assertion there?

codker:
Yeah, I will add them asap

mcollina (approved):
lgtm

PR #745, opened 2022-03-01
fix: select node when a type is duplicated

Eomm:
Fixes https://github.com/mercurius-js/mercurius-gateway/issues/31
Supersedes mercurius-js/mercurius#744 

When a Type is duplicated across multiples node, Mercurius define its service as `null`

https://github.com/mercurius-js/mercurius/blob/4549028e489e834d378acc66549b5ff2c43cf310/lib/gateway.js#L410

Instead, its value should be dynamically chosen by the node that is running the query/mutation.

This is a draft PR because there is one test failing due to the service resolution that I did not solve yet: sometimes the duplicated logic should be ignored, based on the input query.

### DIFF
--- a/lib/gateway.js
+++ b/lib/gateway.js
@@ -124,7 +124,7 @@ function defineResolvers (schema, typeToServiceMap, serviceMap, typeFieldsToServ
 
   function getService (key, priorityService) {
     const service = typeToServiceMap[key]
-    if (service || !duplicatedTypes[key]) {
+    if (service || !duplicatedTypes[key] || priorityService === false) {
       return service
     }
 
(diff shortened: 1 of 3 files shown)


Eomm:
I think the logic that must be implemented to solve the failing tests is:

given one single type `PageInfo`, the gateway's resolver function must choose to execute:

1. the simple resolver (aka `(parent) => parent[path]`)
2. the external resolution (aka calling another gateway's node)

The choice can be made based on the node's gql type, which is unknown at the moment because when the types are the same, it is merged/overwritten.

Right now, my change runs the `1.` option every time, for this reason, some tests are failing.

mcollina:
> given one single type PageInfo, the gateway's resolver function must choose to execute:

Can you link to the resolver that is being called in this case?

brandomeniconi:
Hi, we have the same issue as we are developing our new federated schema. 
The only workaround we found is to prefix the duplicated types (es: `UserPageInfo`) so they are unique and don't get a null resolver, not ideal.

We already reversed most of Mercurius federation logic code to verify the bug, but there are still some grey areas. If you can guide us in what's missing, we can allocate some dev time to help close this.

Eomm:
@brandomeniconi I think to implement this logic https://github.com/mercurius-js/mercurius/pull/745#issuecomment-1057775819

The steps are:

- when the gateway merges the same types `UserPage`, it should keep all the versions for each node
- when the gateway has a type, it should use the type defined by the node itself

PR #750, opened 2022-03-09
build: reduce dependabot update frequency

Fdawgs:
- Reduces Dependabot update frequency
- Standardizes YML formatting used to be in line with [GitHub's examples](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates) (double quotes, whitespace etc)
- Removes `ignore` from github actions updates, as discussed with @simoneb in https://github.com/fastify/github-action-merge-dependabot/pull/127 it seems it is longer need it aslong as major tags are used for actions

If happy with this PR then i will go and make the same change to the rest of the Mercurius org repos.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,16 +1,13 @@
 version: 2
 updates:
-- package-ecosystem: github-actions
-  directory: "/"
-  ignore:
-    - dependency-name: "actions/*"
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #752, opened 2022-03-09
Add backport GitHub action

jonnydgreen:
As titled!

From the updated docs: 

As soon as one opens a PR against the default branch, and the change should be backported to `v8.x`, you should add the corresponding backport label. For example, if we need to backport something to `v8.x`, we add the following label:

- `backport v8.x`

And you are done! If there are no conflicts, the action will open a separate PR with the backport for review.

If the PR can't be automatically backported, the GitHub bot will comment the failure on the PR.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -1,0 +1,16 @@
+name: Backport
+on:
+  pull_request:
+    types:
+      - closed
+      - labeled
+
+jobs:
+  backport:
+    runs-on: ubuntu-latest
+    name: Backport
+    steps:
+      - name: Backport
+        uses: tibdex/backport@v1
+        with:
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

jonnydgreen:
Thanks for sorting the labels and branch - beat me to it! :)

jonnydgreen:
Thanks for sorting the releases!

Looks like the backport worked locally within the action but failed at the push step with a 403 - hence the check failure. I can't see anything in the docs that suggest any permissions one needs to sort to get this to work but was there anything that needed to be set for the github actions bot in the fastify repo so it can push new branches?

Run ref: https://github.com/mercurius-js/mercurius/runs/5492635758?check_suite_focus=true

PR #749, opened 2022-03-08
Fix #748 Ensure FederatedError `errors` argument is an array.

tomi-bigpi:
* Add constructor argument validation to `FederatedError` to ensure `errors` is an Array
* Add input validation tests for FederatedError constructor argument
* Fix `request` and `sendRequest` to wrap caught errors in array when creating new `FederatedError`
* Add new test file for `request` to ensure array is used
* Modify existing `sendRequest` test to verify array is used
* Add support for passing in undici agent as part of the federated service configuration. Required for injection during tests. See https://github.com/nodejs/undici/issues/996

Fixes mercurius-js/mercurius#748

PR #753, opened 2022-03-09
Document new `service.agent` configuration option.

tomi-bigpi:
Adds information about the new `service.agent` option added in #749


mcollina (approved):
lgtm

Issue #751, opened 2022-03-09
Different reference types at the same path are not correctly handled by the gateway


jonnydgreen:
Howdy! I found an issue with the gateway resolver cache resolution where different reference types at the same path are not correctly handled by the gateway. This happens when we define a union list response where each union type contains different reference types at the same field path. Consider the following example as an illustration:

## Problem

Consider the following message service schema:

### Message service

```graphql
type Query @extends {
  getMessages: [Message!]!
}

union Message = EmailMessage | TextMessage

type TextMessage {
  id: ID!
  message: String!
  recipients: [TextMessageUser!]!
}

type EmailMessage {
  id: ID!
  message: String!
  header: String!
  recipients: [EmailMessageUser!]!
}

type TextMessageUser @extends @key(fields: "id") {
  id: ID! @external
}

type EmailMessageUser @extends @key(fields: "id") {
  id: ID! @external
}
```

With the resolvers:

```js
const resolvers = {
  Query: {
    getMessages: () => {
      return [
        {
          id: '1',
          message: 'Text Message.',
          recipients: [{ id: 'u1' }]
        },
        {
          id: '2',
          header: 'Email Header',
          message: 'Email Message.',
          recipients: [{ id: 'u2' }]
        }
      ]
    }
  },
  Message: {
    resolveType: (message) => {
      if ('header' in message) {
        return 'EmailMessage'
      }
      return 'TextMessage'
    }
  }
}
```

Consider the following user service schema:

### User service

```graphql
type TextMessageUser @key(fields: "id") {
  id: ID!
  name: String!
  fields: [String!]!
}

type EmailMessageUser @key(fields: "id") {
  id: ID!
  name: String!
  fields: [String!]!
}
```

With the resolvers:

```js
const resolvers = {
  TextMessageUser: {
    __resolveReference: (user) => {
      return users[user.id]
    }
  },
  EmailMessageUser: {
    __resolveReference: (user) => {
      return users[user.id]
    }
  }
}
```

### Replication

If we setup the gateway to federate the above service and we make the following query to the gateway:

```graphql
query {
  getMessages {
    ... on TextMessage {
      id
      message
      recipients {
        name
        fields
      }
    }
    ... on EmailMessage {
      id
      header
      message
      recipients {
        name
        fields
      }
    }
  }
}
```

The system should use the above resolvers to return data such that each union type recipients list is correctly populated. However, we get the following error:

```json
{
  "errors": [
    {
[...]

PR #754, opened 2022-03-09
Correctly handle reference types at the same path in the gateway

jonnydgreen:
Fixes: https://github.com/mercurius-js/mercurius/issues/751


mcollina (approved):
lgtm

PR #756, opened 2022-03-18
Keep requires directive in gateway schema

### LINKED ISSUE #729: @requires directive not found in gateway list of directives
Hi! 

My team writes the gateway schema to disk for a number of reasons. We like seeing what our federated graph schema compiles to and its nice for communicating what the schema looks like to other teams (we dont use a schema registry).

We recently started to use the `@requires` directive in a service and the gateway doesn't list it in the directives available. Because of this, `@requires` does not get defined at the top of the schema file even though its being used in the schema. This breaks our linting and also doesn't follow the GraphQL schema spec closely.

Functionally `@requires` works as expected for the federated services.

## Steps to reproduce

1. Create a new node project on your computer.

1. [Copy and paste the example gateway code](https://github.com/mercurius-js/mercurius/blob/master/examples/gateway.js) from the repo into a index.js file.

1. Run the gateway and the two services (does this by default) by running `node index.js`.

1. Query the gateway for directives

1. See that no requires directive will exist

1. Query service 2 for directives

1. See that the `@requires` directive exists

Here is the query you can run to get directives.
```graphql
query IntrospectionQuery {
  __schema {
    directives {
      name
      description
      locations
    }
  }
}
```

aleccool213:
Fixes #729

### DIFF
--- a/test/gateway/pollingInterval.js
+++ b/test/gateway/pollingInterval.js
@@ -710,16 +710,6 @@ test('Polling schemas (should properly regenerate the schema when a downstream s
   })
   t.teardown(() => clock.uninstall())
   const oldSchema = `
-    directive @extends on INTERFACE | OBJECT
-
-    directive @external on FIELD_DEFINITION | OBJECT
-
-    directive @key(fields: String!) on INTERFACE | OBJECT
-
-    directive @provides(fields: String!) on FIELD_DEFINITION
-
-    directive @requires(fields: String!) on FIELD_DEFINITION
-
     type Query {
       me: User
     }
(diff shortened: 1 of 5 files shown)


test/gateway/pollingInterval.js
```diff
@@ -710,16 +710,6 @@ test('Polling schemas (should properly regenerate the schema when a downstream s
   })
   t.teardown(() => clock.uninstall())
   const oldSchema = `
-    directive @extends on INTERFACE | OBJECT
```
aleccool213, line 713: It doesn't make sense for a service to have this defined in its schema. This is federation metadata.

test/gateway/pollingInterval.js
```diff
@@ -804,16 +794,6 @@ test('Polling schemas (should properly regenerate the schema when a downstream s
   })
 
   const refreshedSchema = `
-    directive @extends on INTERFACE | OBJECT
```
aleccool213, line 807: It doesn't make sense for a service to have this defined in its schema. This is federation metadata.

mcollina (approved):
lgtm, good spot!

mcollina:
@jonnydgreen could you take a look as well?

jonnydgreen:
Since it's a bugfix, I'm thinking we should backport this to v8 as well - what do you think @mcollina?

mcollina:
Great idea, let's try the action.

PR #773, opened 2022-04-15
Added async support in rewriteHeaders

all2pie:


mcollina (approved):
lgtm

index.d.ts
```diff
@@ -384,8 +384,13 @@ export interface MercuriusGatewayService {
   schema?: string;
   wsUrl?: string;
   mandatory?: boolean;
-  initHeaders?: (() => OutgoingHttpHeaders | Promise<OutgoingHttpHeaders>) | OutgoingHttpHeaders;
-  rewriteHeaders?: <TContext extends MercuriusContext = MercuriusContext>(headers: IncomingHttpHeaders, context: TContext) => OutgoingHttpHeaders;
+  initHeaders?:
+    | (() => OutgoingHttpHeaders | Promise<OutgoingHttpHeaders>)
+    | OutgoingHttpHeaders;
+  rewriteHeaders?: <TContext extends MercuriusContext = MercuriusContext>(
```
jonnydgreen, line 390: could we add a type test for this? See here for an example of how to do this: https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts#L276

jonnydgreen (commented):
This is great - will really help for things like setting up a gateway with auth protected federated services for example :)

index.d.ts
all2pie: I don't have any experience with testing, I checked out the `tsd` library and it seems simple but I don't know how to use it in proper context to test the `rewriteHeaders` type.

jonnydgreen: No problem! Basically, you just need to write out what you would expect to write in typescript in that test file that I linked, with some extra type tests inside the function to verify that everything is of the correct type. Then, if running tsd passes (In mercurius, I think it's `npm run typescript` off the top of my head), you're all good.

As an example, in the link that I mentioned, this would be something like the below to prove that we can define what we expect to be able to define according to the types. This ensures that the types are working as expected from a TypeScript point of view:

```ts
gateway.register(mercurius, {
  gateway: {
    services: [
      {
        name: 'user',
        async rewriteHeaders: (headers, context) => {
          expectAssignable<MercuriusContext>(context)
          return {
            authorization: headers.authorization
          }
        }
      }
    ]
  }
})
```

Here is an example commit that adds testing to the types: https://github.com/mercurius-js/mercurius/commit/7ec10306b2d6261d26986da02893b3c4fe459687 . Let me know if you have any more questions :)

all2pie: Thanks, please check new commit.

mcollina (approved):
lgtm

jonnydgreen (approved):
LGTM

Issue #725, opened 2022-02-06
Reduce package dependencies


p-kuen:
I installed mercurius the first time today and I recognized it has a lot of dependencies which could be reduced quite easily in my opinion:

#### [promise.allsettled](https://www.npmjs.com/package/promise.allsettled):
As of NodeJS 12.9 and up Promise.allsettled is available natively without a package dependency. If there would be a node minimal version of 12.9 set, we could remove this dependency (especially because this package has a lot of polyfill dependencies too)

#### [events.on](https://www.npmjs.com/package/events.on):
As of NodeJS 12.16 and up events.on is available natively without a package dependency. If there would be a node minimal version of 12.16 set, we could remove this dependency

#### [ws](https://www.npmjs.com/package/ws) and [fastify-websocket](https://www.npmjs.com/package/fastify-websocket)
It is sad that everyone who does not use the websocket feature has to install these dependencies. In my opinion it would be better if this feature is a plugin for mercurius.

#### [graphql-jit](https://www.npmjs.com/package/graphql-jit)
In my case, I am using code-first graphql (no gql language used). Therefore I don't need graphql-jit, yet I have the dependency installed and even get a peer dependency error because I am using graphql >= 16. I would recommend using graphql-jit as a plugin.

#### [readable-stream](https://www.npmjs.com/package/readable-stream)
I do not exactly know as of which version of NodeJS readableStreams are supported, but I am sure, most of the users of mercurius do not need this polyfill.

#### [undici](https://www.npmjs.com/package/undici)
In the code I identified undici to be used for gateways. It would be perfect, if this was a plugin for mercurius too as I think most users do not use this feature.

I would be glad to help reducing dependencies of this package. :)

mcollina:
If you would like to send PRs for promise.allsettled and events.on, it would be amazing.

As for graphql-jit, it seems it is not clear to you how it works. It will improve the performance of your queries independently on how you generate the schema. The peerDep warning does not affect the actual operations and it will soon be fixed. It's great to keep it where it is.

readable-stream is not removable as it will be installed by other dependencies too.

As for the gateway and websockets, they are not that easy to remove and it would severely complicate the maintenance of this module. Extracting them as plugins is possible - but we would have to shift this repo to a monorepo as they are fundamental part of this module. If you'd like the challenge, I'll be happy to review a PR.

p-kuen:
I already created a PR for promise.allsettled, events.on would be my next try. Will do that soon!

[...]

PR #774, opened 2022-04-16
refactor!: replace promise.allsettled dependency with node implementation
BREAKING CHANGE: drop support for node <12.9.0

Refs: #725

p-kuen:
This PR replaces the `promise.allsettled` dependency with the node internal function.
This is only available since node version 12.9.0, so this PR also adds engine constraint for node in package.json.


mcollina (approved):
lgtm, it's not breaking anyway, we support only the latest v12.

Issue #771, opened 2022-04-11
Graphiql cannot be used with helmet on fresh browsers


capaj:
ATM graphiql cannot be used with helmet. This config has worked for me in the past, but I suspect new browser versions broke it:

```ts
app.register(fastifyCors)
  const trustedDomains = ['https://unpkg.com']

  app.register(fastifyHelmet, {
    // TODO figure this out
    contentSecurityPolicy: {
      directives: {
        defaultSrc: ["'self'", "'unsafe-inline'", "'unsafe-eval'"].concat(
          trustedDomains
        ),

        scriptSrc: ["'self'", "'unsafe-inline'", "'unsafe-eval'"].concat(
          trustedDomains
        ),
        styleSrc: ["'self'", "'unsafe-inline'", "'unsafe-eval'"].concat(
          trustedDomains
        )
      }
    }
  })
```

and with the newest chrome(100)/firefox(99) I started getting:

```
main.js:11          GET https://unpkg.com/subscriptions-transport-ws@0.9.19/browser/client.js net::ERR_BLOCKED_BY_RESPONSE.NotSameOriginAfterDefaultedToSameOriginByCoep 200
(anonymous) @ main.js:11
url @ main.js:5
urls @ main.js:15
importDependencies @ main.js:50
Promise.then (async)
(anonymous) @ main.js:62
main.js:48          GET https://unpkg.com/graphiql@1.4.2/graphiql.css net::ERR_BLOCKED_BY_RESPONSE.NotSameOriginAfterDefaultedToSameOriginByCoep 200
importDependencies @ main.js:48
Promise.then (async)
(anonymous) @ main.js:62
main.js:11          GET https://unpkg.com/graphiql@1.4.2/graphiql.min.js net::ERR_BLOCKED_BY_RESPONSE.NotSameOriginAfterDefaultedToSameOriginByCoep 200
```

But it worked a few chrome versions back for sure.
As a workaround I won't use helmet for now, but it certainly isn't ideal.

Not sure how to fix it-maybe as a workaround serve those packages from node_modules?

mcollina:
I would still prefer to keep serving them from unpkg so that we avoid their weight on the package and avoid a build step here. Anyway, everything is up for grabs.

mcollina:
When was this changed? Because things where working for me two weeks ago.

conradthegray:
I'm looking into this. I have recreated the issue.

conradthegray:
[...]

PR #775, opened 2022-04-22
fix: allow cross origin loading for graphiql

conradthegray:
Fixes https://github.com/mercurius-js/mercurius/issues/771

unpkg is not setting up correct cross origin headers causing browsers to block scripts and styles loaded from them. The proposed change tells the browser to load the content from unpkg.

**How to test the fix**

1. Create a test server ([gist here](https://gist.github.com/conradthegray/16751dd0870376e366d82280b92da265))
2. Run the test server and navigate to `http://localhost:3000/graphiql`
3. You should see a blank page without the fix applied
4. Apply the fix in `mercurius-771/node_modules/mercurius/static/main.js` file
5. Reload the test server and navigate to `http://localhost:3000/graphiql`
6. You should see GraphiQL interface

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -6,6 +6,7 @@ const importer = {
       const script = document.createElement('script')
       script.type = 'text/javascript'
       script.src = url
+      script.crossOrigin = 'anonymous'
       script.addEventListener('load', () => resolve(script), false)
       script.addEventListener('error', (err) => reject(err), false)
       document.body.appendChild(script)
@@ -45,6 +46,7 @@ function importDependencies () {
   link.type = 'text/css'
   link.rel = 'stylesheet'
   link.media = 'screen,print'
+  link.crossOrigin = 'anonymous'
   document.getElementsByTagName('head')[0].appendChild(link)
 
   return importer.urls([


static/main.js
```diff
@@ -6,6 +6,7 @@ const importer = {
       const script = document.createElement('script')
       script.type = 'text/javascript'
       script.src = url
+      script.crossOrigin = 'anonymous'
```
kurtmilam, line 9: Minor point, but the actual name of the property is 'crossorigin' (all lower case).

I know the camel case version works, and I know that html attribute names are case-insensitive, but I think I'd prefer to use the casing indicated by the spec.

simoneb:
@conradthegray can you please link the issue properly in the PR description?

static/main.js
conradthegray: According to MDN, the proper name is `crossOrigin` when we refer to it from JS https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement
The fix won't work if the casing is changed to all lower case.

simoneb (approved):
LGTM

jonnydgreen (approved):
LGTM

simoneb:
Build is failing but I don't see anything in this PR which may cause that. On the other hand the build is green on master and it doesn't have a history of failing, so I'm confused what's going on

simoneb:
I'll go ahead and merge this. If the failure occurs on master, we'll look into it straight away

jonnydgreen:
yeah that's very strange, I also can't see how this change would affect the tests. Looking at the file in question, it seems like this was initially observed when creating the relevant PR but the failures seemed to be fixed in later commits: https://github.com/mercurius-js/mercurius/pull/706#issuecomment-1023191531 . To me, this suggests that it is intermittent so maybe a rerun of the jobs will be okay - like you mentioned, if it happens on master, we can take a closer look!

simoneb:
master build is green 🤷

PR #780, opened 2022-04-28
chore: update fastify deps to @fastify

simone-sanfratello:
as titled


jonnydgreen (approved):
LGTM - thoughts on semver upgrade for this?

mcollina:
patch

PR #789, opened 2022-05-03
fix: gateway configuration validation

### LINKED ISSUE #787: bug: in gateway, name for services must be mandatory
as titled, otherwise only the last service is registered

if we want to leave it optional, we must change [here](https://github.com/mercurius-js/mercurius/blob/master/lib/gateway.js#L391) adding something like `service || i` (where `i` is the counter of the loop)

example

```js
app.register(mercurius, {
  gateway: { services: [
    { url: `http://localhost:3001/graphql` },
    { url: `http://localhost:3002/graphql` }
  ] },
  jit: 1
})
```

temsa:
Fixes #787 

Checks that the `services` passed to `gateway` option are matching what is documented (in the docs and in TS types) and used so far:
1. `services` should have a unique (String) `name` 
2. `services` should have one or multiple (as an Array) `url` that is of type `String`

Given the current `gateway` `services` option is really permissive and this PR restricts it, this may be considered as a BREAKING CHANGE, but given TS types explicitly restrains the options to what is implemented by this PR, we might want not to consider it breaking... WDYT ?

### DIFF
--- a/index.js
+++ b/index.js
@@ -151,6 +151,25 @@ const plugin = fp(async function (app, opts) {
     throw new MER_ERR_INVALID_OPTS('Adding "schema", "resolvers" or "loaders" to plugin options when plugin is running in gateway mode is not allowed')
   }
 
+  if (gateway && Array.isArray(gateway.services)) {
+    const serviceNames = new Set()
+    for (const service of gateway.services) {
+      if (typeof service !== 'object') {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
I think it's good!

Issue #790, opened 2022-05-04
¿Is Subscription a reserved type Name?


fjmoreno:
Hi, ¿Is Subscription a reserved type Name? i expect to redefine Subscription rootType name as stated on graphql specs:  https://spec.graphql.org/October2021/#sec-Root-Operation-Types

### Steps to reproduce

**Given following example:**
```
'use strict'

const Fastify = require('fastify')
const mercurius = require('..')

const app = Fastify()

const Series = [{
  name: 'Serie1',
  subPids: ['Sub1','Sub2', 'Sub3']
}, {
  name: 'Serie2',
  subPids: ['Sub1','Sub2']
}, {
  name: 'Serie3',
  subPids: ['Sub1']
}]

const Subscriptions = [
  {
    Pid: 'Sub1',
    Name: 'Subscription1',
  }, {
    Pid: 'Sub2',
    Name: 'Subscription1',
  }, {
    Pid: 'Sub3',
    Name: 'Subscription1',
  },
]

const schema = `
  type Serie {
    name: String!
    Subscriptions: [Subscription]
  }

  type Subscription {
    Pid: ID!
    ContentType: String!
  }

  type Query {
    Series: [Serie]
  }
  
`

const resolvers = {
  Query: {
    Series (_, params, { reply }) {
      return Series
    }
  },
  Subscription: {
    ContentType: () => 'SUB',
  },
  Serie: {
    Subscriptions (parent, params, { reply }) {
      return Subscriptions.filter( (e) => parent.subPids.includes(e.Pid))
    }
  }
}

app.register(mercurius, {
  schema,
  resolvers,
  graphiql: true
})

app.listen(3000)

```

**I expect following response:** 
```
{
  "data": {
    "Series": [
      {
        "name": "Serie1",
        "Subscriptions": [
          {
            "Pid": "Sub1",
            "ContentType": "SUB"
          },
          {
            "Pid": "Sub2",
            "ContentType": "SUB"
          },
          {
            "Pid": "Sub3",
            "ContentType": "SUB"
          }
        ]
      },
      {
        "name": "Serie2",
        "Subscriptions": [
          {
            "Pid": "Sub1",
            "ContentType": "SUB"
          },
          {
            "Pid": "Sub2",
            "ContentType": "SUB"
          }
        ]
      },
      {
        "name": "Serie3",
        "Subscriptions": [
          {
            "Pid": "Sub1",
            "ContentType": "SUB"
          }
        ]
      }
    ]
[...]

PR #800, opened 2022-05-22
Subscription root types

mattcan:
This is a set of fixes for #790.

---

There are two changes here:

* f10aa84 makes a change so that a redefined `subscription` type is correctly identified
* f839b62 ignores the custom parsing when the API option `subscription` is `false` or `undefined`

I packed these changes into one PR as they are affecting the same lines of code but they are really separate changes. Happy to split them into separate PRs if that is easier to deal with.

When I was reviewing the code in Gateway mode, it seemed like the best case is to not make the same changes there. My reasoning:
* in Gateway mode, the schema isn't adjustable
* while `Subscription` does not need to be used as a root type, it has become de facto standard and it is far more likely that other services will change their type


mcollina (approved):
lgtm

jonnydgreen (approved):
LGTM

simoneb:
shall this close #790?

PR #803, opened 2022-05-31
fix: check null lruGatewayResolvers when cache option is disabled 

atjhoendz:
since we cannot pass function in validationRules options because it's incompatible with query caching [#L73](https://github.com/mercurius-js/mercurius/blob/117662dad9f0d7977acfdb04d88d349d8e87f8e0/index.js#L73), so I set the cache option to false, but I got an error like this `Cannot read properties of null (reading 'get')` 

when I check the code there is no null check in this line [#L407](https://github.com/mercurius-js/mercurius/blob/117662dad9f0d7977acfdb04d88d349d8e87f8e0/lib/gateway/make-resolver.js#L407), 

so I proposed this MR to fix the issue above, please kindly check this TIA 🙏


mcollina (approved):
lgtm

PR #805, opened 2022-05-31
Fix for cached jit query is not a function

### LINKED ISSUE #799: cached.jit.query is not a function
i have faced this error
`{ "data": null, "errors": [ { "message": "cached.jit.query is not a function" } ] }`

it only happens when jit is enabled and i query with input. queries without input pass.

this query throws cached.jit.query is not a function.
`query findManySpecialty($where: SpecialtyWhereInput) { findManySpecialty(where: $where) { id } }`

this query work fine.
`query findManySpecialty { findManySpecialty { id } } `

i don't fully understand the problem here, but `isCompiledQuery(cached.jit)` fixed it - from this [commit](https://github.com/mercurius-js/mercurius/pull/174/commits/be88b7b7dcce886e238dde0e8496c2078549f7a4)

GeertWille:
Fixes #799

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
 const fp = require('fastify-plugin')
 const LRU = require('tiny-lru')
 const routes = require('./lib/routes')
-const { compileQuery } = require('graphql-jit')
+const { compileQuery, isCompiledQuery } = require('graphql-jit')
 const { Factory } = require('single-user-cache')
 const {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm, great job!

Gdewilde:
Awesome job, Geert!

Issue #777, opened 2022-04-23
Any way to forward headers from a downstream service in federation mode?


reezpatel:
I want to send `Set-Cookie` from one of the federated service back to user? So, I want to know, right now, is there any way to achieve this?

mcollina:
I don't think so, but it would be great to add a callback to do that.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

harshitkumar31:
Started a WIP PR: https://github.com/mercurius-js/mercurius/pull/820

flyweight:
I've been looking at #820, and it doesn't seem to cover the following case:

Imagine there there are two subgraphs, `authentication` and `posts`, behind a gateway. The `authentication` subgraph has an `authenticate` mutation that sets a cookie on success.  It seems that `set-cookie` header gets eaten by the gateway at the moment. Is it possible to passthrough specific headers when running in federation mode?

PR #820, opened 2022-07-03
Set Headers from subgraphs

harshitkumar31:
Addresses https://github.com/mercurius-js/mercurius/issues/777

Letting subgraphs/downstream services set headers in federation mode

### DIFF
--- a/examples/gateway.js
+++ b/examples/gateway.js
@@ -174,7 +174,10 @@ async function start () {
     gateway: {
       services: [{
         name: 'user',
-        url: 'http://localhost:4001/graphql'
+        url: 'http://localhost:4001/graphql',
+        setResponseHeaders: (reply) => {
+          reply.header('abc', 'abc')
+        }
       }, {
         name: 'post',
         url: 'http://localhost:4002/graphql'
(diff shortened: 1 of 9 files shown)


mcollina (changes requested):
Good work!

Could you add a test for the type? We used http://npm.im/tsd.

Could you add docs for the new option?

examples/gateway.js
```diff
@@ -174,7 +174,10 @@ async function start () {
     gateway: {
       services: [{
         name: 'user',
-        url: 'http://localhost:4001/graphql'
+        url: 'http://localhost:4001/graphql',
+        setResponseHeaders: (reply) => {
```
jonnydgreen, line 178: wondering if this would be better as a service-level hook? That would provide us with a lot more flexibility going forward! WDYT? See fastify per route hooks for an example of what I mean: https://www.fastify.io/docs/latest/Reference/Hooks/#route-level-hooks

harshitkumar31: @jonnydgreen ,
If we define it as a service-level hook, we'd need some reconciliation logic in gateway which combines headers sent from multiple subgraphs.
That would be ideal, but I'm not sure where to make that change. Would you guys be able to guide me on where I can write that logic.

jonnydgreen: The idea I had in my was to make it independent of setting headers and leave the reconciliation logic up to the user (which would also fit your use case). E.g. as a pseudo-code example, one could define a service as follows:

```js
{
  name: 'user',
  onFederatedServiceResponse(gatewayRequest, gatewayReply, serviceDefinition, serviceRequest, serviceReply) {
    // Service specific logic defined by user here
  }
}
```

Looking at the Mercurius codebase, I think this probably requires work that is out of scope of this PR and probably makes sense to do separately (if you're interested, I'd be very happy to guide you with this) so I'm completely fine with your initial approach :)

jonnydgreen: Aces :)

jonnydgreen (approved):
LGTM!

mcollina (approved):
lgtm

PR #822, opened 2022-07-04
POC: passing extensions from subgraph

harshitkumar31:


marco-ippolito:
this has already implemented in  [mercurius-gateway](https://github.com/mercurius-js/mercurius-gateway), you can use [extensions collectors](https://github.com/mercurius-js/mercurius-gateway#collectors)

PR #829, opened 2022-07-23
Fix LRU require for module bundlers

tobythomassen:
When using Mercurius with a module bundler such as esbuild or webpack, running will output `TypeError: LRU is not a function`.

Not sure what originates the error, but it has already been fixed in fastify (see https://github.com/fastify/fastify/pull/2918).

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const fp = require('fastify-plugin')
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

tetchel:
can/will this be backported to v8 branch for those of us still on fastify 3?

PR #830, opened 2022-07-26
Throw GraphQL Schema error when federated schemas are not extend each other properly

### LINKED ISSUE #826: federation gateway should throw on type's conflicts
Given the following schemas:

### Node 1

```gql
    extend type Query {
      getOrder: Order
    }

    type Order @key(fields: "id") {
      id: ID!
      name: String
    }
```

### Node 2

```
  type Order @key(fields: "id") {
    id: ID!
    surname: String
  }
```

The gateway starts correctly instead of throwing.

This is a clear developer's error, as the `Node 2`'s schema should use the `extends` keyword:

```gql
  extends type Order {
    surname: String
  }
```

In this case, Mercurius should throw an error when:

- there is a naming conflict without the `extends` keyword
- the `extends` entity must not use the `ID!` property value

In both the above cases, Apollo server trigger this error:

```
This data graph is missing a valid configuration. Field "Order.id" can only be defined once.

There can be only one type named "Order".
```

This would help developers to spot these schema conflicts at first gateway run.

JamesLMilner:
Aims to resolve #826 

This is currently an experimental draft PR as I get an understanding of what the desired behaviours are for merging federated schema types. The overall idea is that if a schema comes together and non value types do not correctly extend each other within that, then it should throw a GraphQL schema validation error.

I'm pretty certain the current code doesn't do exhaustive checks on ensuring entity types are identical when they haven't been extended - but hopefully at least it gives some ideas on how that might work. I know in another part of the code base for value types there is also a TODO: for ensuring equality between value types.

One other aspect I am not totally sure on is if the order of extension matters, for example if the schema extends a type before it has been defined, but it is defined later in the schema.

### DIFF
--- a/test/gateway/schema.js
+++ b/test/gateway/schema.js
@@ -202,7 +202,7 @@ test('It builds the gateway schema correctly', async (t) => {
     content
     ...AuthorFragment
   }
-  
+
   fragment AuthorFragment on Post {
     author {
       ...UserFragment
@@ -1211,7 +1211,7 @@ test('Should handle union with InlineFragment', async (t) => {
       ...ShelveInfos
     }
   }
-  
+
   fragment ShelveInfos on Shelve {
     id
     products {
@@ -1428,7 +1428,7 @@ test('Should handle interface', async (t) => {
       ...ShelveInfos
     }
   }
-  
+
   fragment ShelveInfos on Shelve {
     id
     products {
@@ -2167,7 +2167,7 @@ test('Uses the supplied schema for federation rather than fetching it remotely',
     content
     ...AuthorFragment
   }
-  
+
   fragment AuthorFragment on Post {
     author {
       ...UserFragment
@@ -2655,7 +2655,7 @@ test('It builds the gateway schema correctly with two services query extension h
     content
     ...AuthorFragment
   }
-  
+
   fragment AuthorFragment on Post {
     author {
       ...UserFragment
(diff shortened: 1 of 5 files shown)


test/gateway/schema.js
```diff
@@ -2754,3 +2754,420 @@ test('It builds the gateway schema correctly with two services query extension h
...
+  })
+
+  const gateway = Fastify()
+  t.teardown(async () => {
+    await gateway.close()
+    await orderService.close()
+    await conflictingService.close()
+  })
+
+  t.rejects(() => gateway.register(GQL, {
```
Eomm, line 3154: Could we check the error message?

Here an example from apollo:

```
This data graph is missing a valid configuration. [accounts] Order.name -> Field "Order.name" already exists in the schema. It cannot also be defined in this type extension. If this is meant to be an external field, add the `@external` directive.
```

test/gateway/schema.js
```diff
@@ -2754,3 +2754,420 @@ test('It builds the gateway schema correctly with two services query extension h
...
+        }
+      }, {
+        name: 'conflicting-order',
+        url: `http://localhost:${conflictingServicePort}/graphql`
+      }]
+    }
+  })
+})
+
+test('Types with duplicate names that are value types should not throw', async (t) => {
```
Eomm, line 2964: With this test, Apollo throws:

```
This data graph is missing a valid configuration. [accounts] Order -> Value types cannot be entities (using the `@key` directive). Please ensure that the `Order` type is extended properly or remove the `@key` directive if this is not an entity.
```

If I remove the `@key` all works

PR #846, opened 2022-08-23
Always teardown subscriptions on completion

fathyb:
Always clear `SubscriptionConnection.subscriptionContexts` and `SubscriptionConnection.subscriptionIters` after a subscription completes, instead of waiting for the client to send a `complete` message.

This prevents buggy clients from leaking memory on the server.


mcollina (commented):

mcollina (approved):
lgtm

PR #515, opened 2021-06-17
fix: change context merging behavior

DanTsk:
Use provided `context` as a target in merging to keep the object prototype.
Resolves: #514

PR #849, opened 2022-08-26
feat: uniquely identify batched queries

### LINKED ISSUE #842: Uniquely identify batched queries through all their lifecycle
While trying to give support to https://github.com/nearform/mercurius-apollo-tracing/issues/125 we noticed that we were unable to uniquely identify each operation definition contained in batched queries through their whole lifecycle. This is needed in order to know which tracer to start or stop at any given time.

We tried two approaches and faced two different issues with them:

1. Using the `document` object contained in the hooks to extract the relevant operation information. This simply can't be done since the `onResolution` hook doesn't receive the `document` object, only the `execution` and `context` ones. See https://github.com/mercurius-js/mercurius/blob/master/docs/hooks.md#onresolution

Also, in the specific case of mercurius-apollo-tracing, it hooks into schema resolvers that don't provide all the `document` details (and they don't need to), just some `info` object. So it's best to go with the next approach...

2. Using the `context.__currentQuery` attribute. Turns out that this attribute is only updated during the `preParsing` stage and left untouched afterwards, so it's basically useless for batched queries during the rest of the lifecycle.

For reproducing this we can setup a basic Fastify + Mercurius server with `allowBatchedQueries: true` and a typical `add` query, and include these debugs on all hooks:
```
app.graphql.addHook('preParsing', async (_, __, context) => {
  app.log.info(`preParsing: ${context.__currentQuery}`)
})

app.graphql.addHook('preValidation', async (_, __, context) => {
  app.log.info(`preValidation: ${context.__currentQuery}`)
})

app.graphql.addHook('preExecution', async (_, __, context) => {
  app.log.info(`preExecution: ${context.__currentQuery}`)
})

app.graphql.addHook('onResolution', async (__, context) => {
  app.log.info(`onResolution: ${context.__currentQuery}`)
})
```

Then fire the following queries from a client that supports batched queries:
```
query { add(x: 1, y: 1) }
query { add(x: 2, y: 2) }
```

One would expect to have each hook print out the two queries, but as we can see only the `preParsing` hook does that while the rest of them print out the second query twice:
```
{"level":30,"time":1660745778776,"pid":74281,"hostname":"BBaidal-SPMAC21","reqId":"req-2","req":{"method":"POST","url":"/graphql","hostname":"localhost:3000","remoteAddress":"::1","remotePort":50784},"msg":"incoming request"}
{"level":30,"time":1660745778780,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"preParsing: {\n  add(x: 1, y: 1)\n}"}
{"level":30,"time":1660745778781,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"preParsing: {\n  add(x: 2, y: 2)\n}"}
{"level":30,"time":1660745778781,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"preValidation: {\n  add(x: 2, y: 2)\n}"}
{"level":30,"time":1660745778782,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"preValidation: {\n  add(x: 2, y: 2)\n}"}
{"level":30,"time":1660745778786,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"preExecution: {\n  add(x: 2, y: 2)\n}"}
{"level":30,"time":1660745778787,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"preExecution: {\n  add(x: 2, y: 2)\n}"}
{"level":30,"time":1660745778788,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"onResolution: {\n  add(x: 2, y: 2)\n}"}
{"level":30,"time":1660745778788,"pid":74281,"hostname":"BBaidal-SPMAC21","msg":"onResolution: {\n  add(x: 2, y: 2)\n}"}
{"level":30,"time":1660745778789,"pid":74281,"hostname":"BBaidal-SPMAC21","reqId":"req-2","res":{"statusCode":200},"responseTime":12.858789920806885,"msg":"request completed"}
```

I'm sure that not only mercurius-apollo-tracing would benefit from having a way to uniquely identify batched queries through all their lifecycle, but also other potential plugins that could handle batched queries.

drakhart:
resolves #842

Fixes `context.__currentQuery` being overwritten by subsequent batched query operations.
Stores `context.operationsCount` and `context.operationId` (same naming as with subscriptions) to uniquely identify batched query operations at a later stage.

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -220,16 +220,11 @@ module.exports = async function (app, opts) {
       return new MER_ERR_GQL_PERSISTED_QUERY_NOT_FOUND('Unknown query')
     }
 
-    // Create individual contexts for multiple operations, otherwise reference the original context
-    const context = request[kRequestContext].operationsCount > 1
-      ? Object.create(request[kRequestContext])
-      : request[kRequestContext]
-
     // Handle the query, throwing an error if required
     return reply.graphql(
       query,
       Object.assign(
-        context,
+        request[kRequestContext],
         { pubsub: subscriber, __currentQuery: query }
       ),
       variables,
@@ -357,13 +352,18 @@ module.exports = async function (app, opts) {
 
     if (allowBatchedQueries && Array.isArray(request.body)) {
       // Batched query
-      Object.assign(request[kRequestContext], { operationsCount: request.body.length })
+      const operationsCount = request.body.length
+
+      Object.assign(request[kRequestContext], { operationsCount })
 
       return Promise.all(request.body.map(async (r, operationId) => {
-        Object.assign(request[kRequestContext], { operationId })
+        // Create individual reqs for multiple operations, otherwise reference the original req
+        const operationReq = operationsCount > 1 ? Object.create(request) : request
+
+        Object.assign(operationReq[kRequestContext], { operationId })
 
         try {
-          return await execute(r, request, reply)
+          return await execute(r, operationReq, reply)
         } catch (e) {
           const { response } = errorFormatter(e, request[kRequestContext])
           return response
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -220,11 +220,16 @@ module.exports = async function (app, opts) {
       return new MER_ERR_GQL_PERSISTED_QUERY_NOT_FOUND('Unknown query')
     }
 
+    // Create individual contexts for multiple operations, otherwise reference the original context
+    const context = request[kRequestContext].operationsCount > 1
+      ? Object.create(request[kRequestContext])
+      : request[kRequestContext]
+
     // Handle the query, throwing an error if required
     return reply.graphql(
       query,
       Object.assign(
-        request[kRequestContext],
+        context,
         { pubsub: subscriber, __currentQuery: query }
       ),
       variables,
@@ -352,13 +357,18 @@ module.exports = async function (app, opts) {
 
     if (allowBatchedQueries && Array.isArray(request.body)) {
       // Batched query
-      return Promise.all(request.body.map(r =>
-        execute(r, request, reply)
-          .catch(e => {
-            const { response } = errorFormatter(e, request[kRequestContext])
-            return response
-          })
-      ))
+      Object.assign(request[kRequestContext], { operationsCount: request.body.length })
+
+      return Promise.all(request.body.map(async (r, operationId) => {
+        Object.assign(request[kRequestContext], { operationId })
+
+        try {
+          return await execute(r, request, reply)
+        } catch (e) {
+          const { response } = errorFormatter(e, request[kRequestContext])
+          return response
+        }
+      }))
     } else {
       // Regular query
       return execute(request.body, request, reply)
--- a/test/batched.js
+++ b/test/batched.js
@@ -2,6 +2,7 @@
 
 const { test } = require('tap')
 const Fastify = require('fastify')
+const sinon = require('sinon')
 const GQL = require('..')
 
 test('POST regular query', async (t) => {
@@ -345,4 +346,45 @@ test('POST batched query with a resolver which succeeds and a resolver which thr
   })
 
   t.same(JSON.parse(res.body), [{ data: { add: 3 } }, { data: null, errors: [{ message: 'Internal Server Error' }] }])
+})
+
+test('POST batched query has an individual context for each operation', async (t) => {
+  const app = Fastify()
+
+  const contextSpy = sinon.spy()
+
+  const schema = `
+      type Query {
+        test: String
+      }
+    `
+
+  const resolvers = {
+    test: (_, ctx) => contextSpy(ctx.operationId, ctx.operationsCount, ctx.__currentQuery)
+  }
+
+  app.register(GQL, {
+    schema,
+    resolvers,
+    allowBatchedQueries: true
+  })
+
+  await app.inject({
+    method: 'POST',
+    url: '/graphql',
+    body: [
+      {
+        operationName: 'TestQuery',
+        query: 'query TestQuery { test }'
+      },
+      {
+        operationName: 'DoubleQuery',
+        query: 'query DoubleQuery { test }'
+      }
+    ]
+  })
+
+  t.ok(contextSpy.calledTwice)
+  t.ok(contextSpy.calledWith(0, 2, sinon.match(/TestQuery/)))
+  t.ok(contextSpy.calledWith(1, 2, sinon.match(/DoubleQuery/)))
 })
--- a/test/batched.js
+++ b/test/batched.js
@@ -384,7 +384,7 @@ test('POST batched query has an individual context for each operation', async (t
     ]
   })
 
-  t.ok(contextSpy.calledTwice)
-  t.ok(contextSpy.calledWith(0, 2, sinon.match(/TestQuery/)))
-  t.ok(contextSpy.calledWith(1, 2, sinon.match(/DoubleQuery/)))
+  sinon.assert.calledTwice(contextSpy)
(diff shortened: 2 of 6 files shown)


lib/routes.js
```diff
@@ -220,11 +220,16 @@ module.exports = async function (app, opts) {
       return new MER_ERR_GQL_PERSISTED_QUERY_NOT_FOUND('Unknown query')
     }
 
+    // Create individual contexts for multiple operations, otherwise reference the original context
+    const context = request[kRequestContext].operationsCount > 1
+      ? Object.create(request[kRequestContext])
```
drakhart, line 225: `Object.create` is preferred over `Object.assign` in order to preserve the original class and constructor, since Mercurius accepts custom context classes.

Check https://github.com/mercurius-js/mercurius/pull/515 for more information.

mcollina (approved):
lgtm, good work!

jonnydgreen (commented):
Functionally looks great! Should we update the TypeScript typings (with type tests for these) for the context as well?

RafaelGSS (approved):

drakhart:
> Functionally looks great! Should we update the TypeScript typings (with type tests for these) for the context as well?

I've added the new (and missing) context properties, but I'm unsure how to test these types: never wrote type tests and by looking at the existing ones it's unclear to me how to test this specific ones. Perhaps you can lend a hand? 🙏

simoneb:
> > Functionally looks great! Should we update the TypeScript typings (with type tests for these) for the context as well?
> 
> I've added the new (and missing) context properties, but I'm unsure how to test these types: never wrote type tests and by looking at the existing ones it's unclear to me how to test this specific ones. Perhaps you can lend a hand? 🙏

Check this out @drakhart https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts

lib/routes.js
simoneb: Just checking, there is no chance that request[kRequestContext] is null or undefined here?

drakhart:
> > > Functionally looks great! Should we update the TypeScript typings (with type tests for these) for the context as well?
> > 
> > 
> > I've added the new (and missing) context properties, but I'm unsure how to test these types: never wrote type tests and by looking at the existing ones it's unclear to me how to test this specific ones. Perhaps you can lend a hand? 🙏
> 
> Check this out @drakhart https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts

Yes yes, I did, that's what I meant with "by looking at the existing ones it's unclear to me how to test this specific ones". I'll try my best, no promises made.

lib/routes.js
```diff
@@ -352,13 +357,18 @@ module.exports = async function (app, opts) {
...
-        execute(r, request, reply)
-          .catch(e => {
-            const { response } = errorFormatter(e, request[kRequestContext])
-            return response
-          })
-      ))
+      Object.assign(request[kRequestContext], { operationsCount: request.body.length })
+
+      return Promise.all(request.body.map(async (r, operationId) => {
+        Object.assign(request[kRequestContext], { operationId })
```
simoneb, line 363: ⚠️ this is a race condition

simoneb (changes requested):

lib/routes.js
RafaelGSS: I don't think so. Promises are synchronous until reaching an async io operation. In this case `execute()`.

drakhart: I don't think `request[kRequestContext]` will ever be null or undefined at that point, it gets created [right here](https://github.com/drakhart/mercurius/blob/c31bee092d344e458c3f6621147d2b2b5cf1b389/lib/routes.js#L348) before executing the query operation a few lines below.

drakhart: And yes, that's certainly a race condition due to the `Promise.all`. Looking into it.

jonnydgreen:
> Yes yes, I did, that's what I meant with "by looking at the existing ones it's unclear to me how to test this specific ones". I'll try my best, no promises made.

You probably want a test similar in structure to this one where the parameter types are checked: https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts#L748

Happy to help if you have any further questions :)

drakhart:
> > Yes yes, I did, that's what I meant with "by looking at the existing ones it's unclear to me how to test this specific ones". I'll try my best, no promises made.
> 
> You probably want a test similar in structure to this one where the parameter types are checked: https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts#L748
> 
> Happy to help if you have any further questions :)

I think it should be fine now, but I'm not sure because it felt too easy in the end. Please have a look and let me know.

lib/routes.js
drakhart: Race condition should be solved now, by cloning the whole request object (not just Mercurius' context inside it) instead of mutating the original one inside the `Promise.all`.

test/batched.js
```diff
@@ -346,3 +347,44 @@ test('POST batched query with a resolver which succeeds and a resolver which thr
...
+      {
+        operationName: 'DoubleQuery',
+        query: 'query DoubleQuery { test }'
+      }
+    ]
+  })
+
+  t.ok(contextSpy.calledTwice)
+  t.ok(contextSpy.calledWith(0, 2, sinon.match(/TestQuery/)))
+  t.ok(contextSpy.calledWith(1, 2, sinon.match(/DoubleQuery/)))
```
simoneb, line 389: use sinon.assert, otherwise if these tests fail you'll just get a generic error instead of a telling error

lib/routes.js
```diff
@@ -352,13 +352,23 @@ module.exports = async function (app, opts) {
...
-            return response
-          })
-      ))
+      const operationsCount = request.body.length
+
+      Object.assign(request[kRequestContext], { operationsCount })
+
+      return Promise.all(request.body.map(async (r, operationId) => {
+        // Create individual reqs for multiple operations, otherwise reference the original req
+        const operationReq = operationsCount > 1 ? Object.create(request) : request
```
simoneb, line 361: I would like to hear @mcollina's opinion on the approach to cloning a request object. We need it to be cloned because every query in the batch needs a different context, but I don't know if this is a sensible way to clone it

test/batched.js
drakhart: Absolutely right. I took the chance to do it like this in other tests as well.

simoneb (approved):

test/batched.js
```diff
@@ -346,3 +347,44 @@ test('POST batched query with a resolver which succeeds and a resolver which thr
...
+      {
+        operationName: 'DoubleQuery',
+        query: 'query DoubleQuery { test }'
+      }
+    ]
+  })
+
+  sinon.assert.calledTwice(contextSpy)
+  sinon.assert.calledWith(contextSpy, 0, 2, sinon.match(/TestQuery/))
+  sinon.assert.calledWith(contextSpy, 1, 2, sinon.match(/DoubleQuery/))
```
simoneb, line 389: why are we not testing that operationId and operationCount are also passed around properly?

drakhart: I may not be understanding what you're saying, because I believe that's what we're doing here: when calling the resolver we're reading the `operationId` and `operationsCount` attributes from the context, and checking that they've been passed as expected: a call with `operationId === 0`, another call with `operationId === 1`, both of them with `operationsCount === 2`, and each one of them with the correct query contents.

simoneb: sorry I read that wrong, I thought those were the number of the call and the position of the argument for some reason!

drakhart: Haha, I feel you: I even had to double check the spy parameters when writing that assertion.

PR #852, opened 2022-08-27
feat: implement graphql-ws ping/pong

### LINKED ISSUE #851: Compatibility with `graphql-ws` `keepAlive`
I'm experimenting with switching from Apollo Express to Fastify and Mercurius.

I discovered something a bit weird. If I enable `keepAlive` in `graphql-ws`, it will send messages that look like `{"type": "ping"}` but it seems that something on the Mercurius side is not processing them correctly.

Screenshot 2022-08-27 at 17 50 32

On the other hand, If I enable `keepAlive` in Mercurius, it sends messages like `{type: "ka"}`:
Screenshot 2022-08-27 at 17 52 03
 
`graphql-ws` completely crashes on these messages and breaks the subscription:
Screenshot 2022-08-27 at 17 52 38
 
My understanding is that Mercurius uses `graphql-ws`, so what's the deal here? 🙂

andreialecu:
Fixes #851 

Here's a screenshot to confirm:

Screenshot 2022-08-27 at 20 12 54


andreialecu:
Not sure why some other lines got reformatted, I ran `npm run lint:fix`

andreialecu:
I set it to draft while I verify some things. I'm still unsure if the `graphql-ws` client can properly understand server-sent keep-alive messages, a.k.a. Server -> Client.

However, at least this works Client -> Server and mercurius now responds properly.

andreialecu:
Confirmed to work bidirectionally. There was a minor issue that I fixed and force pushed.

Notice by the arrows how some are incoming and some outgoing, but it works both ways.

Screenshot 2022-08-27 at 20 40 30

mcollina (approved):
lgtm

PR #853, opened 2022-08-27
fix: handle flaky tests

anonrig:
I still believe fake-timers are the main issue behind this. There are still some other tests being flaky due to fake timers tickAsync. I opened a pull request to discuss a potential solution.

Steps I've taken:

- Understand multiple tests were contradicting each other due to the same port number
- Run the test couple of times, it was still flaky
- Updated sinon fake timers and tap, since the only ambigious side of the test was the fake timers.
- Found the following issue: https://github.com/tapjs/node-tap/issues/709
- Changed for loop 10 times tickAsync(1000) with for loop 100 times with tickAsync(100). Which solved the issue on the retry-failed-services.js test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,7 @@
     "@graphql-tools/merge": "^8.0.0",
     "@graphql-tools/schema": "^8.0.0",
     "@graphql-tools/utils": "^8.0.0",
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #859, opened 2022-08-31
Properly clone context for batched queries

### LINKED ISSUE #858: Properly clone context for batched queries
#849 didn't finally fix #842.

For some reason I'm still investigating, `Object.create(request)` is not creating a truly independent clone of the request object, so the Mercurius context is still getting shared among all the batched queries.

This needs further investigation to find a proper fix, and once fixed hook spies (i.e. `onResolution`) should be added to double check the context values at different points of the request (since the resolver spy has proven to be incapable of catching the error).

drakhart:
fixes #858

- Create a shallow copy of the context inside `operationReq` to make sure that context's top level properties are cloned and not just referenced.
- Add more tests to make sure that the new context's top properties remain isolated through all the request lifecycle.

Request before the fix:
request_before

Request after the fix:
request_after

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -358,7 +358,12 @@ module.exports = async function (app, opts) {
 
       return Promise.all(request.body.map(async (r, operationId) => {
         // Create individual reqs for multiple operations, otherwise reference the original req
-        const operationReq = operationsCount > 1 ? Object.create(request) : request
+        const operationReq = operationsCount > 1
+          ? {
+              ...request,
+              [kRequestContext]: Object.create(request[kRequestContext])
+            }
+          : request
 
         Object.assign(operationReq[kRequestContext], { operationId })
 
--- a/test/hooks-with-batching.js
+++ b/test/hooks-with-batching.js
@@ -1,0 +1,78 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('fastify')
+const sinon = require('sinon')
+const GQL = require('..')
+
+test('batched query has an individual context for each operation through all the lifecycle hooks', async (t) => {
+  const app = Fastify()
+
+  const preParsingSpy = sinon.spy()
+  const preValidationSpy = sinon.spy()
+  const preExecutionSpy = sinon.spy()
+  const onResolutionSpy = sinon.spy()
+
+  const schema = `
+      type Query {
+        test: String
+      }
+    `
+
+  const resolvers = {
+    test: () => 'test'
+  }
+
+  await app.register(GQL, {
+    schema,
+    resolvers,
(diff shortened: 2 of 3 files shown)


lib/routes.js
```diff
@@ -358,7 +358,12 @@ module.exports = async function (app, opts) {
 
       return Promise.all(request.body.map(async (r, operationId) => {
         // Create individual reqs for multiple operations, otherwise reference the original req
-        const operationReq = operationsCount > 1 ? Object.create(request) : request
+        const operationReq = operationsCount > 1
+          ? {
+              ...request,
+              [kRequestContext]: Object.create(request[kRequestContext])
```
drakhart, line 364: Please note that this will do a shallow copy of `mercurius.requestContext`, which is all is needed to uniquely identify batched queries. If a deep clone is needed in the future we'll need to replace `Object.create` with a proper deep cloning function.

mcollina (approved):
lgtm

marco-ippolito (approved):

test/hooks-with-batching.js
```diff
@@ -0,0 +1,78 @@
...
+  sinon.assert.calledWith(preValidationSpy, 0, 2, sinon.match(/TestQuery/))
+  sinon.assert.calledWith(preValidationSpy, 1, 2, sinon.match(/DoubleQuery/))
+
+  sinon.assert.calledTwice(preExecutionSpy)
+  sinon.assert.calledWith(preExecutionSpy, 0, 2, sinon.match(/TestQuery/))
+  sinon.assert.calledWith(preExecutionSpy, 1, 2, sinon.match(/DoubleQuery/))
+
+  sinon.assert.calledTwice(onResolutionSpy)
+  sinon.assert.calledWith(onResolutionSpy, 0, 2, sinon.match(/TestQuery/))
+  sinon.assert.calledWith(onResolutionSpy, 1, 2, sinon.match(/DoubleQuery/))
```
simoneb, line 77: same comment as https://github.com/mercurius-js/mercurius/pull/849/files#r960757581

drakhart: Same response, but this time at hook level instead of at resolver level.

simoneb (approved):

lib/routes.js
drakhart: I tried that combination, but [here they say](url) "once you throw in an empty object parameter to Object.assign(), the object spread operator is consistently faster", and I thought that would be beneficial for a complex object like the request.

I also tried many, many other combinations (trust me I did 🥲), and in the end this was the most readable and convenient I could find.

simoneb:
I'll go ahead and merge this

PR #861, opened 2022-09-01
Breaking Change: statusCode 200 for all responses

### LINKED ISSUE #860: GQL spec compliance for http status codes (200 everywhere)
I've been migrating from Apollo to Mercurius and noticed that queries that return GQL errors have status code 500. This trips up our client code, thinking it's a Network Error instead of an application error.

I took a look at the spec here:
https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#applicationjson

It seems to mention that:

> The server SHOULD NOT use a 4xx or 5xx status code.

> Note: For compatibility with legacy servers, this specification allows the use of 4xx or 5xx status codes for failed requests where the response uses the application/json media type, but it is strongly discouraged. To use 4xx and 5xx status codes, please use the application/graphql-response+json media type

This has already been discussed here: https://github.com/mercurius-js/mercurius/issues/599#issuecomment-940734753 

I implemented a custom errorFormatter to address this as mentioned in the documentation.

However, now that the spec recommends `200` everywhere, I think it might make sense to change it in the next Mercurius version accordingly.

andreialecu:
Closes #860

### DIFF
--- a/test/errors.js
+++ b/test/errors.js
@@ -1030,7 +1030,7 @@ test('errors - should default to `statusCode` from error if present, when there 
   t.equal(res.statusCode, 409)
 })
 
-test('errors - should default to HTTP Status Code `200 Bad Request` if multiple errors are present and no data in the response', async (t) => {
+test('errors - should default to HTTP Status Code `200 OK` if multiple errors are present and no data in the response', async (t) => {
   t.plan(2)
 
   const schema = `
@@ -1128,7 +1128,7 @@ test('errors - should default to HTTP Status Code `400 Bad Request` if GraphQL v
   t.equal(res.statusCode, 400)
 })
 
-test('errors - should default to HTTP Status Code `200 Internal Server Error` if single error present but no status code defined', async (t) => {
+test('errors - should default to HTTP Status Code `200 OK` if single error present but no status code defined', async (t) => {
   t.plan(2)
 
   const schema = `
--- a/test/errors.js
+++ b/test/errors.js
@@ -1030,7 +1030,7 @@ test('errors - should default to `statusCode` from error if present, when there 
   t.equal(res.statusCode, 409)
 })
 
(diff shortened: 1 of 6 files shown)


andreialecu:
Note that this is a breaking change so it should be a major.

test/errors.js
```diff
@@ -1030,7 +1030,7 @@ test('errors - should default to `statusCode` from error if present, when there
   t.equal(res.statusCode, 409)
 })
 
-test('errors - should default to HTTP Status Code `400 Bad Request` if multiple errors are present and no data in the response', async (t) => {
+test('errors - should default to HTTP Status Code `200 Bad Request` if multiple errors are present and no data in the response', async (t) => {
```
andreialecu, line 1033:
```suggestion
test('errors - should default to HTTP Status Code `200 OK` if multiple errors are present and no data in the response', async (t) => {
```

test/errors.js
```diff
@@ -1128,7 +1128,7 @@ test('errors - should default to HTTP Status Code `400 Bad Request` if GraphQL v
   t.equal(res.statusCode, 400)
 })
 
-test('errors - should default to HTTP Status Code `500 Internal Server Error` if single error present but no status code defined', async (t) => {
+test('errors - should default to HTTP Status Code `200 Internal Server Error` if single error present but no status code defined', async (t) => {
```
andreialecu, line 1131:
```suggestion
test('errors - should default to HTTP Status Code `200 OK` if single error present but no status code defined', async (t) => {
```

mcollina (approved):
lgtm

mcollina:
@jonnydgreen should we ship a major now or wait?

jonnydgreen:
> @jonnydgreen should we ship a major now or wait?

@mcollina I don't see any existing PRs for minor/patch changes that could be released first so that sounds like a plan :)

In terms of what to include in the semver major release, what do you think about this PR, should we include as well? https://github.com/mercurius-js/mercurius/pull/710- I'm not 100% if there are conflicts with this PR though

PR #865, opened 2022-09-15
docs: replace uses of variadic `.listen` function

blimmer:
The documentation references `.listen(<port>)` in several places, but this variadic function is deprecated in Fastify v4 and throws this warning:

> (node:254) [FSTDEP011] FastifyDeprecation: Variadic listen method is deprecated. Please use ".listen(optionsObject)" instead. The variadic signature will be removed in `fastify@5`.

Since the current `package.json` requires Fastify v4+, this should be safe to do: https://github.com/mercurius-js/mercurius/blob/bbb8b269344f04ad9a50ae9210c69d31761ab6a7/package.json#L44

Deprecation guide: https://github.com/fastify/fastify/blob/2fd918298d55aaab74585c0b71081ad40988b280/docs/Guides/Migration-Guide-V4.md#deprecation-of-variadic-listen-signature

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ app.get('/', async function (req, reply) {
   return reply.graphql(query)
 })
 
-app.listen(3000)
+app.listen({ port: 3000 })
 ```
 
 ## Examples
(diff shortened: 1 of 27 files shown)


mcollina (approved):
lgtm

Issue #872, opened 2022-09-24
errorFormatter has wrong typescript types


fabulator:
errorFormatter in options has following typescript definition:

```typescript
interface ExecutionResult<
  TData = ObjMap<unknown>,
  TExtensions = ObjMap<unknown>,
> {
  errors?: ReadonlyArray<GraphQLError>;
  data?: TData | null;
  extensions?: TExtensions;
}

errorFormatter?: <TContext extends MercuriusContext = MercuriusContext>(
    execution: ExecutionResult,
    context: TContext
  ) => {
    statusCode: number;
    response: ExecutionResult;
  };
```

But `execution` doesn't always get `ExecutionResult`. It can also get an Error type. Here is an example:

```typescript
import Fastify from 'fastify';
import mercurius from 'mercurius';

const app = Fastify()

const schema = `
  type Query {
    field: Int
  }
`;

app.register(mercurius, {
    path: '/',
    schema,
    errorFormatter: (response) => {
        // @ts-ignore
        console.log(response.code);
        return { statusCode: 200, response }
    }
})


app.listen({ port: 3000 })
```

When I run the app and send GET to / it will log `MER_ERR_GQL_PERSISTED_QUERY_NOT_FOUND` which doesn't corespond with types.

I'm not sure if bug is wrong type (that there should be union with Error) or that Error shouldn't by passed here.

```typescript
const defaultErrorFormatter: (
    execution: ExecutionResult | Error,
    context: MercuriusContext
  ) => { statusCode: number; response: ExecutionResult };
```

But the types for the defaultErrorFormatter have Error as union option. So fix might be only changing the type.

fabulator:
@mcollina I checked the code and I think I know where is the problem. The [defaultErrorFormatter](https://github.com/mercurius-js/mercurius/blob/master/lib/errors.js#L49) has following:

```javascript
let errors = [err instanceof ErrorWithProps ? formatError(toGraphQLError(err)) : { message: err.message }]
```

Basically the first line do the all transformation to handle the Error type.

I could try to move it out of the default function and put it into the [lib/routes.js](https://github.com/mercurius-js/mercurius/blob/master/lib/routes.js). The errorFormatter would get transformed results without an Error Type.

I would also have to correct the state, when error is not instance of ErrorWithProps and it is transformed to `{ message: err.message }`. This doesn't satisfy typescript type `ReadonlyArray<GraphQLError>` which is specified in execution type.

If it sounds OK I could work on PR.

[...]

PR #879, opened 2022-09-28
fix: input type of errorFormatter

fabulator:
Solves issue https://github.com/mercurius-js/mercurius/issues/872

1. I change the type of execution input of formatter that it always has some errors. When execution doesn't have errors, it wouldn't fire formatter
2. I also changed output type of default formatter to `FormattedExecutionResult` so it reflects reality. But I left union of ExecutionResult and FormattedExecutionResult for errorFormatter in options. Based on graphql spec https://spec.graphql.org/draft/#sec-Errors GraphQLFormattedError (with a restricted data, only message, locations, path and extensions) is preferred. Having other properties "While non-specified entries are not violations, they are still discouraged."
3. I changed results of one test. It will now return correct graphql response with error array which I think is right.


mcollina (approved):
lgtm

jonnydgreen (approved):
LGTM

PR #881, opened 2022-09-29
fix: as gateway resolver key should handle digits (#880)

### LINKED ISSUE #880: As gateway resolverKey should handle digits
Hi! Using mercurius as gateway I found (thanks to @Eomm) inconsistencies in responses when the **queryId** and the consequent **requestkey** contain digits

## The Problem

Consider to execute the following query on a gateway connected to a node with the **Quick Start** schema:
```jsx
query { 
  line0: add(x: 2, y: 2) 
  line1: add(x: 1, y: 5) 
  line2: add(x: 2, y: 1) 
}
```

The **expected** result shoud be:
```jsx
data {
  line0: 4
  line1: 6
  line2: 3
}
```

But, due to **string replacement** on requestKey the inputs are squashed causing a response where **each** result key has the **same** value of the first element:
```jsx
data {
  line0: 4
  line1: 4
  line2: 4
}
```

## Solution

Replace digits values of the queryId **IMHO** is superfluous: 
```jsx
// lib/gateway/make-resolver.js:404
const resolverKey = `${queryId.replace(/\d/g, '_IDX_')}.${type.toString()}`
```

and once removed:
```jsx
// lib/gateway/make-resolver.js:404
const resolverKey = `${queryId}.${type.toString()}`
```
the inconsistencies are resolved probably gaining a small performance improvement

marcobeca:
Fixes #880

### DIFF
--- a/lib/gateway/make-resolver.js
+++ b/lib/gateway/make-resolver.js
@@ -401,7 +401,7 @@ function makeResolver ({ service, createOperation, transformData, isQuery, isRef
     const type = getNamedType(returnType)
 
     const queryId = generatePathKey(info.path).join('.')
-    const resolverKey = `${queryId.replace(/\d/g, '_IDX_')}.${type.toString()}`
+    const resolverKey = `${queryId}.${type.toString()}`
     const { reply, __currentQuery, lruGatewayResolvers, pubsub } = context
 
     const cached = lruGatewayResolvers != null && lruGatewayResolvers.get(`${__currentQuery}_${resolverKey}`)
--- a/test/gateway/resolver-key-handle-digits.js
+++ b/test/gateway/resolver-key-handle-digits.js
@@ -1,0 +1,186 @@
+'use strict'
+
+const { test } = require('tap')
+const Fastify = require('fastify')
+const GQL = require('../..')
+
+async function createTestService (t, schema, resolvers = {}) {
+  const service = Fastify()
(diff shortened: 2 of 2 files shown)


mcollina:
@jonnydgreen the replacement was added by you in https://github.com/mercurius-js/mercurius/pull/754, however the tests are not failing.

jonnydgreen:
> @jonnydgreen the replacement was added by you in https://github.com/mercurius-js/mercurius/pull/754, however the tests are not failing.

Interesting, I'll take a look this evening!

jonnydgreen:
Apologies, something came up so I wasn't able to have a look last night - I've blocked some time over the weekend so I'll have a look then if that's okay!

jonnydgreen:
@mcollina am I missing something here, it seems that the replacement (assuming you're talking about the `_IDX_` bit) wasn't added in the PR you linked? I only added the `type` suffix in #754 , which is not changed in this PR - so it should be fine that the tests are still passing?

Screenshot 2022-10-01 at 17 25 49

mcollina:
Ah,yeah! I made a mistake. I'll keep looking with git blame and see where it come from!

mcollina:
This was added in https://github.com/mercurius-js/mercurius/pull/130. I think it's safe to land.

mcollina (approved):
lgtm

PR #886, opened 2022-10-05
Fix: subscription.onConnect types

### LINKED ISSUE #871: Possibly incomplete return type of subscription.onConnect
I am trying to setup subscriptions and need some additional logic around connecting the client to the server using Websockets. After reading the docs on the `subscription.onConnect` method the return type is supposed to be "truthy" OR and object. When using mericurius in a TypeScript project the return type is set to `Record<string, any> | Promise<Record<string, any>>` but this will not allow for returning true/false which seem ok according to both the docs and the code itself.

The docs clearly states that:
> subscription.onConnect: Function A function which can be used to validate the connection_init payload. If defined it should return a truthy value to authorize the connection. If it returns an object the subscription context will be extended with the returned object.

Returning true/false in onConnect gives me the functionality I need but TypeScript is not happy with that. How are you supposed to return a truthy OR falsy value when the return type is set to `Record<string, any> | Promise<Record<string, any>>`? I would propose extending the return type with `boolean` to handle this inconsistency.

sameer-coder:
This PR extends the return types of `subscription.onConnect` to include `boolean` type

The docs mention:

> subscription.onConnect: Function A function which can be used to validate the connection_init payload. If defined it should return a **truthy value** to authorize the connection. If it returns an object the subscription context will be extended with the returned object.

closes #871 

cc @simoneb

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -527,7 +527,7 @@ export interface MercuriusCommonOptions {
         onConnect?: (data: {
           type: 'connection_init';
           payload: any;
-        }) => Record<string, any> | Promise<Record<string, any>>;
+        }) => Record<string, any> | Promise<Record<string, any>> | boolean;
         onDisconnect?: (context: MercuriusContext) => void | Promise<void>;
         keepAlive?: number,
         fullWsTransport?: boolean,
--- a/index.d.ts
+++ b/index.d.ts
@@ -527,7 +527,7 @@ export interface MercuriusCommonOptions {
         onConnect?: (data: {
           type: 'connection_init';
           payload: any;
-        }) => Record<string, any> | Promise<Record<string, any>> | boolean;
+        }) => any;
         onDisconnect?: (context: MercuriusContext) => void | Promise<void>;
         keepAlive?: number,
         fullWsTransport?: boolean,
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -544,6 +544,14 @@ app.register(mercurius, {
   schema,
   resolvers,
   subscription: {
+    onConnect: () => true
+  }
+})
+
+app.register(mercurius, {
+  schema,
+  resolvers,
+  subscription: {
     emitter
   }
 })


simoneb:
This addresses the OP but I wonder if we shouldn't just be more permissive here. If the docs say that any truthy value goes, then it means that this function should be able to return anything. Wouldn't simply model the return type to be any work? Or what about the existing types, plus `| any`? not sure if this makes any sense though

sameer-coder:
> This addresses the OP but I wonder if we shouldn't just be more permissive here. If the docs say that any truthy value goes, then it means that this function should be able to return anything. Wouldn't simply model the return type to be any work? Or what about the existing types, plus `| any`? not sure if this makes any sense though

Good point, I agree. I think it makes more sense to extend the existing types with `any` since a truthy value is not just limited to boolean values.

simoneb:
> > This addresses the OP but I wonder if we shouldn't just be more permissive here. If the docs say that any truthy value goes, then it means that this function should be able to return anything. Wouldn't simply model the return type to be any work? Or what about the existing types, plus `| any`? not sure if this makes any sense though
> 
> Good point, I agree. I think it makes more sense to extend the existing types with `any` since a truthy value is not just limited to boolean values.

With that in mind, does `| any` make sense (or is it even valid TS) or should it just be `any` alone?

sameer-coder:
> 



> > > This addresses the OP but I wonder if we shouldn't just be more permissive here. If the docs say that any truthy value goes, then it means that this function should be able to return anything. Wouldn't simply model the return type to be any work? Or what about the existing types, plus `| any`? not sure if this makes any sense though
> > 
> > 
> > Good point, I agree. I think it makes more sense to extend the existing types with `any` since a truthy value is not just limited to boolean values.
> 
> With that in mind, does `| any` make sense (or is it even valid TS) or should it just be `any` alone?

I think specifying `any` would disable type checking for the field so there is no point in doing  `| any`. I have updated the code to use only `any`.

simoneb (commented):
I wonder if having more explicit typing wouldn't anyway serve as documentation, although not useful for type checking. As an example, the docs don't say that this can be an async function, but the typings do. I think it may make sense to still use unnecessary typing for the sake of documentation. If we wanted to capture everything we should add `| any | Promise<any>` to the existing typings, but maybe just `any | Promise<any>` is good enough.

I dunno, not a typescript expert here.

simoneb:
Then I'd say let's go with `any` and change the docs to clarify what can be returned from this function (including the fact that it can be async)

index.d.ts
```diff
@@ -527,7 +527,7 @@ export interface MercuriusCommonOptions {
         onConnect?: (data: {
```
ilteoood, line 527:
```suggestion
        onConnect?: <T = any,>(data: {
```

index.d.ts
```diff
@@ -527,7 +527,7 @@ export interface MercuriusCommonOptions {
...
           type: 'connection_init';
           payload: any;
-        }) => Record<string, any> | Promise<Record<string, any>>;
+        }) => any;
```
ilteoood, line 530:
```suggestion
        }) => T;
```

ilteoood (commented):
I've made a suggestion to better explain what I mean with fallback of a generic type.

simoneb:
@ilteoood I don't see the value of doing that. That function is not used by user code, it's used by mercurius. The point of adding typing would be to either 1) prevent users from typing the return value wrong, which we can't do as the function can return anything or 2) serve as documentation, which is the only valuable thing we can do with typings here. For this, using `any` is as good as using generics.

Issue #887, opened 2022-10-10
Clean up index.js file from gateway init instructions


codeflyer:
The initialization of the gateway is done, if an attribute is set, in the main `index.js` file. (https://github.com/mercurius-js/mercurius/blob/master/index.js#L154)

Clean up the file to remove all the logic related to the `gateway` and the `federation` and transform in something like:

```
if(gateway) {
   initGateway(app, opts)
}
```

where 'initGateway' is a method exported from the `gateway.js` file.

This is the first step of a refactoring that will move the `gateway/federation` to an external module.

PR #889, opened 2022-10-11
Clean up index.js file from gateway init instructions #887

codeflyer:
Cleanup of the index file.

the stuff related to the gateway are moved in a separate file and the initialization is done with:

```
  let gateway
  let lruGatewayResolvers
  if (opts.gateway) {
    lruGatewayResolvers = buildCache(opts)
    gateway = await initGateway(opts, fastifyGraphQl, app)

    schema = gateway.schema
  }
```

Further improvement are required but this is the first step required for the gateway extraction.

@simoneb @mcollina


mcollina (approved):
lgtm

jonnydgreen (approved):
LGTM

PR #896, opened 2022-10-24
Use the real Request object in subscriptions

mcollina:
Signed-off-by: Matteo Collina <hello@matteocollina.com>

This PR does a few things:

1. uses the real `Request` object in the subscription `Context`. Reply is still "faked" as there is no `Reply` for websockets.
2. copies over all the data incoming as part of `connection_init` packet to `req.headers`. If there is a `headers` object that's used instead (as it was before). This change was needed because GraphiQL treats all the payload as headers.


jonnydgreen (approved):
LGTM - reckon we should add the details/behaviour in the PR description to the docs for future reference?

mcollina:
Docs added.

PR #898, opened 2022-10-25
feat: @defer support

igrlk:
Hi everyone ✋  As some of you may have noticed, **apollo** and **graphql-js** added support for `@defer` directive, so it's time to do the same in Mercurius!

### 🤔 What is @defer and why do we need that?
`@defer` is a directive that allows you to receive query data incrementally. There is a bunch of information about it but a decent source is [Apollo Docs @defer](https://www.apollographql.com/docs/react/data/defer/).

### ⌛ Current state of things
* [graphql-spec for @defer](https://github.com/graphql/graphql-spec/pull/742) is not yet finished, although looks like we can rely on it, because 👇 
* [graphql-js](https://github.com/graphql/graphql-js/pull/3659) implemented `@defer` and made it available in `graphql@17.0.0-alpha.2`
* [apollo-client](https://github.com/apollographql/apollo-client/pull/10018) and [apollo-server](https://github.com/apollographql/apollo-server/pull/6827) added support for `@defer`.
  * Apollo-Server changes are not yet in main branch, but in the [next major version branch](https://github.com/apollographql/apollo-server/tree/version-4)

### 📦 What this PR does
It uses the new [GraphQL incremental api](https://github.com/graphql/graphql-js/blob/main/src/execution/execute.ts#L316) **(if it's available! otherwise we use old api for backwards-compatibility)** to execute queries.
The incremental api itself does all the magic for us and returns either the classic response `(body.data.myField)` or the incremental one with stream (see more in the [api type definitions](https://github.com/graphql/graphql-js/blob/main/src/execution/execute.ts#L316))

Client should send us the following header `accept: multipart/mixed; deferSpec=20220824` in order to get a stream response back. Otherwise, the error is thrown that doesn't allow client to get the data.

The PR also moves to a new signature of `GraphQLError` since the old one is deprecated and its support is removed in [graphql@17](https://github.com/graphql/graphql-js/pull/3577)

Additionally, it bumps the `graphql-js` to `17.0.0-alpha.2` (latest available right now). Once PR has all the tests passed and feedback incorporated, it's going to be cut off to a separate release branch, instead of getting merged to master directly because of `graphql@17` which is still in alpha

### 🧰 Local development / testing
If you wish to see how it's going to look like in a real app, I've put together an [Apollo Client/Server + Mercurius setup](https://github.com/igrlk/defer-directive-example) that allows you to run a query with `@defer` against either:
* This Mercurius branch, to see what we've got
* Apollo Server, to see how it works there

### ⚠️ Things that aren't yet done
- [ ] `graphql-jit` is not adopted to use `graphql@17` yet and I asked code owners if they want to make an upgrade https://github.com/zalando-incubator/graphql-jit/issues/184

:octocat: Please take into account that all the things I mentioned in the PR are my suggestions based on what I've read and heard. I am open to the ideas you might have, especially because it's my first PR to this repository.

I want to give a huge shoutout to the Mercurius team for their **amazing** hard work maintaining this project and to Apollo team who were the first in `@defer` implementation, and whose examples I've used to implement the changes in this repo.

Cheers ✌️


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mercurius",
-  "version": "11.1.0",
+  "version": "11.2.0",
```
mcollina, line 3: please revert this

test/directives.js
```diff
@@ -962,3 +963,147 @@ type User {
     }
   })
 })
+
+if (canUseIncrementalExecution) {
```
mcollina, line 967: Please move all of those tests into their own file.

test/directives.js
```diff
@@ -962,3 +963,147 @@ type User {
...
+
+  test('errors with @defer when used with wrong "accept" header', async t => {
+    const wrongAcceptValues = [
+      '',
+      'application/json',
+      'multipart/mixed',
+      'multipart/mixed; deferSpec=12345'
+    ]
+
+    for (const accept of wrongAcceptValues) {
```
mcollina, line 1035: move this for loop outside of the test definition

test/directives.js
```diff
@@ -962,3 +963,147 @@ type User {
...
+  })
+
+  test('works with @defer when used with correct "accept" header', async t => {
+    const correctAcceptValues = [
+      'multipart/mixed; deferSpec=20220824',
+      'multipart/mixed; deferSpec=20220824, application/json',
+      'application/json, multipart/mixed; deferSpec=20220824'
+    ]
+
+    for (const accept of correctAcceptValues) {
```
mcollina, line 1072: move this loop outside of the test definition

test/directives.js
```diff
@@ -962,3 +963,147 @@ type User {
...
+        body: `\r
+---\r
+content-type: application/json; charset=utf-8\r
+\r
+{"hasNext":true,"data":{"allProducts":[{"delivery":{},"sku":"sku","id":"0"}]}}\r
+---\r
+content-type: application/json; charset=utf-8\r
+\r
+{"hasNext":false,"incremental":[{"path":["allProducts",0,"delivery"],"data":{"estimatedDelivery":"25.01.2000","fastestDelivery":"25.01.2000"}}]}\r
+-----\r
```
mcollina, line 1100: Can you also add a test using undici.fetch()? It _should_ work and support multipart responses.

index.js
```diff
@@ -555,9 +557,95 @@ const plugin = fp(async function (app, opts) {
...
+          'HTTP responses. To enable incremental delivery support, add the HTTP ' +
+          "header 'Accept: multipart/mixed; deferSpec=20220824'.",
+          // Use 406 Not Accepted TODO
+          { extensions: { http: { status: 406 } } }
+        )
+      }
+
+      reply.header('content-type', 'multipart/mixed; boundary="-"; deferSpec=20220824')
+
+      return Readable.from(
```
mcollina, line 592: amazing ;).

index.js
```diff
@@ -555,9 +557,95 @@ const plugin = fp(async function (app, opts) {
...
+          new Negotiator({
+            headers: { accept: acceptHeader }
+          }).mediaType([
+            // mediaType() will return the first one that matches, so if the client
+            // doesn't include the deferSpec parameter it will match this one here,
+            // which isn't good enough.
+            MEDIA_TYPES.MULTIPART_MIXED_NO_DEFER_SPEC,
+            MEDIA_TYPES.MULTIPART_MIXED_EXPERIMENTAL
+          ]) === MEDIA_TYPES.MULTIPART_MIXED_EXPERIMENTAL
+        )
```
mcollina, line 576: instead of negotiatior, please use 

https://github.com/fastify/fastify-accepts

test/directives.js
```diff
@@ -962,3 +963,147 @@ type User {
...
+
+  const allProducts = new Array(1).fill(0).map((_, index) => ({
+    id: `${index}`,
+    sku: 'sku'
+  }))
+
+  const resolvers = {
+    Query: {
+      allProducts: () => {
+        return allProducts
```
mcollina, line 997: Could you return a stream here?

index.js
```diff
@@ -555,9 +557,95 @@ const plugin = fp(async function (app, opts) {
...
+      ) {
+        // The client ran an operation that would yield multiple parts, but didn't
+        // specify `accept: multipart/mixed`. We return an error.
+        throw new Error(
+          'Mercurius server received an operation that uses incremental delivery ' +
```
mcollina, line 581: Do not share that the server is implemented on Mercurius

index.js
```diff
@@ -555,9 +557,95 @@ const plugin = fp(async function (app, opts) {
...
+          '(@defer or @stream), but the client does not accept multipart/mixed ' +
+          'HTTP responses. To enable incremental delivery support, add the HTTP ' +
+          "header 'Accept: multipart/mixed; deferSpec=20220824'.",
+          // Use 406 Not Accepted TODO
+          { extensions: { http: { status: 406 } } }
+        )
+      }
```
mcollina, line 588: If I read this error correctly, in case the request was not made with multipart accept it will fail.
Can we fall back to the previous behavior instead, i.e. accumulate all results.

mcollina (changes requested):
This is a amazing on multiple level! Deferred resolver is massive new feature coming in graphql v17!

(Given that this PR is also made possible by Readable.from() and node stream iterator support... I'm even more in awe. I championed and implemented those features).

I'm a bit concerned how we could land this and let people experiment about it for now.

---

It's better if we put `@defer` under an option. This will allow to throw if jit is enabled or it's missing graphql-jit.

package.json
igrlk: Is there a rule on when the package version gets bumped in this repo?

index.js
igrlk: Not sure I understand this. What do you mean by `previous behavior` and `accumulate all results`?

This line, btw, isn't necessary and I removed it. Although, I still throw an error [here](https://github.com/mercurius-js/mercurius/pull/898/files#diff-3fad171d5e074854125d96b9140d539b46dc2eae91814bf3b7b17fd99b2e2d55R93).

As I understand, there is no way of not executing `@defer` if it's in the query, so we have the deferred result anyway. Are you asking about just sending it back anyway, and not throwing an error if the `accept` header is incorrect?

mcollina: You are on the spot. I completely misunderstood this implementation and throwing is the correct behavior.

index.js
```diff
@@ -555,9 +557,93 @@ const plugin = fp(async function (app, opts) {
...
+            // doesn't include the deferSpec parameter it will match this one here,
+            // which isn't good enough.
+            MEDIA_TYPES.MULTIPART_MIXED_NO_DEFER_SPEC,
+            MEDIA_TYPES.MULTIPART_MIXED_EXPERIMENTAL
+          ]) === MEDIA_TYPES.MULTIPART_MIXED_EXPERIMENTAL
+        )
+      ) {
+        // The client ran an operation that would yield multiple parts, but didn't
+        // specify `accept: multipart/mixed`. We return an error.
+        throw new Error(
```
mcollina, line 580: Please create an error in https://github.com/mercurius-js/mercurius/blob/master/lib/errors.js and use it here

lib/errors.js
```diff
@@ -35,12 +35,14 @@ function toGraphQLError (err) {
...
-    err,
-    err.extensions
+    {
+      nodes: err.nodes,
+      source: err.source,
+      positions: err.positions,
+      path: err.path,
+      originalError: err,
+      extensions: err.extensions
+    }
```
mcollina, line 45: why this change?

igrlk: That's because positional arguments in GraphQLError were deprecated but `graphql@16` was still supporting them.
And `graphql@17` drops their support in favour of object argument - https://github.com/graphql/graphql-js/pull/3577

mcollina:
How could I ship this today and keep the current dependency at `graphql@16`?

lib/errors.js
igrlk: Yup! `16.0` that Mercurius uses is already having this code that migrates the positional args into a new signature if needed
Screenshot 2022-10-26 at 11 46 51

mcollina: awesome!

package.json
mcollina: the releaser bumps the version when they release.

igrlk:
> How could I ship this today and keep the current dependency at `graphql@16`?

I didn't update any gql-related dependencies in `package.json` and the whole change is not breaking, since the new functionality only works if `graphql@17` is installed.

So I think it's safe to ship this once I incorporate the feedback and CI is green. Appreciate if you can help with the things you mentioned - otherwise I'll do this later today 🙂

test/defer.js
```diff
@@ -0,0 +1,146 @@
...
+const Fastify = require('fastify')
+const mercurius = require('../index')
+const { canUseIncrementalExecution } = require('../lib/util')
+
+if (canUseIncrementalExecution) {
+  const schema = `
+    directive @defer(
+      if: Boolean! = true
+      label: String
+    ) on FRAGMENT_SPREAD | INLINE_FRAGMENT
```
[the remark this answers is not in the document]
mcollina: I would do this behind a `defer: true` option, set to false by default.

package.json
```diff
@@ -64,6 +64,7 @@
     "graphql": "^16.0.0",
```
mcollina, line 64: I think we are not testing this behavior in CI, right? We should. The `peerDependency` might not be correct.

igrlk: I could try adding CI tests that install `graphql@17` if you can point me to the right direction - how and where you add them?

Also, what do you mean by `peerDependency not correct`? It didn't change in my PR, right?

mcollina: How would an end user use defer? As you can see we depend upon `graphql`.

igrlk: I see. So the Apollo Server, as a reference, didn't update graphql version and they will only do that when graphql@17 is released. Until that, it's the change that is implemented and tested on CI, but not available to the end users since it's still in alpha

igrlk: That being said, the idea of mine was to do the same here - start supporting it but not allow users to use it until `graphql@17` is officially released

igrlk: In the future, when `graphql@17` is live, we will update the mercurius dependency and everyone will be able to use defer

mcollina: I'm happy to cut a next branch for you, so we can land this with the alpha version of graphql-js.

wdyt?

mcollina: +1

test/defer.js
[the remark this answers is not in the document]
igrlk: Hey @mcollina I've added `opts.defer`, but I'm not quite sure if we really need that, since we decided to cut this off to a new release branch where we assume `@defer` will be available to everyone.

I can remove `opts.defer` and just leave its directive definition so that we always have this functionality with `graphql@17`.

What do you think?

test/defer.js
```diff
@@ -0,0 +1,211 @@
...
+-----\r
+`)
+
+    t.same(res.status, 200)
+    t.same(res.headers.get('content-type'), 'multipart/mixed; boundary="-"; deferSpec=20220824')
+
+    t.teardown(async () => {
+      await reader.releaseLock()
+      app.close()
+      process.exit()
```
igrlk, line 206: Not sure if I wasn't iterating through the stream correctly, but the test never stops - added `process.exit()` to make it passing.
@mcollina can you please help here, maybe you see a problem in my code?

package.json
igrlk: I incorporated the feedback and going to see what's going on there with JIT compiler when I have some time

igrlk:
@mcollina I believe you saw my issue in `graphql-jit` - looking forward to hearing back from them re graphql 17 support. Should we keep this PR open until they give us more details?

mcollina:
It's likely better to wait until we hear back, I might be wrong but there is no urgency to ship given graphl@17 is still in alpha.

glasser:
> Apollo-Server changes are not yet in main branch, but in the [next major version branch](https://github.com/apollographql/apollo-server/tree/version-4)

Just to clarify: we have released support for defer/stream in Apollo Server 4 (which is now the current major version), but we are considering it experimental for now for various reasons including that it requires unreleased `graphql-js`.

Exciting to see Mercurius implementing it as well! Happy to see you're also requiring the client to opt in with `deferSpec` so that the whole ecosystem can cleanly adjust to any final changes that may show up before the GraphQL spec receives the final version.

mcollina (commented):
I've rebased this on top of the next branch.
We can land as soon as the tests pass

index.js
```diff
@@ -187,6 +189,20 @@ const plugin = fp(async function (app, opts) {
     })
   }
 
+  if (opts.defer) {
```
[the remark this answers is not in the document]
igrlk: I've added an error throwing when both jit and defer are enabled & updated the docs to have `defer` option there and in typescript definitions

igrlk:
> I've rebased this on top of the next branch.
We can land as soon as the tests pass

Thanks! I've made more changes that you requested.

I'm going to check if we can get the response from graphql-jit team or I can maybe make the necessary changes in their project, since we're blocked by that.

igrlk:
I didn't get any response from `graphql-jit` so decided to make a PR to migrate it to `graphql@17` - hopefully someone will respond on it.
https://github.com/zalando-incubator/graphql-jit/pull/185

BorisTB:
Hi! What needs to be done here to merge this? I’d really like to use this feature so I’ll gladly help.

mcollina:
I think a public release of GraphQL v17?

ddeboer:
Would be great to have this available in Mercurius!

mcollina:
Might be the time! See #1256

PR #900, opened 2022-10-27
Prevent parsing schema exceptions when importing directives 

Igloczek:
This fix prevents the app from breaking if schema contains [directives imports](https://www.apollographql.com/docs/federation/federated-types/federated-directives/#importing-directives) like this:

```gql
extend schema
  @link(url: "https://specs.apollo.dev/federation/v2.0",
        import: ["@key", "@shareable"])
```

I'm not sure if those imports actually work with Mercurius, but are a valid syntax, so shouldn't break it.


mcollina (approved):
lgtm

PR #901, opened 2022-10-29
Explicitly say in the docs that JIT is disabled by default

igrlk:
I think that's a good thing to mention that JIT is disabled by default and should be explicitly enabled to get the performance benefits. I personally thought that it's enabled by default, but then I started digging into Mercurius and found how JIT is actually used there, and that it's not enabled by default.

Screenshot of the updated docs:
Screenshot 2022-10-29 at 21 21 04

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -52,6 +52,7 @@
 
 - `jit`: Integer. The minimum number of execution a query needs to be
   executed before being jit'ed.
+  - Default: `0`, jit is disabled.
 - `routes`: boolean. Serves the Default: `true`. A graphql endpoint is
   exposed at `/graphql`.
 - `path`: string. Change default graphql `/graphql` route to another one.


igrlk:
Also curious, why not is not enabled by default? Is it because it's just a plugin, addition to the mercurius, not the main point of it?

Just interesting what are the reasons not to enable it as a default experience? Thanks!

smolinari:
I personally think JIT is a feature that is a main selling point for Mercurius. It's where it's superior performance comes from. That being said, JIT is only really interesting in a system running in production. As long as code is changing (i.e. during development), JIT is less important. I'm wondering though, does "0" calls of a query to JIT it mean JIT is completely off, or does it mean is always on? :grin: 

But for sure, it would be a good idea to mention it being "0" as default (and note what that means), since all defaults should be mentioned for configs and most are for the other plugin options.

Scott

mcollina (approved):
lgtm

PR #907, opened 2022-11-01
feat: add types for object in graphiql configuration

codeflyer:
Adds the type definition for the configuration of graphiql using an object.


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? We use tsd for typescript.

mcollina (approved):
lgtm

PR #916, opened 2022-11-14
feat: remove gateway and federation modules

codeflyer:


lib/errors.js
```diff
@@ -57,19 +43,14 @@ function defaultErrorFormatter (execution, ctx) {
   const errors = execution.errors.map((error) => {
     log.info({ err: error }, error.message)
 
-    // parses, converts & combines errors if they are the result of a federated request
-    if (error.message === FEDERATED_ERROR.toString() && error.extensions) {
-      return error.extensions.errors.map(err => toGraphQLError(err))
-    }
```
mcollina, line 63: Are we overriding the error handler in the gateway?

lib/hooks.js
```diff
@@ -152,21 +109,13 @@ function onEndHookRunner (fn, request) {
...
   hooksRunner,
   preExecutionHooksRunner,
   hookRunner,
-  gatewayHookRunner,
-  preGatewayExecutionHooksRunner,
   preParsingHookRunner,
   onResolutionHookRunner,
-  onEndHookRunner,
-  onGatewayReplaceSchemaHookRunner
+  onEndHookRunner
```
mcollina, line 120: The functionality of Gateway hooks seems to be missing in the new package.

package.json
```diff
@@ -1,20 +1,19 @@
...
-  "description": "Fastify GraphQL adapter with gateway and subscription support",
+  "description": "Fastify GraphQL adapter with subscription support",
   "main": "index.js",
   "types": "index.d.ts",
   "scripts": {
     "docs": "docsify serve",
-    "unit": "tap test/*.js test/gateway/*.js test/internals/*.js",
-    "cov": "tap  --coverage-report=html -J test/*.js test/gateway/*.js",
+    "unit": "tap test/*.js test/internals/*.js",
+    "cov": "tap --jobs=1 --coverage-report=html test/*.js test/internals/*.js",
```
mcollina, line 10: why?

test/subscription-hooks.js
```diff
@@ -116,9 +116,9 @@ test('subscription - hooks basic', async t => {
     t.type(context, 'object')
     t.ok('preSubscriptionExecution called')
   })
-  app.graphql.addHook('preGatewaySubscriptionExecution', async (schema, document, context) => {
-    t.fail('preGatewaySubscriptionExecution should not be called in non-gateway mode')
-  })
+  // app.graphql.addHook('preGatewaySubscriptionExecution', async (schema, document, context) => {
+  //   t.fail('preGatewaySubscriptionExecution should not be called in non-gateway mode')
+  // })
```
mcollina, line 121: remove

lib/subscription.js
```diff
@@ -6,7 +6,7 @@ const { kHooks } = require('./symbols')
 const SubscriptionConnection = require('./subscription-connection')
 const { getProtocolByName } = require('./subscription-protocol')
 
-function createConnectionHandler ({ subscriber, fastify, onConnect, onDisconnect, lruGatewayResolvers, entityResolversFactory, subscriptionContextFn, keepAlive, fullWsTransport }) {
+function createConnectionHandler ({ subscriber, fastify, onConnect, onDisconnect, entityResolversFactory, subscriptionContextFn, keepAlive, fullWsTransport }) {
```
mcollina, line 9: How is this functionality implemented in the gateway?

lib/subscription.js
```diff
@@ -62,7 +61,7 @@ function createConnectionHandler ({ subscriber, fastify, onConnect, onDisconnect
 }
 
 module.exports = async function (fastify, opts) {
-  const { getOptions, subscriber, verifyClient, onConnect, onDisconnect, lruGatewayResolvers, entityResolversFactory, subscriptionContextFn, keepAlive, fullWsTransport } = opts
+  const { getOptions, subscriber, verifyClient, onConnect, onDisconnect, entityResolversFactory, subscriptionContextFn, keepAlive, fullWsTransport } = opts
```
mcollina, line 64: How is this functionality implemented in the gateway?

lib/subscription-connection.js
```diff
@@ -278,13 +275,11 @@ module.exports = class SubscriptionConnection {
           return print(document)
         },
         pubsub: sc,
-        lruGatewayResolvers: this.lruGatewayResolvers,
         reply: {
           entityResolversFactory: this.entityResolversFactory,
           request: context.request,
           [kLoaders]:
-            subscriptionLoaders && subscriptionLoaders.create(context),
-          [kEntityResolvers]: this.entityResolvers
```
mcollina, line 287: This was important.

mcollina (changes requested):

lib/errors.js
codeflyer: Yes, the gateway replace the errorHandler, if not passed as opts. And the gateway handler has the check above.

lib/hooks.js
codeflyer: It's there: https://github.com/mercurius-js/mercurius-gateway/blob/main/lib/handlers.js#L12

package.json
codeflyer: that's an error I needed to run to check what test was covering a specific code and forgot to remove it.

test/subscription-hooks.js
codeflyer: yes

lib/subscription.js
codeflyer: Instead of being added to the `context` is passed as parameter to the `makeResolver` function where it's used.

https://github.com/mercurius-js/mercurius-gateway/blob/main/lib/gateway/build-gateway.js#L120
https://github.com/mercurius-js/mercurius-gateway/blob/main/lib/gateway/make-resolver.js#L453

lib/subscription.js
codeflyer: As wrote above, it's passed with a closure instead of having it in the context.

lib/subscription-connection.js
codeflyer: `this.entityResolvers` seems not initialized. [kEntityResolvers] is always undefined. Unless I'm missing something

lib/subscription.js
codeflyer: While checking this stuff I found an error in the gateway, the values is not passed in the refresh action. Pr is coming

lib/hooks.js
```diff
@@ -1,17 +1,13 @@
 'use strict'
 
-const applicationHooks = [
-  'onGatewayReplaceSchema'
```
jonnydgreen, line 4: I think this was used in the validation of hooks. How does the validation still work for `onGatewayReplaceSchema` with the new package?

package.json
```diff
@@ -50,7 +49,7 @@
     "snazzy": "^9.0.0",
     "split2": "^4.0.0",
     "standard": "^17.0.0",
-    "tap": "^16.3.0",
+    "tap": "16.3.0",
```
jonnydgreen, line 52: Is this meant to be pinned?

jonnydgreen (commented):
What are your plans with:
 - Updating the documentation
 - Updating the examples
 - Updating the benchmark stuff
Will this come in subsequent PRs?

package.json
codeflyer: YEs it is, there is an issue with the 16.3.1 version due a babel dependency that breaks the `npm install` on windows/node 14

https://github.com/tapjs/node-tap/issues/843

lib/subscription-connection.js
mcollina: I'm pretty sure those are not _always_ undefined.

lib/hooks.js
mcollina: https://github.com/mercurius-js/mercurius-gateway/blob/main/lib/handlers.js#L6 is pretty bad. I missed it when we did the extraction there, it should not do in-depth requirement and depend on an internal.

Moreover the hooks are not covered in the gateway docs.

lib/subscription.js
codeflyer: as @jonnydgreen mentioned in another comment the hooks as is are not working at all. The validation is done by mercusiur, then once removed the `gateway` hooks is not possible to add it.


I will add a hook management internal to the gateway, that should be called with:

```
  app.graphql.gateway.addHook('onSomeGatewayHook', async (execution, context) => {
    ...
  })
```

WDYT? @mcollina @jonnydgreen

lib/subscription-connection.js
codeflyer: The `this.entityResolvers` isn't mentioned anywhere else in the `SubscriptionConnection` class, is not initialized.
Where the subscriptionConnection class is used there is no any external assignment like `connection. entityResolvers = something`

I checked on the original commit when it was added the line as well and same stuff.

I've added a log for the value in the tests, same. The `this.entityResolvers` is always undefined.

lib/subscription.js
codeflyer: And once we have defined the right approach I'll update docs and examples properly

codeflyer:
@jonnydgreen @mcollina a question related docs, examples and benchmarks.

Where do we want to keep them.

I was thinking to keep the docs for the gateway in mercurius. 

Instead benchmark and examples will be in his own repo.

package.json
jonnydgreen: gotcha!

jonnydgreen:
R.e. where to keep docs, very good question. I think at the very least, the Mercurius website (mercurius.dev) should be the entry point for developers. Now that things are split up into packages, I think there are several options for implementation, each with their own advantages and disadvantages

|Potential solution| Notes|
|-|-|
|Provide links to each package docs from mercurius.dev|Scalable. Lose initial information (it's still there but you have to link out) about gateways - I think this is core info that should be retained in the website in some way|
|README file in each package repo but also have full docs about each package in core mercurius repo as you suggested|No loss in info, but hard to keep in sync - must be vigilant on this|
|Build a website that combines docs from the github pages site for each repo|Scalable but fair bit of initial work|

WDYT?

codeflyer:
I think the best solution in the long term is a repo with just the docs. Even if there is some work to do in the beginning is a required step to create a common place for the mercurius ecosystem.
As a developer that's deciding to adopt `mercurius` as GQL server I consider it a plus to have access to a complete solution that solves the main issues.
With the federation integrated this wasn't a big problem but with the separation I thinks is important to show that the product is unique even if is split in modules.

examples/gateway-subscription.js
```diff
@@ -1,14 +1,14 @@
 'use strict'
 const Fastify = require('fastify')
-const mercurius = require('..')
+const mercuriusWithFederation = require('@mercurius/federation')
+const mercuriusWithGateway = require('@mercurius/gateway')
```
mcollina, line 4:
```suggestion
const mercuriusWithFederation = require('@mercuriusjs/federation')
const mercuriusWithGateway = require('@mercuriusjs/gateway')
```

examples/gateway.js
```diff
@@ -1,15 +1,16 @@
 'use strict'
 const Fastify = require('fastify')
+const mercuriusWithFederation = require('@mercurius/federation')
+const mercuriusWithGateway = require('@mercurius/gateway')
```
mcollina, line 4:
```suggestion
const mercuriusWithFederation = require('@mercuriusjs/federation')
const mercuriusWithGateway = require('@mercuriusjs/gateway')
```

examples/hooks-gateway.js
```diff
@@ -1,15 +1,16 @@
 'use strict'
 const Fastify = require('fastify')
+const mercuriusWithFederation = require('@mercurius/federation')
+const mercuriusWithGateway = require('@mercurius/gateway')
```
mcollina, line 4:
```suggestion
const mercuriusWithFederation = require('@mercuriusjs/federation')
const mercuriusWithGateway = require('@mercuriusjs/gateway')
```

marco-ippolito:
It's been a couple of weeks, if  @codeflyer is too busy I can take over @mcollina

marco-ippolito:
replaced by: https://github.com/mercurius-js/mercurius/pull/942

PR #917, opened 2022-11-15
feat: adds graphql resolve info to loader ts definition

dsoffiantini:
Adds info object into loader definition, as a loader with disabled caching will provide the info object


dsoffiantini:
> Can you add a test for these typeS?

@mcollina Added!

mcollina (approved):
lgtm

PR #920, opened 2022-11-21
feat: export SubscriptionClient class

codeflyer:
in this PR:

- Export the SubscriptionClient
- update the type definitions
- add test to the types exported

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -833,3 +833,36 @@ export interface PreExecutionHookResponse<TError extends Error> {
   document?: DocumentNode
   errors?: TError[]
 }
+
+export interface SubscriptionClientConfig {
+  protocols: Array<string>,
+  reconnect: boolean,
+  maxReconnectAttempts: number,
+  serviceName: string,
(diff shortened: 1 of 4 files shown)


mcollina:
I would rather export it in a separate module instead. Wdyt?

codeflyer:
> I would rather export it in a separate module instead. Wdyt?

You mean export just the Subscription Client? Or all the subscription feature?

In the first case we can build a module like `@mercurius/subscription-tools` and use them directly in the `mercurius` part that manage the subscription.

I wont extract instead the feature completely.

Wdyt?

mcollina:
I would extract only the SubscriptionClient.

marco-ippolito:
> I would extract only the SubscriptionClient.

this is being done here https://github.com/mercurius-js/mercurius-subscription-client/pull/3

marco-ippolito:
@mcollina @codeflyer we can close this has has been resolved by https://github.com/mercurius-js/mercurius/pull/938

Issue #876, opened 2022-09-27
No way to operate with variables in hooks


npdev453:

mcollina:
Can you make an example?

npdev453:
Also Variables is very closes, and strange for me that it no passed inside hook:
https://github.com/mercurius-js/mercurius/blob/f6b6b858c24bc4bc919b25e9ac5cbf35df568d5d/index.js#L635

And slso no way to return it modified as Document do

(in apollo requestContext  (requestDidStart) its possible)

matijagaspar:
So if I understand you correctly @npdev453, you are saying variables should be returned separately ie: `app.graphql.addHook('preExecution', (schema, document, context, variables) => {` 
and possibly be able to return them modified:

```
app.graphql.addHook('preExecution', (schema, document, context, variables) => {

  return {
    modifiedVariables:  doSomethingWithVariables(variables)
  }
}
```

is that correct @npdev453 ?

Since variables are already included in the context via `context.req` would having a correctly typed context be enough?

npdev453:
> So if I understand you correctly @npdev453, you are saying variables should be returned separately ie: `app.graphql.addHook('preExecution', (schema, document, context, variables) => {` and possibly be able to return them modified:
> 
> ```
> app.graphql.addHook('preExecution', (schema, document, context, variables) => {
> 
>   return {
>     modifiedVariables:  doSomethingWithVariables(variables)
>   }
> }
> ```
> 
> is that correct @npdev453 ?
> 
> Since variables are already included in the context via `context.req` would having a correctly typed context be enough?

I think that it should  be chosen by library designer (an author @mcollina  or any active maintainer)

But in my option:
 1) using variables from context req is not stable solution and users code can be broken after some updates
 2) ability of accessing and modifying variables should be available because hook name is `preExecution` and `variables` is part of execution (also it can be safely logged in this place with Pino and redact function)
 3) context and context.req - should be typed as well, as you can absolutely know that it can be only passed from `fastify`

mcollina:
I'm kind of lost in all of this. By the look of https://github.com/mercurius-js/mercurius/issues/876#issuecomment-1259506419, this seemed an issue about lack of types and or docs.

```js
        app.graphql.addHook('preExecution', (schema, document, context) => {
                 console.log('variables', context.req.body.variables)
                 console.log('ops', readOpsNames(document)) // filtering for OperationDefinition selectionSet names
        });
```

[...]

PR #921, opened 2022-11-25
feat: Allow preExecution hook to modify variables

curtiswilkinson:
I saw discussions about this on https://github.com/mercurius-js/mercurius/issues/876, which mainly talked about **accessing** variables, but there is a comment that also mentions updating them.

I require this use-case, where hooks can modify the incoming variables based on logic or contacting external systems. See the test included with the PR for an example. 

This could have be achieved by putting the logic in a plugin earlier in the chain than Mercurius, but preventing double query parsing would be difficult, as the hooks need to be query-aware. 

I'll admit I don't have a complete understanding of this codebase so forgive me if there are any things I've missed here, it's mainly to frame the requirement I have and start a discussion. 

If there is any modifications you'd like here, let me know and I'd be happy to contribute; I can update any documentation as necessary, but wanted to wait until this feature was discussed.

### DIFF
--- a/docs/hooks.md
+++ b/docs/hooks.md
@@ -62,17 +62,24 @@ fastify.graphql.addHook('preValidation', async (schema, document, context) => {
 In the `preExecution` hook, you can modify the following items by returning them in the hook definition:
   - `document`
   - `schema`
+  - `variables`
   - `errors`
 
 Note that if you modify the `schema` or the `document` object, the [jit](./api/options.md#plugin-options) compilation will be disabled for the request.
 
 ```js
-fastify.graphql.addHook('preExecution', async (schema, document, context) => {
-  const { modifiedSchema, modifiedDocument, errors } = await asyncMethod(document)
+fastify.graphql.addHook('preExecution', async (schema, document, context, variables) => {
+  const {
+    modifiedSchema,
+    modifiedDocument,
+    modifiedVariables,
+    errors
+  } = await asyncMethod(document)
 
   return {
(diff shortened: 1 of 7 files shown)


mcollina (changes requested):
Can you document this new behavior?

curtiswilkinson:
Apologies @mcollina, I didn't expect the "re-review" button to drop the reviewers you requested, but apparently it does for some reason

mcollina (approved):
lgtm

docs/hooks.md
```diff
@@ -62,17 +62,24 @@ fastify.graphql.addHook('preValidation', async (schema, document, context) => {
...
   - `schema`
+  - `variables`
   - `errors`
 
 Note that if you modify the `schema` or the `document` object, the [jit](./api/options.md#plugin-options) compilation will be disabled for the request.
 
 ```js
-fastify.graphql.addHook('preExecution', async (schema, document, context) => {
-  const { modifiedSchema, modifiedDocument, errors } = await asyncMethod(document)
+fastify.graphql.addHook('preExecution', async (schema, document, context, variables) => {
```
jonnydgreen, line 71: Could we update the typings and add some type tests as well please?

jonnydgreen (commented):
Functionally all good - just a small comment to add these changes to the typings :)

docs/hooks.md
curtiswilkinson: Done I think! Let me know if there is anything else I've missed.

jonnydgreen (approved):
LGTM!

PR #922, opened 2022-11-29
refactor: removed events.on and use native api

marco-ippolito:
from issue https://github.com/mercurius-js/mercurius/issues/725.
events.on is native on node since added in: v13.6.0, v12.16.0 so I think there is no reason to keep this polyfill for 10.x

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const on = require('events.on')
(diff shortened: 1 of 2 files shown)


marco-ippolito:
@simoneb

simoneb (approved):
LGTM

simoneb:
@mcollina I'm going to merge this without further approvals as it's a fairly straightforward change

Issue #929, opened 2022-12-09
roadmap: migration of mercurius-gateway


marco-ippolito:
I'm creating a roadmap to keep track of the work on the migration of `mercurius-gateway` as npm dependency.

- [x]  create `mercurius-subscription-client` https://github.com/mercurius-js/mercurius-subscription-client/issues/2 @marco-ippolito 
- [x] release `mercurius-subscription-client` on npm https://www.npmjs.com/package/@mercuriusjs/subscription-client
- [x] change `subscription-client` on the gateway with the new package https://github.com/mercurius-js/mercurius-gateway/pull/49 @marco-ippolito 
- [x] remove `subscription-client` from `mercurius/lib` https://github.com/mercurius-js/mercurius/pull/938 @marco-ippolito supersedes: https://github.com/mercurius-js/mercurius/pull/920 @Mazuh 
- [x] release `mercurius-gateway` on npm @mcollina 
- [ ] remove `gateway` from `mercurius/lib` https://github.com/mercurius-js/mercurius/pull/942 @marco-ippolito

Mazuh:
> remove subscription-client from mercurius/lib

I'm going to work on it.

Mazuh:
I left a PR open: https://github.com/mercurius-js/mercurius/pull/930
but it's blocked until https://github.com/mercurius-js/mercurius/pull/916 gets merged.

mcollina:
@marco-ippolito could you finish this up?

marco-ippolito:
@mcollina should we remove the gateway and the subscription client from mercurius before or after releasing mercurius-gateway?

marco-ippolito:
@mcollina any blockers on the release on mercurius-gateway we can help? should we take over @codeflyer pr on removing gateway from mercurius?  what are the next steps?

mcollina:
I've released the new gateway. @marco-ippolito could you finish https://github.com/mercurius-js/mercurius/pull/916?

PR #930, opened 2022-12-12
chore: remove subscription-client from mercurius/lib

Mazuh:
relates to part of https://github.com/mercurius-js/mercurius/issues/929

I'm also removing the failing tests related to gateway because that's going to be removed too -- see https://github.com/mercurius-js/mercurius/pull/916 **(blocker)**


marco-ippolito:
This can be closed at it has been resolved by https://github.com/mercurius-js/mercurius/pull/938

PR #938, opened 2023-01-06
feat: replace subscription client with the standalone package

marco-ippolito:
As stated here: https://github.com/mercurius-js/mercurius/issues/929
Replacing subscription client with the standalone package.


mcollina (approved):
lgtm

PR #960, opened 2023-02-10
fix: add __resolveReference support

codeflyer:
This PR fixes is required to support properly the gateway.

### DIFF
--- a/index.js
+++ b/index.js
@@ -271,6 +271,10 @@ const plugin = fp(async function (app, opts) {
               ...fields[prop],
               ...resolver[prop]
             }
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
Should I just ship this very quick?

codeflyer:
> Should I just ship this very quick?

Yes, thanks. An I'll update the gateway as well

mcollina:
What needs to be changed in the gateway?

codeflyer:
> What needs to be changed in the gateway?

Update the package version, and fix a test.

Issue #959, opened 2023-02-06
Subscriptions in the federated service are not disposed correctly in gateway mode


hizo:
GraphQL subscriptions are not terminated correctly when there is gateway in the chain. Works fine when client connects directly to a backend service.

I have created a minimalistic reproduction with a service, gateway and frontend https://github.com/hizo/mercurius-gateway-subscriptions-repro . You can find all necessary information in the readme.

Mercurius version: 11.5.0
Note: I have also tried with latest v12 version and @mercuriusjs/gateway and @mercuriusjs/federation packages - the issue is still there, but to debug I needed to insert logs into the mercurius dependency of each gateway and federation packages. 

Actual result:
While it seems subscriptions are closed in gateway when there is a variable change, it's not promoted to the federated service and a subscription callback is executed as many times as variables changed.

Expected result:
Subscription termination is propagated to the federated service and subscription callback is triggered only for the active subscription.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

hizo:
@mcollina I'm afraid the reproduction is as far as I can get at the moment as we are just before a first release of a client project where this issue was noticed.

mcollina:
cc @marco-ippolito @simoneb

simoneb:
I think this is something that we (specifically @marco-ippolito) came across in the past when extracting [subscription-client](https://github.com/mercurius-js/mercurius-subscription-client) from this repo. 

If I remember correctly what we noticed was this:

- as the extraction had considerable risks of breaking changes, we tested the extracted subscription client in a "black box" fashion, meaning that ran a ws client against it. We did spot a strange behavior which seems to correspond to what @hizo is describing in this issue. @marco-ippolito initially captured it in [this comment](https://github.com/mercurius-js/mercurius-subscription-client/pull/3#issuecomment-1353257192)
- on the other hand, we tested the same against the original version of mercurius with the still-embedded subscription-client, and noticed the same behavior
- hence, we thought we'd caught a red herring and we were content that, although that behavior was somewhat weird, maybe we were missing something an in any case the before-after behavior was the same, so we were confident we'd not broken anything

This report makes me think that there was, and there still is, an issue in the way that the gateway handles subscriptions

mcollina:
I don't think this is a _new_ bug either, it's something that was present in the previous codebase as well.

simoneb:
@romulovitoi is working on this

[...]

PR #963, opened 2023-02-15
feat: include subscription id in onSubscriptionEnd hook

RomuloVitoi:
This is related to #959, it will be needed in the gateway to identify when a subscription has ended and which subscription was that.


mcollina (commented):
Can you add a test that verifies that the id matches whatever it was passed before?

mcollina (approved):
lgtm

Issue #966, opened 2023-02-20
Intercepting errors when resolving subscriptions


LewisYearsley:
Hey. We've been using `errorFormatter` Mercurius option to intercept and log errors. i.e.

```
app.register(mercurius, {
   // ...
    errorFormatter: (executionResult, context) => {
       // Send any errors to our logging service
    },
});
```

This has served us well for queries and mutations, but now we're using subscriptions.

In the case of subscriptions, we've noticed that the callback provided as `errorFormatter` doesn't seem to get run, is that expected?

I've come up with something that works, which is to move all of the logging logic to graphql hooks. i.e.

```
const handleErrorResponse = () => {
  // Do stuff with any errors
}

graphql.addHook("onResolution", handleErrorResponse);
graphql.addHook("onSubscriptionResolution", handleErrorResponse);
```

So to summarise:
- Is it expected that `errorFormatter` isn't executed for subscriptions?
- Is my alternative a good idea? if not, what is the recommended approach, please? 

Using Mercurius 11.5.0

mcollina:
thanks for reporting! The answer is that I don't know, and it's a bit hard to verify without a reproduction script.

This might well be a missing feature in Mercurius. If so, would you like to send a Pull Request?

LewisYearsley:
Thanks, @mcollina - I'll try to figure out what the behaviour should be and based on that I'll be happy to open a PR with the changes if required.

LewisYearsley:
Thanks again @mcollina for getting my changes in - we have them out in prod now and already showing results. We're good to close.

PR #967, opened 2023-02-21
Use error formatter for subscription GQL errors

LewisYearsley:
RE: https://github.com/mercurius-js/mercurius/issues/966

Currently, if there are graphql errors returned from a subscription in the standard "errors" property on a response - these do not get the opportunity to be passed through the potentially provided `errorFormatter`. This means any formatting and any side effects you may have in the error formatter are not run for these cases.

This PR applies the provided error formatter or merely the default error formatter as a fallback. 

NOTE:

I'm not 100% convinced on this change as is, my main thing is that in the case of mutation/query, we get to override the status code for the response in addition to the main payload. This doesn't apply to subscriptions and therefore that is thrown away in that case. It will work, but semantically _**may**_ be slightly weird. Open to feedback and suggestions of course.

### DIFF
--- a/lib/routes.js
+++ b/lib/routes.js
@@ -325,7 +325,8 @@ module.exports = async function (app, opts) {
       entityResolversFactory,
       subscriptionContextFn,
       keepAlive,
-      fullWsTransport
+      fullWsTransport,
+      errorFormatter
     })
   } else {
     app.route(getOptions)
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

mcollina:
This seems correct.

marco-ippolito (approved):
LGTM good job!

jonnydgreen (approved):
LGTM!

PR #979, opened 2023-04-14
Docs: document schema by http header (#913)

### LINKED ISSUE #913: Execute against different schemas based on request headers
I have a scenario where based on a request header coming from the client, the query should be executed against a completely seperate GraphQL Schema. 

`req.header.schema === 'A'` -> execute against `schema A`
`req.header.schema === 'B'` -> execute against `schema B`

I've handled this situation in a number of other GraphQL implementations, and was interested in using Mercurius if it's straight-forward. 

Apologies if this is an obvious question or if it's more related to fastify (which I admittedly have very little experience with). I've done some digging and couldn't see a straight-forward way without doing something very hand-rolled. 

I'd appreciate any guidance if you have any! I'd love to use Mercurius, as the benchmarking I've done has been **very** positive.

brainrepo:
closes #913 
This PR explains exposing two different mercurius instances to the same route using the fastify/find-my-way custom constraints.

- [x] Add a working example to the examples folder
- [x] Add a new paragraph to the FAQ section

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -29,3 +29,125 @@ app.register(mercurius, {
   validationRules: process.env.NODE_ENV === 'production' && [NoSchemaIntrospectionCustomRule],
 });
 ```
+
+## Execute against different schemas based on request headers
+
+Sometimes we may face the need to present a scheme that varies depending on specific situations.
+To accomplish this need we can use one powerful fastify/find-my-way feature called **Custom Constraints**.
+
+https://www.fastify.io/docs/latest/Reference/Routes/#asynchronous-custom-constraints
+
+> Fastify supports constraining routes to match only certain requests based on some property of the request, like the Host header, or any other value via find-my-way constraints.
+
+We can then create two mercurius instances that expose the two different schemas and use the constraint on the header to drive the request to one or other mercurius instance.
+
+### 1. Create the constraint and initialize the fastify instance
+```js
+const Fastify = require('fastify')
+const mercurius = require('..')
+
+// Define the constraint custom strategy
+const schemaStrategy = {
+  name: 'schema',
+  storage: function () {
+    const handlers = {}
+    return {
+      get: (type) => { return handlers[type] || null },
+      set: (type, store) => { handlers[type] = store }
+    }
+  },
+  deriveConstraint: (req, ctx) => {
+    return req.headers.schema
+  },
+  validate: () => true,
+  mustMatchWhenDerived: true
+}
+
+// Initialize fastify
+const app = Fastify({ constraints: { schema: schemaStrategy } })
+```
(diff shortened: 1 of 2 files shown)


brainrepo:
@simoneb can you give it a look pls?

simoneb (commented):
nice finding, I didn't know this was possible. although this may work, to me it seems a little hack-ish and I think we should try to come up with something better.

the issues I see at the moment are:

- the fastify docs discourage the creation of custom constraints because of the performance impact on the router. maybe by inspiring from the built-in ones we can come up with an implementation without any performance implications?
- the fact that you're registering two different instances of mercurius may still be necessary, but it would be nice it there wasn't so much work involved. maybe there's a way to build something into mercurius which would allow to achieve this without so much code? e.g. if we solved the first issue above and mercurius supported constraints natively, then maybe we can make it work without having to register routes explicitly because mercurius would know that unless the constraint is matched it wouldn't have to run?

mcollina (approved):
Fastify is amazing from time to time ;)

brainrepo:
I am exploring alternative ways to solve this issue.
The idea is to avoid instantiating multiple Mercurius instances.
I developed an idea by looking at @mercurius-js/auth's schema filtering and pruning features. 
We can use the `preExecution` hook to change the schema on the fly for the current request.
Doing that, we have just one limit. If the schemas differ, the validation is done using the original (default ) schema, and the validation step fails.
We can avoid that by giving the `preValidation` hook the ability to change the schema like done in the `preExecution` hook.
@mcollina @simoneb, do you think this is feasible?

Here a simple example

```javascript
const app = Fastify()
const schema1 = makeExecutableSchema({
  typeDefs: ...
  resolvers: ... 
})
const schema2 = makeExecutableSchema({
  typeDefs: ...
  resolvers: ... 
})

app.register(mercurius, {
  schema: schema1,
})

app.ready(err => {
  if (err) throw err
  app.graphql.addHook('preExecution', (schema, document, context) => {

    if (context.reply.request.headers.schema === "schema2") {
      return { schema: schema2}
    }

    return { schema }
  })
})
``` 

Not sure which is the impact for loaders and jit. Do you have any clue?

mcollina:
I think using the constraints is the best way to solve this. Adding support for multiple schemas in Mercurius will make the code harder to read, add a bunch if edge case we don't know about, and increase the maintenance cost.

simoneb:
@mcollina sounds good, thanks for the feedback. @brainrepo let's go ahead and document this option for the time being, while we think about other options to do this. Based on Matteo's feedback, implementing support for this within mercurius may not be a compelling option, on which case we may consider building a plugin instead, which hides away the implementation details.

brainrepo:
Thanks @mcollina , @simoneb . 
I will merge and create a plugin to hide the implementation details.

PR #982, opened 2023-04-18
docs: custom directive examples and docs page

### LINKED ISSUE #874: How to implement custom graphql directives with mercurius, is there a guide or minimal example?
How to implement custom graphql directives with mercurius, is there a guide or minimal example?

brainrepo:
This pr adds a doc page and an example of implementing a custom directive.
After a few iterations, to keep this example compact and avoid working with the AST, I decided to use two libraries, `@graphql-tools/schema` and `@graphql-tools/utils`.

I also added some comments in the example file, but I am unsure whether to keep them.

closes #874


simoneb (commented):
let's call this redaction rather than censorship :) See e.g. bottom of this page https://www.fastify.io/docs/latest/Reference/Logging/

docs/custom-directive.md
```diff
@@ -0,0 +1,85 @@
+# Custom directive
+A GraphQL directive is a special syntax used to provide additional information to the GraphQL execution engine about how to process a query, mutation, or schema definition. For example, directives can be used to modify the behaviour of fields, arguments, or types in your schema. 
+
+A custom directive is composed of 2 parts:
+- schema definitions
+- schema transformer
+
+## Schema Definition
+It is the syntax used to describe a custom directive within a GraphQL schema.
+To define a custom directive, you must use the directive keyword, followed by its name, arguments (if any), and the locations where it can be applied.
```
simoneb, line 10: I'd link to some docs explaining what possible locations there are besides FIELD_DEFINITION

docs/custom-directive.md
```diff
@@ -0,0 +1,85 @@
...
+It is the syntax used to describe a custom directive within a GraphQL schema.
+To define a custom directive, you must use the directive keyword, followed by its name, arguments (if any), and the locations where it can be applied. 
+
+```
+directive @censorship(find: String) on FIELD_DEFINITION
+```
+
+## Schema transformer
+
+A schema transformer is a function that takes a GraphQL schema as input and modifies it somehow before returning the modified schema.
```
simoneb, line 18: I'd clarify that this is the actual implementation of what the directive does

docs/custom-directive.md
```diff
@@ -0,0 +1,85 @@
...
+const censorshipSchemaTransformer = (schema) => mapSchema(schema, {
+  // When parsing the schema we find a FIELD
+  [MapperKind.FIELD]: fieldConfig => {
+    // Get the directive information
+    const censorshipDirective = getDirective(schema, fieldConfig, "censorship")?.[0]
+    if (censorshipDirective) {
+      // Get the resolver of the field
+      const innerResolver = fieldConfig.resolve
+      // Extract the find property from the directive
+      const { find } = censorshipDirective
```
simoneb, line 32: what is the find property from the directive?

docs/custom-directive.md
```diff
@@ -0,0 +1,85 @@
...
+      // Define a new resolver for the field
+      fieldConfig.resolve = async (_obj, args, ctx, info) => {
+        // Run the original resolver to get the result
+        const document = await innerResolver(_obj, args, ctx, info)
+        // Apply censorship only if context censored is true
+        if (!ctx.censored) {
+          return document
+        }
```
simoneb, line 40: for the sake of a simple example, I'd not mention the ability to configure the behavior of the directive via context. It is still useful to know that you can do that with the context though, so maybe we should first show the basic example, then mention that you can do more advanced things by accessing the context and making the behavior of the directive configurable

docs/custom-directive.md
```diff
@@ -0,0 +1,85 @@
...
+    # Define the directive schema
+    directive @censorship(find: String) on FIELD_DEFINITION
+    
+    type Document {
+      id: String!
+      text: String! 
+    }
+
+    type Query {
+      document: Document @censorship(find: "password")
```
simoneb, line 63: ah ok I see now what "find" is. I think we should keep this simpler. how about we apply the directive on the field we want to redact, rather than on a parent and specify the text to find and replace with? the example you've come up with is not realistic, in most cases you don't know what text you want to redact, but you know the field that may contain that text and you want to redact the whole field. so in summary: apply the directive on a field and redact the whole field

simoneb (changes requested):

brainrepo:
Thank you, All your comments make sense, I am going to implement your suggestions.

docs/custom-directive.md
```diff
@@ -0,0 +1,130 @@
...
+A GraphQL directive is a special syntax used to provide additional information to the GraphQL execution engine about how to process a query, mutation, or schema definition. 
+For example, directives can be used to modify the behaviour of fields, arguments, or types in your schema. 
+
+A custom directive is composed of 2 parts:
+- schema definitions
+- transformer
+
+## Schema Definition
+
+**Let's explore the custom directive creation process by creating a directive to redact some fields value hiding a specific word.**
```
simoneb, line 15: why bold this sentence?

docs/custom-directive.md
```diff
@@ -0,0 +1,130 @@
...
+    }`
+```
+
+To define a custom directive, we must use the directive keyword, followed by its name prefixed by a `@`, the arguments (if any), and the locations where it can be applied. 
+
+```
+directive @redact(find: String) on FIELD_DEFINITION
+```
+
+The directive can be applied in multiple locations.
```
simoneb, line 40: oh no don't list them all here, link to the docs!

examples/custom-directive.js
```diff
@@ -0,0 +1,69 @@
...
+    }
+  }
+})
+
+// Register mercurius and run it
+app.register(mercurius, {
+  schema: redactionSchemaTransformer(schema),
+  context: (request, reply) => {
+    return {
+      redaction: true
```
simoneb, line 63: let's call this redact

simoneb (commented):
Looks better but I still think that the use case you're using is not fitting. Maybe It wasn't clear in my previous comment but you can't redact something that you don't know what will be. You will need to redact the whole field instead. So I imagine the @redact directive to be applied to all fields of a type which possibly contain sensitive information, and they will redact the content of the whole field. Does this make sense now?

docs/custom-directive.md
```diff
@@ -0,0 +1,127 @@
...
+    }`
+```
+
+To define a custom directive, we must use the directive keyword, followed by its name prefixed by a `@`, the arguments (if any), and the locations where it can be applied.
+
+```
+directive @redact(find: String) on FIELD_DEFINITION
+```
+
+According to the graphql specs the directive can be applied in multiple locations. See the list on https://spec.graphql.org/October2021/#sec-Type-System.Directives.
```
simoneb, line 38: display the link using markdown syntax, not a raw inline link please

docs/custom-directive.md
```diff
@@ -0,0 +1,127 @@
...
+
+```js
+const { mapSchema, getDirective, MapperKind } = require('@graphql-tools/utils')
+
+const redactionSchemaTransformer = (schema) => mapSchema(schema, {
+  // When parsing the schema we find a FIELD
+  [MapperKind.FIELD]: fieldConfig => {
+    // Define the regexp
+    const PHONE_REGEXP = /(?:\+?\d{2}[ -]?\d{3}[ -]?\d{5}|\d{4})/g;
+    const EMAIL_REGEXP = /([^\s@])+@[^\s@]+\.[^\s@]+/g
```
simoneb, line 57: these regexps don't need to live inside this function

docs/custom-directive.md
```diff
@@ -0,0 +1,127 @@
...
+```js
+app.register(mercurius, {
+  schema: newSchema,
+  graphiql: true,
+})
+```
+
+
+
+
```
simoneb, line 124: what's all these new lines?

examples/custom-directive.js
```diff
@@ -0,0 +1,76 @@
...
+    `,
+  resolvers
+})
+
+const redactionSchemaTransformer = (schema) => mapSchema(schema, {
+  [MapperKind.OBJECT_FIELD]: fieldConfig => {
+    const PHONE_REGEXP = /(?:\+?\d{2}[ -]?\d{3}[ -]?\d{5}|\d{4})/g
+    const EMAIL_REGEXP = /([^\s@])+@[^\s@]+\.[^\s@]+/g
+    const redactDirective = getDirective(schema, fieldConfig, 'redact')?.[0]
+    if (redactDirective) {
```
simoneb, line 44: add some blank lines here and there for readability please

simoneb (approved):
some comments, but looking better now, thanks

brainrepo:
I implemented the suggestions

simoneb (approved):

simoneb:
@mcollina this looks good to me but I'd like another pair of eyes before we merge this

mcollina (changes requested):
Please add it to https://github.com/mercurius-js/mercurius/blob/master/docsify/sidebar.md

docs/custom-directive.md
```diff
@@ -0,0 +1,125 @@
...
+Every directive needs its transformer.
+A transformer is a function that takes an existing schema and applies the modifications to the schema and resolvers.
+
+To simplify the process of creating a transformer, we use the `mapSchema` function from the `@graphql-tools` library.
+
+The `mapSchema` function applies each callback function to the corresponding type definition in the schema, creating a new schema with the modified type definitions. The function also provides access to the field resolvers of each object type, allowing you to alter the behaviour of the fields in the schema.
+
+```js
+const { mapSchema, getDirective, MapperKind } = require('@graphql-tools/utils')
```
mcollina, line 50: We should document somewhere what version of graphql-tools this guide refers to `graphql-tools`.

I would also prefer to not use graphql-tools here or show both examples.

simoneb: I agree with Mauro, we have discussed in detail and replicating the features coming from the graphql-tools library for the sake of this example is just not worth it, and quite frankly in general it's not worth it. Nevertheless since this issue comes up quite often, we'll come up with a couple of ideas that we'll open as issues in this repo:

- one specific to this request, a way built-into mercurius (or a plugin) to make it easier to define and attach custom directives
- one generic about creating a new library alternative to graphql-tools with similar features

brainrepo: Yes, I have provided a more detailed argument on this issue.

mcollina (approved):
lgtm

PR #986, opened 2023-04-25
fix mutation over websocket/subscription

tomi-bigpi:
Fixes issue where mutations are not allowed over web sockets (subscriptions).

See: https://github.com/mercurius-js/mercurius/discussions/977

### DIFF
--- a/index.js
+++ b/index.js
@@ -471,7 +471,7 @@ const plugin = fp(async function (app, opts) {
       document = cached.document
     }
 
-    if (reply && reply.request.raw.method === 'GET' && reply.request.headers.upgrade?.toLowerCase() !== 'websocket') {
+    if (reply && reply.request.raw.method === 'GET' && !reply.request.ws) {
       // let's validate we cannot do mutations here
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -471,7 +471,7 @@ const plugin = fp(async function (app, opts) {
       document = cached.document
     }
 
-    if (reply && reply.request.raw.method === 'GET') {
+    if (reply && reply.request.raw.method === 'GET' && reply.request.headers.upgrade?.toLowerCase() !== 'websocket') {
```
mcollina, line 474:
```suggestion
    if (reply && reply.request.raw.method === 'GET' && !reply.request.ws) {
```

mcollina: This should work, check it out!

tomi-bigpi: Thanks, it worked with manual testing, so I updated the corresponding test to set that flag on the request like `fastify-websocket` does.

mcollina (approved):
lgtm

PR #987, opened 2023-04-26
Update to @fastify/websocket v8

mcollina:
Updates to `@fastify/websocket` v8 to address a leak and crash in certain condition. This requires a minimum version of Fastify of v4.16.0 (just released), hence the major update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,7 @@
     "autocannon": "^7.3.0",
     "concurrently": "^8.0.1",
     "docsify-cli": "^4.4.3",
-    "fastify": "^4.0.0",
+    "fastify": "^4.16.0",
     "graphql-ws": "^5.11.2",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.1.3",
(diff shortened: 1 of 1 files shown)


mcollina:
cc @simoneb @codeflyer @marco-ippolito @jonnydgreen wdyt?

This fixes a horrible crash in Platformatic, so I'm keen on shipping this sooner rather than later.

simoneb:
I'm afraid I'm missing some context but it looks like this is part of a group of upgrades to various packages that need to go together (I see the major in fastify-ws, a minor followed by various patches in fastify, fastify-json-stringify and compiler, ...). There's an overarching plan I'm missing, but I guess we're good as I wouldn't be in any better position than you are to validate this change.

simoneb (approved):

mcollina:
There is no grand plan, just this bug to be fixed https://github.com/fastify/fastify-websocket/issues/249. That required a new hook in Fastify and and a semver-major change jn fastify-websocket.

The change in fast-json-stringify was 100% unrelated as well as all the problems in v4.16.3.

marco-ippolito (approved):

PR #992, opened 2023-05-11
schema option: Better error handling when providing arrays with non-string schemas

### LINKED ISSUE #984: Better error messages on passing array of executable schemas.
In Mercurius' schema option, we can pass schemas as either a string or an executable schema. 

If we pass them as a string, we can also pass an array of schemas in the string type, in which case Mercurius will merge them. 

However, if we pass them as an executable schema, we can only pass one. 
Currently, there is no check to prevent the passing of an array of executable schemas. 
If we pass two executable schemas, we get an error.

Possible solutions:

- improve error handling
- use schema stitching against the executable schemas to get just one executable schema

https://github.com/mercurius-js/mercurius/blob/b6ac05724b7715ba90d8686aa8218fc8de10ffd1/index.js#L148-L171

alcidesqueiroz:
closes #984

### DIFF
--- a/test/app-decorator.js
+++ b/test/app-decorator.js
@@ -1075,6 +1075,32 @@ test('extended Schema is an object', async t => {
   await app.ready()
 })
 
+test('array of schemas contains a non-string schema', async t => {
+  const app = Fastify()
+
+  app.register(GQL, {
+    schema: [makeExecutableSchema({
+      typeDefs: `
+          type Query {
+            add(x: Int, y: Int): Int
+          }
+        `,
+      resolvers: {
+        Query: {
+          add: async (_, { x, y }) => x + y
+        }
+      }
+    }
+    )]
+  })
+
+  try {
+    await app.ready()
+  } catch (error) {
+    t.equal(error.message, 'Invalid options: when providing an array to the "schema" option, only string schemas are allowed')
+  }
+})
+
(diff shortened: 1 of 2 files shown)


test/app-decorator.js
```diff
@@ -1075,6 +1075,32 @@ test('extended Schema is an object', async t => {
...
+    }
+    )]
+  })
+
+  try {
+    await app.ready()
+  } catch (error) {
+    t.equal(error.message, 'Invalid options: when providing an array to the "schema" option, only string schemas are allowed')
+  }
+})
```
simoneb, line 1102: this is not the right way to test an exception. if the exception is not raised, this test will succeed

simoneb (changes requested):

test/app-decorator.js
alcidesqueiroz: Makes sense. I had followed the test pattern in use in this project. For example, in this same test file, you'll find this pattern on lines 336, 1006 and 1025. But, on a second thought, the test would in fact succeed if the exception isn't raised. To prevent this, I updated the test, adding a `t.plan(1)` call.

mcollina (approved):
lgtm

mcollina:
@simoneb PTAL

simoneb (approved):
LGTM

PR #996, opened 2023-05-16
Refactoring tests that check if an error has been thrown

### LINKED ISSUE #995: Several unit tests checking for errors would pass even if the error isn't thrown
Several unit tests that check if a given error has been thrown are using the following approach:

```js
try {
  await app.ready()
} catch (error) {
  t.equal(error.message, 'The expected error message')
}
```

The problem here is that, without using `t.plan`, these tests would pass even if the error wasn't thrown. And most tests using this pattern are not calling `t.plan`. My suggestion is to replace these try/catch occurrences with `t.rejects`, making the tests simpler and preventing these false positives.

FYI: @simoneb

alcidesqueiroz:
I refactored almost 30 tests that were using try/catch to check if an error had been thrown. There are two problems with this approach:
- It's unnecessarily complex. We can easily replace these try/catch-based assertions with `t.rejects`.   
- But the worst issue is: without using `t.plan()`, these tests would pass even if the error wasn't thrown. And most of these tests weren't using `t.plan`. In fact, 4 tests were passing without the exceptions they test for being thrown. Out of these 4, I was able to confirm that at least 1 was obsolete, so I've removed it. The other 3 tests remain untouched (`test/app-decorator.js:476`, `test/loaders:529` and `test/routes:964`).

closes #995 

FYI: @simoneb

### DIFF
--- a/test/app-decorator.js
+++ b/test/app-decorator.js
@@ -246,11 +246,8 @@ test('replaceSchema (clearing cache)', async (t) => {
   )
 
   query = '{ subtract(x: 4, y: 2) }'
-  try {
-    await app.graphql(query)
-  } catch (err) {
-    t.equal(err.errors[0].message, 'Cannot query field "subtract" on type "Query".')
-  }
+
+  await t.rejects(app.graphql(query), { errors: [{ message: 'Cannot query field "subtract" on type "Query".' }] })
 })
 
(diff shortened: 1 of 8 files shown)


mcollina (approved):
lgtm

simoneb (approved):

PR #1025, opened 2023-09-26
bugfix: handle if errors is not an array

### LINKED ISSUE #1024: Occasional "error.originalError.errors.map is not a function" errors
I very sporadically get errors in my project's tests `"error.originalError.errors.map is not a function"` which I've traced back to [here](https://github.com/mercurius-js/mercurius/blob/master/lib/errors.js#L49) in mercurius. could easily do a PR do add `Array.isArray()` there but this might hint at something else (me?) passing the wrong args to `defaultErrorFormatter`

ashley-conpago:
closes #1024 

- check that `error.originalError.errors` is an array before mapping it.
- if not an array, but is defined, try to handle it anyway.


mcollina (changes requested):

PR #1031, opened 2023-10-05
Remove all listeners on close

aldotms70:
This patch calls the removeListener on emitter for all subscribed topics. It is important to avoid memory leak.
When using MQEmitterRedis is most important to unsubscribe pattern on redis server.


mcollina (approved):
lgtm

Issue #1003, opened 2023-06-26
Playground endpoint is not dynamic


astghikxachatryan:
`const url = `${window.location.protocol}//${host}${window.GRAPHQL_ENDPOINT}``

Playground doesn't work on URLs like this [https://foo.com/comments/graphiql](https://foo.com/comments/graphiql).
[https://foo.com/comments/]( https://foo.com/comments/) is the API getaway URL that routes to our comments service and comments service is based on Mercurius. 
In this case [code](https://github.com/mercurius-js/mercurius/blob/master/static/main.js) just ignores the part after host(in my case **comments**) 

And playground call to wrong qraphql url.  It calls [https://foo.com/graphql](https://foo.com/graphql) instead of [https://foo.com/comments/graphql](https://foo.com/comments/graphql)

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

glg-satish-tripathi:
@mcollina we are deploying graphql api and using this package and we are also facing the same issue. I have created the [fix ](https://github.com/mercurius-js/mercurius/pull/1037) . Please review and approve it.

PR #1037, opened 2023-10-19
make url dynamic so even if the graphiql is deployed on baseUrl it can work.

glg-satish-tripathi:
fix - https://github.com/mercurius-js/mercurius/issues/1003 issue. 

The graphiql failed to load schema when it's been deployed on https://foo.com/bar/graphiql, it will look for the schema at https://foo.com/graphql but it's actually running on https://foo.com/bar/graphql, which is an invalid graphql endpoint as it is missing window.baseUrl. So we are addressing this in the PR. 

image

Thank you @ddai1  for help here.

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -97,11 +97,16 @@ function fetcherWrapper (fetcher, cbs = []) {
 
 function render () {
   const host = window.location.host
-
   const websocketProtocol = window.location.protocol === 'https:' ? 'wss:' : 'ws:'
-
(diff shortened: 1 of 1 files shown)


glg-satish-tripathi:
Thank you @ddai for help here.

glg-satish-tripathi:
@mcollina  please review and approve this, i am working on a graphql api and without these changes i can not deploy the setup. i am not an expert on javascript, so not very sure where unit test needs to be added if any.

mcollina (approved):
lgtm

PR #1041, opened 2023-10-25
bugfix: window.baseurl is not working in all the cases to fix this using window.location.pathname

glg-satish-tripathi:
bugfix:
- The `window.baseurl` is not working as expected as in some cases window.baseurl is undefined. 
- using `window.location.pathname` is better then using `window.baseurl` as in some cases the value is undefined.

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -101,8 +101,14 @@ function render () {
   let url = ''
   let subscriptionUrl = ''
   if (window.baseurl) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

glg-satish-tripathi:
@mcollina it's not auto merged in the actions run i see `Warning: Not a dependabot PR, skipping.`
https://github.com/mercurius-js/mercurius/actions/runs/6637638646/job/18032455908?pr=1041

PR #1050, opened 2023-11-08
feat: support passing options to graphql validate and parse

markrzen:
Support passing options to GraphQL's `validate` and `parse` functions.

Reasoning to allow configuration of these functions topics is fine tuning to prevent DDOS attacks.

This allows developers to set:
1. `validate`'s options of `maxErrors` to something more/less than the 100 default. This prevents an attacker from specifying 100 invalid fields, requiring CPU time generating validation errors.
2. `parse`'s options of `maxTokens` to a value. It has no default. This prevent an attacker from producing a overly high complexity query that requires a high amount of CPU time.

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -34,6 +34,9 @@
 - `loaders`: Object. See [defineLoaders](#appgraphqlextendschemaschema-appgraphqldefineresolversresolvers-and-appgraphqldefineloadersloaders) for more
   details.
 - `schemaTransforms`: Array of schema-transformation functions. Accept a schema as an argument and return a schema.
+- `graphql`: Object. Override options for graphql function that Mercurius utilizes.
+  - `parseOptions`: Object. [GraphQL's parse function options](https://github.com/graphql/graphql-js/blob/main/src/language/parser.ts)
+  - `validateOptions`: Object. [GraphQL's validate function options](https://github.com/graphql/graphql-js/blob/main/src/validation/validate.ts)
 - `graphiql`: boolean | string | Object. Serve
   [GraphiQL](https://www.npmjs.com/package/graphiql) on `/graphiql` if `true` or `'graphiql'`. Leave empty or `false` to disable.
   _only applies if `onlyPersisted` option is not `true`_
--- a/index.js
+++ b/index.js
@@ -78,6 +78,10 @@ const plugin = fp(async function (app, opts) {
   const queryDepthLimit = opts.queryDepth
   const errorFormatter = typeof opts.errorFormatter === 'function' ? opts.errorFormatter : defaultErrorFormatter
 
+  opts.graphql = opts.graphql || {}
+  const gqlParseOpts = opts.graphql.parseOptions || {}
+  const gqlValidateOpts = opts.graphql.validateOptions || {}
+
   if (opts.persistedQueries) {
     if (opts.onlyPersisted) {
       opts.persistedQueryProvider = persistedQueryDefaults.preparedOnly(opts.persistedQueries)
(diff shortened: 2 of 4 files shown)


markrzen:
@mcollina Can we get this on a reviewers queue? I wouldn't usually poke maintainers like this, but previous PRs seem to be getting reviewed within a few days. I wanted to make sure this didn't fall through the cracks.

markrzen:
@simoneb Any chance this is something you can review?

docs/api/options.md
```diff
@@ -34,6 +34,9 @@
 - `loaders`: Object. See [defineLoaders](#appgraphqlextendschemaschema-appgraphqldefineresolversresolvers-and-appgraphqldefineloadersloaders) for more
   details.
 - `schemaTransforms`: Array of schema-transformation functions. Accept a schema as an argument and return a schema.
+- `graphql`: Object. Override options for graphql function that Mercurius utilizes.
+  - `parseOptions`: Object. [GraphQL's parse function options](https://github.com/graphql/graphql-js/blob/main/src/language/parser.ts#L77-L133)
+  - `validateOptions`: Object. [GraphQL's validate function options](https://github.com/graphql/graphql-js/blob/main/src/validation/validate.ts#L44)
```
simoneb, line 39: I would avoid deep links to source code, as it can change any time. The alternative would be to use permalinks, but I'm not sure that's a good option either. Is there official documentation that we can point to instead?

mcollina (approved):
lgtm

index.js
```diff
@@ -428,9 +432,14 @@ const plugin = fp(async function (app, opts) {
       }
 
       try {
-        document = parse(source)
+        document = parse(source, gqlParseOpts)
       } catch (syntaxError) {
         try {
+          // Do not try to JSON.parse maxToken exceeded validation errors
+          if (gqlParseOpts.maxTokens && syntaxError.message === `Syntax Error: Document contains more that ${gqlParseOpts.maxTokens} tokens. Parsing aborted.`) {
```
simoneb, line 439: relying on an external library's error message to decide what to do doesn't feel very reliable

markrzen: Totally agreed. Sadly I could not find another option. Any suggestion?

I did protect against it's potential change in the unit tests below.

docs/api/options.md
markrzen: The published graphql docs do not contain option properties sadly. I could update them to a specific tag if that works for you?

simoneb: not many compelling options then. I'm not sure. if we point at master, those links may refer to wrong lines when those files change. if we use permalinks, the reference will keep existing but the real code might have changed. not sure what's best really. shall we link to the files without any line number perhaps? that would give us some confidence that those files will keep existing and leave it to the reader to figure out where to find the types

index.js
simoneb: I guess we don't have many options here either. What will happen is that when dependabot bumps this dependency, the unit test will break which will give us some clue of what needs fixing.

docs/api/options.md
markrzen: We could just enumerate the literal properties, rather than linking off. It does mean our future selfs will need to go back to the docs and confirm when things have been added/removed.

Not sure there is a good option, other than pushing for these functions to be documented more fully in GraphQL library.

Happy to change to without line numbers if that makes anyone feel more comfortable.

simoneb: No line numbers is the compromise I'd go with, but happy for somebody else to chime in too

markrzen: Done! :)

simoneb (approved):
LGTM, thanks for this!

markrzen:
@simoneb @mcollina Thanks y'all. What sort of release schedule are y'all doing? Any expectations I should have?

Issue #1047, opened 2023-11-05
Types issue with Typescript module resolution node16


iamchathu:
I created a new typescript respository with following Typescript config.

```json
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "display": "Node 20",
  "_version": "20.1.0",
  "compilerOptions": {
    "lib": ["es2023"],
    "module": "node16",
    "target": "es2022",
    "strict": true,
    "esModuleInterop": true,
    "skipLibCheck": true,
    "forceConsistentCasingInFileNames": true,
    "moduleResolution": "node16",
    "isolatedModules": true
  }
}
```

The following in the index.ts file as the example.

```ts
import fastify from 'fastify';
import mercurius, { IResolvers } from 'mercurius';

export const app = fastify({});

const schema = `
  type Query {
    add(x: Int, y: Int): Int
  }
`;

const resolvers: IResolvers = {
  Query: {
    add: async (_, { x, y }) => x + y,
  },
};

app.register(mercurius, {
  schema,
  resolvers,
});

app.get('/', async function (req, reply) {
  const query = '{ add(x: 2, y: 2) }';
  return reply.graphql(query);
});

app.listen({ port: 3000 });

```

The dependencies versions looks like following

```
"fastify": "^4.24.3",
"graphql": "^16.8.1",
"mercurius": "^13.2.2"
"typescript": "^5.2.2"
```

The error when transpiling the Typescript as the following,

```sh
src/index.ts:18:14 - error TS2769: No overload matches this call.
  Overload 1 of 3, '(plugin: FastifyPluginCallback<{ schema: string; resolvers: IResolvers<any, MercuriusContext>; }, RawServerDefault, FastifyTypeProvider, FastifyBaseLogger>, opts?: FastifyRegisterOptions<...> | undefined): FastifyInstance<...> & PromiseLike<...>', gave the following error.
    Argument of type 'typeof import("/Users/Chathu/Workspace/app-v3/node_modules/.pnpm/mercurius@13.2.2_graphql@16.8.1/node_modules/mercurius/index")' is not assignable to parameter of type 'FastifyPluginCallback<{ schema: string; resolvers: IResolvers<any, MercuriusContext>; }, RawServerDefault, FastifyTypeProvider, FastifyBaseLogger>'.
[...]

PR #1056, opened 2023-11-26
Support nodenext

jonnydgreen:
Closes: #1047 

 - I have taken the code as is (to de-risk) and just moved the root level types into the declared namespace, other key bits are detailed as follows:

```typescript
type Mercurius = typeof mercurius

declare namespace mercurius {
  // ...
  
  export const mercurius: Mercurius
  export { mercurius as default }
}

declare module "fastify" {
  // ...
}

declare function mercurius
  (
    instance: FastifyInstance,
    opts: mercurius.MercuriusOptions
  ): void;

export = mercurius;
```

 - All the types tests pass and have added some other tests where needed
 - Have taken some inspiration from the fastify repo and other plugins for consistency
 - Have also tested against this repo and all appears to be okay: https://github.com/iamchathu/mercurius-ts-example

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -16,487 +16,595 @@ import {
   ParseOptions,
 } from "graphql";
 import { SocketStream } from "@fastify/websocket"
-import { IncomingMessage, IncomingHttpHeaders, OutgoingHttpHeaders } from "http";
+import { IncomingMessage, OutgoingHttpHeaders } from "http";
 import { Readable } from "stream";
 
-export interface PubSub {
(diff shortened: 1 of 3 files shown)


jonnydgreen:
Opening an early PR to get your initial thoughts! Essentially, it's just a lift and shift - but keen to get feedback :)

mcollina (approved):
lgtm, this should be good to go

jonnydgreen:
Cool - thanks! Just ran some final testing and all good to go :)

PR #1080, opened 2024-02-07
Feat: add the ability to add in additional route properties

barelyhuman:
# Changes 
ability to add and modify route properties to modify how the routes might behave. One example is to be able to create a schema split which https://github.com/nearform/mercurius-dynamic-schema/ tries to achieve. 

## Reasoning
Helps close https://github.com/nearform/mercurius-dynamic-schema/issues/4

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -89,7 +89,7 @@
 - `allowBatchedQueries`: Boolean. Flag to control whether to allow batched queries. When `true`, the server supports recieving an array of queries and returns an array of results.
 
 - `compilerOptions`: Object. Configurable options for the graphql-jit compiler. For more details check https://github.com/zalando-incubator/graphql-jit
-- `additionalRouteProps`: Object. Takes similar configuration to the Route Options of Fastify. Usecases include being able to add constraints, modify validation schema, increase the `bodyLimit`, etc. You can read more about the possible values on [fastify.dev's Routes Options](https://fastify.dev/docs/latest/Reference/Routes/#options)
+- `additionalRouteOptions`: Object. Takes similar configuration to the Route Options of Fastify. Use cases include being able to add constraints, modify validation schema, increase the `bodyLimit`, etc. You can read more about the possible values on [fastify.dev's Routes Options](https://fastify.dev/docs/latest/Reference/Routes/#options)
 
 #### queryDepth example
 
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -89,6 +89,7 @@
 - `allowBatchedQueries`: Boolean. Flag to control whether to allow batched queries. When `true`, the server supports recieving an array of queries and returns an array of results.
 
 - `compilerOptions`: Object. Configurable options for the graphql-jit compiler. For more details check https://github.com/zalando-incubator/graphql-jit
+- `additionalRouteProps`: Object. Takes similar configuration to the Route Options of Fastify. Usecases include being able to add constraints, modify validation schema, increase the `bodyLimit`, etc. You can read more about the possible values on [fastify.dev's Routes Options](https://fastify.dev/docs/latest/Reference/Routes/#options)
 
 #### queryDepth example
 
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,6 +3,7 @@ import {
   FastifyReply,
   FastifyRequest,
   FastifyInstance,
+  RouteOptions
 } from "fastify";
 import {
   DocumentNode,
@@ -487,6 +488,18 @@ declare namespace mercurius {
      * receive an array of responses within a single request.
      */
     allowBatchedQueries?: boolean;
+
+    /**
+     * Customize the graphql routers initialized by mercurius with 
+     * more fastify route options. 
+     * 
+     * Usecases:
+     * - Add more validation 
+     * - change how the schema behaves 
+     * - Hook on the request's life cycle
+     * - Increase body size limit for larger queries
+    */
+    additionalRouteProps?:Omit<RouteOptions,"handler" | "wsHandler">
   }
   
   export type MercuriusOptions = MercuriusCommonOptions & (MercuriusSchemaOptions)
--- a/index.d.ts
+++ b/index.d.ts
@@ -499,7 +499,7 @@ declare namespace mercurius {
      * - Hook on the request's life cycle
      * - Increase body size limit for larger queries
     */
-    additionalRouteOptions?:Omit<RouteOptions,"handler" | "wsHandler">
+    additionalRouteOptions?:Omit<RouteOptions,"handler"|"wsHandler"|"method"|"url">
   }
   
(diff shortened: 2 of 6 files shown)


simoneb:
For context, we tried various solutions in https://github.com/nearform/mercurius-dynamic-schema to extend core's functionality to achieve what we needed, and eventually realized that with this simple change in mercurius we'd have to get rid of unnecessary complications (and duplications) in our core. Hence we're proposing this change.

lib/routes.js
```diff
@@ -200,7 +200,8 @@ module.exports = async function (app, opts) {
     persistedQueryProvider,
     allowBatchedQueries,
     keepAlive,
-    fullWsTransport
+    fullWsTransport,
+    additionalRouteProps
```
jonnydgreen, line 204: Hello! Worth adding some docs and types for this do you think?

index.js
```diff
@@ -210,7 +210,8 @@ const mercurius = fp(async function (app, opts) {
       entityResolversFactory: undefined,
       subscriptionContextFn,
       keepAlive,
-      fullWsTransport
+      fullWsTransport,
+      additionalRouteProps: opts.additionalRouteProps
```
jonnydgreen, line 214: Hello! Worth adding some docs and types for this do you think?

simoneb: Yes indeed

mcollina (changes requested):
can you add docs and tests?

barelyhuman:
> can you add docs and tests?

Will do

test/routes.js
```diff
@@ -2187,3 +2187,100 @@ test('if ide has plugin set, serve config.js with the correct endpoint', async (
...
+
+test('GET graphql endpoint, additionalRouteProps should ignore custom handler', async (t) => {
+  const app = Fastify()
+  const schema = `
+    type Query {
+      add(x: Int, y: Int): Int
+    }
+  `
+
+  let executed = false
```
simoneb, line 2259: for this type of async tests, using t.plan is preferred. in my opinion this is also the only case in which you want to use t.plan, but in such cases, I prefer to use it versus setting a flag like you did

test/routes.js
```diff
@@ -2187,3 +2187,100 @@ test('if ide has plugin set, serve config.js with the correct endpoint', async (
...
+
+  app.register(GQL, {
+    schema,
+    additionalRouteProps: {
+      // @ts-ignore
```
simoneb, line 2264: why this ignore?

lib/routes.js
```diff
@@ -214,6 +215,12 @@ module.exports = async function (app, opts) {
     notSupportedError
   } = persistedQueryProvider || {}
 
+  const normalizedRouteProps = { ...additionalRouteProps }
+  if (normalizedRouteProps.handler || normalizedRouteProps.wsHandler) {
+    normalizedRouteProps.handler = undefined
+    normalizedRouteProps.wsHandler = undefined
+  }
```
simoneb, line 222: why this change?

simoneb (changes requested):
See inline comments. Plus, we should also include this in the documentation https://mercurius.dev/#/docs/api/options?id=mercurius

test/routes.js
barelyhuman: The types omit it so added it for my editors tslint, accidentally committed it 🤦‍♂️

lib/routes.js
barelyhuman: The post handler adds a callback in which case fastify errors saying that there's 2 handlers. so normalised it to avoid even considering the handler provided from the options

test/routes.js
barelyhuman: wouldn't a t.plan wait for the handler to execute? 

i don't want it to execute here , hence the flag

simoneb: t.plan would cause the test runner to wait to finish the test until all planned assertions have run, which is what you're trying to accomplish here

simoneb: ok fine, as clarified, you're testing that the handler has not executed. this solution works then

docs/api/options.md
```diff
@@ -89,6 +89,7 @@
 - `allowBatchedQueries`: Boolean. Flag to control whether to allow batched queries. When `true`, the server supports recieving an array of queries and returns an array of results.
 
 - `compilerOptions`: Object. Configurable options for the graphql-jit compiler. For more details check https://github.com/zalando-incubator/graphql-jit
+- `additionalRouteProps`: Object. Takes similar configuration to the Route Options of Fastify. Usecases include being able to add constraints, modify validation schema, increase the `bodyLimit`, etc. You can read more about the possible values on [fastify.dev's Routes Options](https://fastify.dev/docs/latest/Reference/Routes/#options)
```
simoneb, line 92:
```suggestion
- `additionalRouteProps`: Object. Takes similar configuration to the Route Options of Fastify. Use cases include being able to add constraints, modify validation schema, increase the `bodyLimit`, etc. You can read more about the possible values on [fastify.dev's Routes Options](https://fastify.dev/docs/latest/Reference/Routes/#options)
```

I was also thinking, since the right term in the fastify docs is route options, why don't we call it options instead of props?

index.d.ts
```diff
@@ -487,6 +488,18 @@ declare namespace mercurius {
...
+     * Customize the graphql routers initialized by mercurius with 
+     * more fastify route options. 
+     * 
+     * Usecases:
+     * - Add more validation 
+     * - change the schema structure 
+     * - Hook on the request's life cycle
+     * - Increase body size limit for larger queries
+    */
+    additionalRouteProps?:Omit<RouteOptions,"handler" | "wsHandler">
```
simoneb, line 502: let's add types tests for this

docs/api/options.md
barelyhuman: Makes sense

index.d.ts
barelyhuman: There's no type tests for the options in place though, do we want it to be a separate PR for all of them ?

simoneb: I mean, just add a test which checks that you can supply route options (and which ones perhaps) here https://github.com/mercurius-js/mercurius/blob/master/test/types/index.ts

simoneb (approved):

simoneb:
@mcollina if you could take another look please

barelyhuman:
@mcollina whenever you get time

mcollina (approved):
lgtm

simoneb:
Can somebody do a release? Speaking of which, is there an established way to do releases in this package (or the rest of the ecosystem)? If not, we shall consider introducing one @Eomm

Eomm:
Cannot release - I'm not part of the `mercurius-js` org 🙌🏼

simoneb:
@Eomm you should be then! In any case I was looping you in mostly to check if you had settled on anything with Fastify

simoneb:
Can anybody do a release of mercurius? We are waiting for this feature to be released to start using it in one of our repos.

PR #1088, opened 2024-03-13
feat: make request available in context by default

adtennant:
I know this is very easy to define a custom context for, but it's annoying to have to do it every time and this is a closer match to standard Fastify handlers.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -30,6 +30,7 @@ declare namespace mercurius {
   
   export interface MercuriusContext {
     app: FastifyInstance;
+    request: FastifyRequest;
     reply: FastifyReply;
     operationsCount?: number;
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

jonnydgreen (approved):
LGTM

mcollina:
CI does not seem happy

jonnydgreen:
Looking at the error in the CI, I think the issue is fixed in main (by this PR the other day: https://github.com/mercurius-js/mercurius/pull/1090)

@adtennant Maybe worth a quick rebase and re-run to be on the safe side?

mcollina (changes requested):
I just noticed this is missing a test. Could you add it?

PR #1090, opened 2024-03-18
Lock graphql-jit to v0.8.4

mcollina:
Ref https://github.com/zalando-incubator/graphql-jit/issues/241


mcollina:
cc @jonnydgreen locking graphql-jit to avoid breaking everybody

PR #1095, opened 2024-03-29
fix: handle if error.originalError.errors is not array

### LINKED ISSUE #1024: Occasional "error.originalError.errors.map is not a function" errors

dancastillo:
fixes #1024

* Current unit test handle coverage so no test was added
* Fix: Check that `error.orginialError.errors` is an `Array`

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -44,7 +44,7 @@ function defaultErrorFormatter (execution, ctx) {
     log.info({ err: error }, error.message)
 
     // it handles fastify errors MER_ERR_GQL_VALIDATION
-    if (error.originalError?.errors) {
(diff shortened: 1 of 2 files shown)


jonnydgreen:
Thanks for submitting! Could you add a test for this please? :)

dancastillo:
@jonnydgreen tests should be added in now

PR #1111, opened 2024-09-09
Update deps with potential breaking changes

mcollina:
The changes to the loaders tests are due to https://github.com/mcollina/single-user-cache/pull/30/.


simoneb (approved):
Not entirely sure what the linked PR fixes as there is no description, although I can see a switch from Object to Map, so I guess it's mostly a matter of uniqueness of keys. Anyhow, this LGTM

PR #1115, opened 2024-09-30
made mercurius's implicitly injected pubsub subscribe/publish methods type safe

adrtivv:
The rationale is that mercurius implicitly injects the `pubsub` field into the context received by the graphql resolvers. For making the context type-safe we need to define a custom context type to annotate its typescript type correctly for usage in the graphql resolvers and its in that context type where we can annotate the typescript type for `pubsub` object correctly for usage in the graphql resolvers by having the option of providing strict `event_name-payload` types to the `pubsub.subscribe` and `pubsub.publish` methods.

Usage is shown in the `example.ts` file present in the root directory of [this](https://stackblitz.com/~/github.com/mercurius-js/mercurius) stackblitz example with proper comments.

This implementation is inspired by [this](https://github.com/dotansimha/graphql-yoga/blob/fbe0bebad53597ab36407e3228ecbb9ca4a330c8/packages/subscription/src/create-pub-sub.ts#L36) implementation of pubsub by graphql-yoga but its made to work with mercurius's implementation of pubsub. 

Ideally mercurius should provide an explicit way of either correctly passing the types for these `event_name-payload` that are to be used in `pubsub.subscribe` and `pubsub.publish` functions or it should provide an explicit way of initializing the pubsub object instead of implicitly injecting it into the graphql context.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -23,9 +23,21 @@ import { Readable } from 'stream'
 type Mercurius = typeof mercurius
 
 declare namespace mercurius {
-  export interface PubSub {
-    subscribe<TResult = any>(topics: string | string[]): Promise<Readable & AsyncIterableIterator<TResult>>;
-    publish<TResult = any>(event: { topic: string; payload: TResult }, callback?: () => void): void;
+  export interface PubSubTopics {
+    [topic: string]: any;
+  }
+
+  export interface PubSub<TPubSubTopics extends PubSubTopics = PubSubTopics> {
+    subscribe<TTopic extends Extract<keyof TPubSubTopics, string>>(
+      topics: TTopic | TTopic[],
+    ): Promise<Readable & AsyncIterableIterator<TPubSubTopics[TTopic]>>;
+    publish<TTopic extends Extract<keyof TPubSubTopics, string>>(
+      event: {
+        topic: TTopic;
+        payload: TPubSubTopics[TTopic];
+      },
+      callback?: () => void,
+    ): void;
   }
 
   export interface MercuriusContext {
--- a/test/types/index.ts
+++ b/test/types/index.ts
@@ -398,7 +398,7 @@ app.graphql.pubsub.publish({
 })
 
 async () => {
-  const subscription = await app.graphql.pubsub.subscribe<{ newNotification: string }>('topic')
+  const subscription = await app.graphql.pubsub.subscribe('topic')
 
   subscription.on('data', (chunk) => {
     console.log(chunk)


adrtivv:
ahh i was fixing some formatting

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "mercurius",
-  "version": "14.1.0",
+  "version": "15.0.0",
```
mcollina, line 3: Please revert, we'll do this ourselves. Out of question, why is this semver-major?

mcollina (changes requested):
Can you add tests for the types? We use tsd.

package.json
adrtivv: I'm not sure, it happened automatically in this commit [6f18c6c](https://github.com/mercurius-js/mercurius/pull/1115/commits/6f18c6cb7553efb32e3e4f331b6ca5c64cd0fe09)

adrtivv:
> Can you add tests for the types? We use tsd.

Not free right now, will try later.

package.json
voxpelli: Looks like a failed rebase – this PR is based on https://github.com/mercurius-js/mercurius/commit/fe4c63d28037ace32f0e9da69a698e5ce48454b0 but seems to be mistakenly squashed with the commit that follows it in the main repository

index.d.ts
```diff
@@ -23,9 +23,20 @@ import { Readable } from 'stream'
...
-  export interface PubSub {
-    subscribe<TResult = any>(topics: string | string[]): Promise<Readable & AsyncIterableIterator<TResult>>;
-    publish<TResult = any>(event: { topic: string; payload: TResult }, callback?: () => void): void;
+  export type PubSubPublishArgsByKey = {
+    [key: string]: any;
+  };
+
+  export interface PubSub<
+    TPubSubPublishArgsByKey extends PubSubPublishArgsByKey,
+  > {
```
voxpelli, line 32: Unless you do this, this is a breaking change, as `PubSub` will require `TPubSubPublishArgsByKey` to be specified:

```suggestion
  export interface PubSub<
    TPubSubPublishArgsByKey extends PubSubPublishArgsByKey = PubSubPublishArgsByKey,
  > {
```

index.d.ts
```diff
@@ -23,9 +23,20 @@ import { Readable } from 'stream'
...
+      subscribe<TKey extends Extract<keyof TPubSubPublishArgsByKey, string>>(
+        topics: TKey | TKey[],
+      ): Promise<Readable & AsyncIterableIterator<TPubSubPublishArgsByKey[TKey]>>;
+      publish<TKey extends Extract<keyof TPubSubPublishArgsByKey, string>>(
+        event: { topic: TKey; payload: TPubSubPublishArgsByKey[TKey] },
+        callback?: () => void,
+      ): void;
```
voxpelli, line 39: The `subscribe` is [a wrapper](https://github.com/mercurius-js/mercurius/blob/a45fd77b7529cd763fba367b2f21eb0c583e5fd6/lib/subscriber.js#L21-L27) for the  `emitter` / `mqemitter` [method `on`](https://github.com/mcollina/mqemitter/blob/6b2e908fa6cd002d7b2227211512f7cc0e6f710d/types/mqemitter.d.ts#L16) and `publish` is [a wrapper](https://github.com/mercurius-js/mercurius/blob/a45fd77b7529cd763fba367b2f21eb0c583e5fd6/lib/subscriber.js#L34) for the `emitter` / `mqemitter` [method `emit`](https://github.com/mcollina/mqemitter/blob/6b2e908fa6cd002d7b2227211512f7cc0e6f710d/types/mqemitter.d.ts#L17): 

```ts
export type Message = Record<string, any> & { topic: string }
```

```ts
on(topic: string, listener: (message: Message, done: () => void) => void, callback?: () => void): this
emit(message: Message, callback?: (error?: Error) => void): void
```

There is some mismatch here, both in the old and in the new types. Eg: the publish callback should be: `callback?: (error?: Error) => void`

I think the `emitter` option of the constructor should also be specified to ensure a match between it and what's sent here: https://github.com/mercurius-js/mercurius/blob/a45fd77b7529cd763fba367b2f21eb0c583e5fd6/index.d.ts#L449

(Ideally the `publish()` would be reimplemented as `public(topic, payload, [callback])` and have some way to resolve to a promise rather than the callback, so that an `await pubsub.publish('topic1', 'value1`)` would be possible, but that's out of scope of this PR)

index.d.ts
```diff
@@ -23,9 +23,20 @@ import { Readable } from 'stream'
...
+  export type PubSubPublishArgsByKey = {
+    [key: string]: any;
+  };
+
+  export interface PubSub<
+    TPubSubPublishArgsByKey extends PubSubPublishArgsByKey,
+  > {
+      subscribe<TKey extends Extract<keyof TPubSubPublishArgsByKey, string>>(
+        topics: TKey | TKey[],
+      ): Promise<Readable & AsyncIterableIterator<TPubSubPublishArgsByKey[TKey]>>;
```
voxpelli, line 35: Does this work properly when `TKey` is something like `'key1' | 'key2'`?

index.d.ts
```diff
@@ -23,9 +23,20 @@ import { Readable } from 'stream'
 type Mercurius = typeof mercurius
 
 declare namespace mercurius {
-  export interface PubSub {
-    subscribe<TResult = any>(topics: string | string[]): Promise<Readable & AsyncIterableIterator<TResult>>;
-    publish<TResult = any>(event: { topic: string; payload: TResult }, callback?: () => void): void;
+  export type PubSubPublishArgsByKey = {
+    [key: string]: any;
+  };
```
voxpelli, line 28: This should probably mimic eg [`MercuriusContext`](https://github.com/mercurius-js/mercurius/blob/a45fd77b7529cd763fba367b2f21eb0c583e5fd6/index.d.ts#L31-L41) and be one can extend, rather than (or in addition) to being needed to specify like `PubSub<MyTopics>` – that way it will work automatically in all the places that use the `PubSub` type.

The name should also be in the style of the other interfaces.

```suggestion
  export interface PubSubTopics {
    [topic: string]: any;
  }
```

voxpelli (commented):
Some comments. I'm not a maintainer of this module though, just an active user.

index.d.ts
adrtivv: It would be a union type of the payloads corresponding to those keys.

index.d.ts
adrtivv: Yeah, the internal types are ambiguous and incorrectly mapped by the library authors but it doesn't affect this pull request.

index.d.ts
adrtivv: Thanks, fixed it.

index.d.ts
adrtivv: Fixed.

adrtivv:
@mcollina The test logic seems to be incorrect [here](https://github.com/mercurius-js/mercurius/blob/0c7a426085aaeac8f66528c71f645878e701f88f/test/types/index.ts#L401):

adrtivv:
> Can you add tests for the types? We use tsd.

Oh just saw this. I'll take a look.

mcollina:
@adrtivv why? What's the problem? (it might be)

adrtivv:
> @adrtivv why? What's the problem? (it might be)

The `publish` and `subscribe` methods are meant to take in generics of type `string` or custom types that extend type `string`. In the test an object `{ newNotification: string }` has been passed instead of a `string`.

mcollina:
Could you fix it?

adrtivv:
> Could you fix it?

Okay I'll try.

PR #1117, opened 2024-10-01
updated graphiql and react to the latest stable versions

adrtivv:
Updated the graphiql and react versions to how they're used in the official graphiql example([link](https://github.com/graphql/graphiql/blob/main/examples/graphiql-cdn/index.html) for reference) for the same.

Because of hard locking graphiql was locked to version 2.0.9 when the latest stable graphiql version is 3.7.1. The graphiql version 3.7.1 seems to work without issues with mercurius. React follows semantic versioning so no point in hard locking it to a patch version. 

Update: Restored hard locking, versions are hard locked to the latest stable versions

### DIFF
--- a/static/main.js
+++ b/static/main.js
@@ -157,7 +157,7 @@ function render () {
 
 function importDependencies () {
   const link = document.createElement('link')
-  link.href = 'https://unpkg.com/graphiql@2.0.9/graphiql.min.css'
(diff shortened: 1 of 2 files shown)


mcollina:
The reason for hard locking is that at some point they had a combination that didn't work. Hard locking prevents that.

mcollina (approved):
lgtm

PR #1121, opened 2024-10-23
feat: improve PersistedQueryProvider types

marcoreni:
While implementing a custom PersistedQueryProvider, we noticed that the types were not correct for the "getQueryFromHash" function. In fact, the function may return undefined if the query does not exist, but as of now the types do not allow that.

This PR fixes this (small) issue.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -629,7 +629,7 @@ declare namespace mercurius {
     /**
      *  Return the query for a given hash.
      */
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? We use tsd for types

mcollina (approved):
lgtm

PR #1124, opened 2024-12-09
feat: APQ should persist queries only when inside an APQ flow

### LINKED ISSUE #1123: Hash for persisted queries stored for standard POST requests
We noticed that Mercurius persists queries that are not stored even if the query is not part of a persisted query flow, and without equality checks in respect of what has been sent from the client:

https://github.com/mercurius-js/mercurius/blob/master/lib/routes.js#L280

Apollo has a different behavior, where queries are persisted only as part of an APQ flow and only if the hash sent by the client matches the hash calculated on the server.

This has a couple of benefits:
1- avoid storing non-needed queries on the cache, in scenarios where the server allows to use both Persisted Queries and standard POST queries at the same time
2- avoid storing the same query multiple times and surface implementation errors (eg. if the client computes a wrong hash, the server won't ever find the corresponding query and will always store the query on a different hash).

Since the reference for Persisted Queries on Mercurius Docs is Apollo, would it be possible to align the behaviors? I'm willing to work on a PR.

marcoreni:
Closes #1123 

While updating tests I noticed that the behavior I'm "fixing" was explicitely tested, so presumably desired. I haven't found any references of this on Apollo Server APQ flow, but now I'm unsure if this should be considered a Breaking Change or not.

### DIFF
--- a/docs/api/options.md
+++ b/docs/api/options.md
@@ -80,12 +80,14 @@
 - `onlyPersisted`: Boolean. Flag to control whether to allow graphql queries other than persisted. When `true`, it'll make the server reject any queries that are not present in the `persistedQueries` option above. It will also disable any ide available (graphiql). Requires `persistedQueries` to be set, and overrides `persistedQueryProvider`.
 - `persistedQueryProvider`
   - `isPersistedQuery: (request: object) => boolean`: Return true if a given request matches the desired persisted query format.
+  - `isPersistedQueryRetry: (request: object) => boolean`: Return true if a given request matches the desired persisted query retry format.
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

mcollina:
I'm a bit conflicted on the semversiness of this. What do you think?

jonnydgreen (approved):
LGTM

jonnydgreen:
> I'm a bit conflicted on the semversiness of this. What do you think?

Good question! Based on the test here, I think this is technically a breaking change? https://github.com/mercurius-js/mercurius/pull/1124/files#diff-5c532c8a8a16879f154d85383fa4f7e9d816ac20e808665e21657277968479f7R398

So maybe a major version? wdyt?

PR #1128, opened 2025-01-06
chore: update fastify to ^5.0.0, because it is currently in alpha version

Tony133:
Proposal:

- Update `fastify` version `5.0.0` in the `package.json`, currently the version of `5.0.0-alpha.4` is used


mcollina (approved):
lgtm

PR #1130, opened 2025-01-07
Subscriptions: support custom WebSockets path via subscription.wsPath option

tniezurawski:


tniezurawski:
Sorry! 😬 I would like to propose such an option but didn't want to open a PR against `mercurius` yet.

PR #1136, opened 2025-02-21
Handle backpressure in subscriptions

mcollina:


simone-sanfratello (approved):

simoneb (approved):
LGTM, simple enough. If I understand correctly, you're simply delegating the responsibility of managing the backpressure to the socket by using its callback function, which I guess is what tells you: go ahead, I'm not pushing back

mcollina:
Yes exactly. Luckily, `ws.send` has a callback (contrary to the spec one).

jonnydgreen:
Looks good and makes sense! Apologies, didn't see this in time!

PR #1138, opened 2025-03-13
fix: linting regressions

voxpelli:
In the canary tests for the `neostandard` lockfile maintenance it was discovered that with the latest dependencies two errors that had previously not been found in Mercurius was now found: https://github.com/neostandard/neostandard/pull/258

This fixes those two errors by exporting the two interfaces that's unused.


mcollina (approved):
lgtm

PR #1140, opened 2025-03-26
fix: removeListener called multiple times

### LINKED ISSUE #1139: The emitter's removeListener is called twice
When a client closes WS connection it calls the close method of the `SubscriptionContext` twice.

Triggered by:
https://github.com/mercurius-js/mercurius/blob/01a007dfab882923932c0276b1da69cac312a9aa/lib/subscription.js#L56-L58

And here:
https://github.com/mercurius-js/mercurius/blob/01a007dfab882923932c0276b1da69cac312a9aa/lib/subscription-connection.js#L331-L333

It produces two `removeListener` calls:
https://github.com/mercurius-js/mercurius/blob/01a007dfab882923932c0276b1da69cac312a9aa/lib/subscriber.js#L17-L19

If there are two connected clients, and one of them closes the connection, then it is possible that the topic message listeners will be removed for both clients if the emitter implementation does not include a flag to check whether the subscription for a given listener has been removed. For example, [mqemitter-redis](https://www.npmjs.com/package/mqemitter-redis) does not do this and sends an unsubscribe command, so the second client will stop receiving new messages from the topic.

Hokid:
Fixes #1139

### DIFF
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -44,6 +44,7 @@ class SubscriptionContext {
       objectMode: true,
       read: () => {}
     })
+    this.closed = false
   }
 
   subscribe (topics) {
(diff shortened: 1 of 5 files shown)


mcollina:
CI seems to be failing everywhere. Can you fix it?

Hokid:
> CI seems to be failing everywhere. Can you fix it?

Done

PR #1143, opened 2025-04-10
Fix loader batching logic in defineLoader + upgrade dependencies and GraphiQL

acoBOYZ:
# Fix internal `defineLoader` batching issue + dependency & GraphiQL updates

This pull request fixes a problem in **Mercurius' internal `defineLoader` function**, where GraphQL field loaders were being called **once per item**, instead of **once per field with batching**.

---

## ✅ What was fixed

The problem was caused by using complex objects (`{ obj, params }`) as loader keys, which prevented Mercurius from recognizing repeated calls as "the same".

We now use a `serialize()` function that:
- Uses `obj.id` and `params.lang` to create a fast and consistent batching key
- Falls back to `JSON.stringify({ obj, params })` if `id` is not available

> We use `id` because it's a **commonly used identifier** in GraphQL schemas, and it's **much faster** than serializing the entire object.

---

## ✅ Also included

- Upgraded all dependencies to their latest stable versions
- Updated GraphiQL static files (`main.js`, `config.js`) to the latest version

---

### Before (no batching):

- ~122ms average response time
- Loader was called **many times per query**

### After (with batching):

- ~37ms average response time
- Loader is called **only once**, with all keys batched
- ~70% performance improvement confirmed under load

---

## 🔒 Notes

- This fix is applied **internally**, users do not need to change how they define their own loaders.
- It's safe and backward-compatible.


static/sw.js
```diff
@@ -2,13 +2,13 @@
...
-    caches.open('graphiql-v3.7.1').then(function (cache) {
+    caches.open('graphiql-v3.8.3').then(function (cache) {
       return cache.addAll([
         './main.js',
-        'https://unpkg.com/graphiql@3.7.1/graphiql.css',
+        'https://unpkg.com/graphiql@3.8.3/graphiql.css',
         'https://unpkg.com/react@18.3.1/umd/react.production.min.js',
         'https://unpkg.com/react-dom@18.3.1/umd/react-dom.production.min.js',
-        'https://unpkg.com/graphiql@3.7.1/graphiql.min.js'
+        'https://unpkg.com/graphiql@3.8.3/graphiql.min.js'
```
mcollina, line 11: Can you send this to a different PR?

index.js
```diff
@@ -342,11 +342,30 @@ const mercurius = fp(async function (app, opts) {
...
+     */
+    function serialize ({ obj, params }) {
+      if (obj?.id !== undefined) {
+        const lang = params?.lang ?? null
+        return JSON.stringify([obj.id, lang])
       }
-      return query
+
+      return stringify({ obj, params })
     }
```
mcollina, line 369: I don't think this is generic enough to be included in the library. How about you add a a way to include a custom serializer for the loader instead?

static/main.js
```diff
@@ -167,7 +167,7 @@ function importDependencies () {
   return importer.urls([
     'https://unpkg.com/react@18.3.1/umd/react.production.min.js',
     'https://unpkg.com/react-dom@18.3.1/umd/react-dom.production.min.js',
-    'https://unpkg.com/graphiql@3.7.1/graphiql.min.js'
+    'https://unpkg.com/graphiql@3.8.3/graphiql.min.js'
```
mcollina, line 170: Can you send this to a different PR?

mcollina (changes requested):
Can you add some tests as well?

static/sw.js
acoBOYZ: Yes of course

index.js
acoBOYZ: You are right. I found a better way to handle it. Also i will add some tests for it. After that if you like it  can pr docs for that too. I will publish it soon

PR #1148, opened 2025-05-06
feat: custom pubsub with custom args on subscribe method

simone-sanfratello:
This PR wants to add the capability to have custom params on `subscribe` method using a `CustomPubSub` implementation. This is useful in case you need to get an argument from the subscription method


mcollina (approved):
lgtm

jonnydgreen (commented):
Functionally looks good to me! Is it worth updating the typings and type tests as well? wdyt? :)

jonnydgreen (approved):
LGTM!

PR #1150, opened 2025-05-18
chore: update dependencie typescript

Tony133:
Proposals:

- Update dependencie `typescript` to version `5.8.3` 🔥
- Renamed the `test2` command for the unit test in the `package.json` to `test:unit`


package.json
```diff
@@ -51,7 +51,7 @@
     "split2": "^4.0.0",
     "tap": "^21.0.0",
     "tsd": "^0.32.0",
-    "typescript": "^5.0.2",
+    "typescript": "^5.8.3",
```
voxpelli, line 54: I think this should be:

```suggestion
    "typescript": "~5.8.3",
```

As typescript has a tendency to release breaking changes in minor releases.

Thought can't remember if that's typically what is done in this repository.

Also:

Can you elaborate on whether there is anything particular that pushed you to make this PR? Generally I would say that basic maintenance of dev dependencies like this one is handled by maintainers unless there's a particular need for updating them outside the general schedule?

Tony133: > I think this should be:
> 
> As typescript has a tendency to release breaking changes in minor releases.
> 
> Thought can't remember if that's typically what is done in this repository.
> 
> Also:
> 
> Can you elaborate on whether there is anything particular that pushed you to make this PR? Generally I would say that basic maintenance of dev dependencies like this one is handled by maintainers unless there's a particular need for updating them outside the general schedule?

yes, I think it is better to use `~5.8.3`, no problem for me to change it, I had seen that it is way behind in version, so I had thought of making a contribution 😉. Seeing all the Fastify plugins packages are all set like this: `~5.8.3` I thought this adapter for fastify also had to be updated. 🙂

mcollina: I concur

mcollina (approved):
lgtm

jonnydgreen (approved):
LGTM!

PR #1158, opened 2025-06-02
test: migrated fix-790.test.js from tap to node:test

Tony133:
Proposals:

  - migrated `fix-790.test.js` from `tap` to `node:test` 🔥
  - added prefix `.test` 🔥


mcollina (approved):
lgtm

PR #1186, opened 2025-09-24
fix: subscription hook onSubscriptionEnd context

simone-sanfratello:
As titled, currently the context passed in the `onSubscriptionEnd` hook is not the same object context passed in the other hooks, because `onSubscriptionEnd` is handled as a connection event and it's detached from the resolver


mcollina (approved):
lgtm

PR #1202, opened 2025-12-30
check socket before awaiting subscription iterator

NeoPhi:
If the socket closes while we are awaiting subscribe() the iterator becomes orphaned since handleConnectionClose() has already run before we could add it to the map. This leads to a memory leak as the iterators are never cleaned up.


jonnydgreen (approved):
LGTM!

PR #1219, opened 2026-03-14
replace p-map with tiny-pmap

ofershap:
p-map v4 has no TypeScript types. The current version (v7) is ESM-only.

tiny-pmap has the same API, ships ESM + CJS with built-in types. Zero deps.

https://npmgraph.js.org/?q=tiny-pmap


mcollina:
I'm good with the change, but is this used anywhere? If not we can drop it?

PR #1251, opened 2026-07-13
fix: do not crash when a subscription message is delivered after the queue has ended

### LINKED ISSUE #1250: Server crashes with stream.push() after EOF when a subscription message arrives after the client unsubscribes (mqemitter-redis)
## Description

Under GraphQL subscription churn (clients that subscribe and unsubscribe in quick succession), the whole Node process crashes:

```
NodeError: stream.push() after EOF
    at MQEmitter._do (node_modules/mqemitter/mqemitter.js:133:8)
    at MQEmitterRedis.emit (node_modules/mqemitter/mqemitter.js:115:10)
    at handler (node_modules/mqemitter-redis/mqemitter-redis.js:56:12)
Emitted 'error' event on Readable instance ...
```

### Root cause

With an emitter that delivers asynchronously (we hit this with `mqemitter-redis`, where messages arrive over the Redis pubsub connection), a message can already be in flight when the client unsubscribes:

1. Client disconnects / unsubscribes → `SubscriptionContext.close()` runs → `queue.push(null)` ends the Readable.
2. A message that was already read off the wire is still delivered to the pubsub `listener` in `lib/subscriber.js`, which does `queue.push(value.payload)` unconditionally.
3. Push after EOF makes the Readable emit `'error'` (`ERR_STREAM_PUSH_AFTER_EOF`). Nothing listens for `'error'` on the subscription queue, so Node treats it as an uncaught exception and kills the process.

So one user navigating away at the wrong moment can take down the whole server. In-process `mqemitter` is mostly immune because delivery ordering makes the window very hard to hit, which is presumably why this hasn't surfaced before; any networked emitter hits it under load. We were seeing our API die reliably during E2E runs that open/close live queries back to back.

Same failure class as #430 (unhandled error on a subscription-owned resource escalating to process death), but a different vector.

### Environment

- mercurius 16.8.0 (code is unchanged in 16.9.0 and current master `lib/subscriber.js`)
- mqemitter-redis 7.2.0, mqemitter 7.1.0
- fastify 5.x, Node 22

### Reproduction

Deterministic unit repro (models async delivery: listener matched before `removeListener`, invoked after `close()`):

```js
const { PubSub, SubscriptionContext } = require('mercurius/lib/subscriber')

const listeners = new Map()
const emitter = {
  on (topic, listener, done) { listeners.set(listener, topic); done() },
  removeListener (topic, listener) { listeners.delete(listener) },
  emit (event, cb) {
    const matched = [...listeners.keys()]
    setImmediate(() => { for (const l of matched) l(event, () => {}); cb() })
  }
}

const pubsub = new PubSub(emitter)
const sc = new SubscriptionContext({ pubsub })
await sc.subscribe('TOPIC')
emitter.emit({ topic: 'TOPIC', payload: 1 }, () => {})  // message in flight...
sc.close()                                               // ...client goes away
// next tick: ERR_STREAM_PUSH_AFTER_EOF, process dies
```

### Proposed fix

Two guards in `lib/subscriber.js` (PR incoming):

1. In the pubsub `listener`, drop the message if the queue has already ended or been destroyed — the subscriber is gone.
2. Defence in depth: attach an `'error'` handler to the subscription queue that logs instead of crashing the process.

We've been running this as a patch in production and it has held up under the same E2E churn that previously killed the API on every run.

kyleshepherd:
Fixes #1250

An emitter with asynchronous delivery (e.g. `mqemitter-redis`) can deliver a message that was already in flight when `SubscriptionContext.close()` ended the queue with `push(null)`. Pushing into the ended Readable emits an unhandled `'error'` event (`ERR_STREAM_PUSH_AFTER_EOF`), which brings down the whole server. See the linked issue for the full analysis and stack trace.

Two guards:

- the pubsub `listener` drops a message whose queue has already ended or been destroyed — the subscriber is gone, there is nothing to deliver to
- `SubscriptionContext` attaches an `'error'` handler to its queue as defence in depth, logging instead of crashing

Tests cover all three paths: in-flight message after `close()` (fails with `ERR_STREAM_PUSH_AFTER_EOF` on master), message to a destroyed queue, and the queue `'error'` handler. `lib/subscriber.js` is at 100% line/branch/function coverage; the remaining global coverage shortfall (`index.js`, `adaptive-jit.js`) is pre-existing on master.

`npm test` (lint + unit + typescript) passes. No hot-path cost worth noting: the guard is two property reads per delivery.

We have been running this fix as a production patch against 16.8.0 — the subscription churn that previously killed our API on every E2E run no longer does.

### DIFF
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -9,14 +9,17 @@ class PubSub {
 
   subscribe (topic, queue) {
     return new Promise((resolve, reject) => {
+      // A message can be delivered after the subscription has closed:
+      // emitters with asynchronous delivery (e.g. mqemitter-redis) may have
+      // a message in flight when close() runs, and deliver it after
+      // SubscriptionContext.close() has ended the queue with push(null).
+      // Pushing after EOF emits an unhandled 'error' on the Readable and
+      // crashes the process. The subscriber is gone — drop the message.
+      // (`readableEnded` alone cannot guard this: it only becomes true once
+      // 'end' has been emitted, which may be after the late delivery.)
+      let unsubscribed = false
       function listener (value, cb) {
-        // A message can be delivered after the subscription queue has ended:
-        // emitters with asynchronous delivery (e.g. mqemitter-redis) may have
-        // a message in flight when SubscriptionContext.close() pushes null.
-        // Pushing after EOF emits an unhandled 'error' on the Readable and
-        // crashes the process. The subscriber is gone — drop the message.
-        const state = queue._readableState
-        if (queue.destroyed || (state && state.ended)) {
+        if (unsubscribed || queue.destroyed || queue.readableEnded) {
           return cb()
         }
         queue.push(value.payload)
@@ -24,6 +27,7 @@ class PubSub {
       }
 
       const close = () => {
+        unsubscribed = true
         this.emitter.removeListener(topic, listener)
       }
 
--- a/lib/subscriber.js
+++ b/lib/subscriber.js
@@ -10,6 +10,15 @@ class PubSub {
   subscribe (topic, queue) {
     return new Promise((resolve, reject) => {
       function listener (value, cb) {
+        // A message can be delivered after the subscription queue has ended:
+        // emitters with asynchronous delivery (e.g. mqemitter-redis) may have
+        // a message in flight when SubscriptionContext.close() pushes null.
+        // Pushing after EOF emits an unhandled 'error' on the Readable and
+        // crashes the process. The subscriber is gone — drop the message.
+        const state = queue._readableState
+        if (queue.destroyed || (state && state.ended)) {
+          return cb()
+        }
         queue.push(value.payload)
         cb()
       }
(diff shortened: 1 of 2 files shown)


lib/subscriber.js
```diff
@@ -10,6 +10,15 @@ class PubSub {
   subscribe (topic, queue) {
     return new Promise((resolve, reject) => {
       function listener (value, cb) {
+        // A message can be delivered after the subscription queue has ended:
+        // emitters with asynchronous delivery (e.g. mqemitter-redis) may have
+        // a message in flight when SubscriptionContext.close() pushes null.
+        // Pushing after EOF emits an unhandled 'error' on the Readable and
+        // crashes the process. The subscriber is gone — drop the message.
+        const state = queue._readableState
```
mcollina: I don't think there is a need to use `_readableState`. We have https://nodejs.org/docs/latest/api/stream.html#readablereadableended and https://nodejs.org/docs/latest/api/stream.html#readabledestroyed for this.

mcollina (changes requested):

lib/subscriber.js
kyleshepherd: Reworked to drop `_readableState` entirely. One wrinkle with `readableEnded`: it only flips after `'end'` has been emitted, and in the reported race (mqemitter-redis delivering an in-flight message just after close) the queue has been ended with `push(null)` but not yet drained, so it's still `false` at delivery time and the push would still trigger `ERR_STREAM_PUSH_AFTER_EOF`. Since there's no public getter for that pre-`'end'` state, the listener now tracks an `unsubscribed` flag set by the existing close closure that removes it (invoked by `SubscriptionContext.close()` before it pushes null), which guards the actual condition without touching stream internals. `queue.destroyed` and `queue.readableEnded` remain as the public-API checks for the destroyed/drained cases.

mcollina (approved):
lgtm

PR #1255, opened 2026-07-16
fix: make sendError payload format respect WebSocket subprotocol

### LINKED ISSUE #1132: Incorrect error message format for the graphql-ws protocol
Hi, I've found a gap in support for the legacy graphql-ws protocol. The [spec](https://github.com/apollographql/subscriptions-transport-ws/blob/51270cc7dbaf09c7b9aa67368f1de58148c7d334/PROTOCOL.md#gql_error) indicates that for error messages the payload should be a single "Error" object, but currently Mercurius only adheres to the [graphql-transport-ws protocol error format](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#error) by always returning an array of GraphQL errors.

I realize that the graphql-ws protocol is considered deprecated, but given the docs state both are supported I think this is worth addressing. Thanks!

tsushanth:
## Problem

`sendError` in `lib/subscription-connection.js` unconditionally wraps the error in an array:

```js
this.sendMessage(this.protocolMessageTypes.GQL_ERROR, id, [convertedError])
```

This is correct for the `graphql-transport-ws` protocol (the newer spec), but breaks clients using the legacy `subscriptions-transport-ws` protocol. In mercurius/Fastify, that legacy protocol is identified by `socket.protocol === 'graphql-ws'` — and it expects a **single** error object as the payload, not an array.

Sending an array to a `subscriptions-transport-ws` client causes the client to fail to parse the error, surfacing as an unhandled error or silent failure rather than a proper GraphQL error.

## Fix

The active subprotocol is already available on `this.socket.protocol` and is used elsewhere in the same class for protocol-branching logic. The fix is a one-line conditional:

```js
const payload = this.socket.protocol === 'graphql-ws' ? convertedError : [convertedError]
```

- `'graphql-ws'` → legacy `subscriptions-transport-ws` → single error object
- anything else (e.g. `'graphql-transport-ws'`) → newer spec → array

Closes #1132

### DIFF
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -445,8 +445,11 @@ module.exports = class SubscriptionConnection {
 
   sendError (err, id) {
     const convertedError = toGraphQLError(err)
-    // Graphql over websocket spec requires that errors are wrapped in an array
-    this.sendMessage(this.protocolMessageTypes.GQL_ERROR, id, [convertedError])
+    // graphql-transport-ws requires errors wrapped in an array; the legacy
+    // subscriptions-transport-ws protocol (identified as 'graphql-ws') expects
+    // a single error object instead.
+    const payload = this.socket.protocol === 'graphql-ws' ? convertedError : [convertedError]
+    this.sendMessage(this.protocolMessageTypes.GQL_ERROR, id, payload)
   }
 
   async handleConnectionInitExtension (extension) {
--- a/lib/subscription-connection.js
+++ b/lib/subscription-connection.js
@@ -8,7 +8,7 @@ const sJSON = require('secure-json-parse')
 const { MER_ERR_GQL_VALIDATION, MER_ERR_GQL_SUBSCRIPTION_FORBIDDEN, MER_ERR_GQL_SUBSCRIPTION_UNKNOWN_EXTENSION, MER_ERR_GQL_SUBSCRIPTION_INVALID_OPERATION } = require('./errors')
 const { preSubscriptionParsingHandler, onSubscriptionResolutionHandler, preSubscriptionExecutionHandler, onSubscriptionEndHandler, onSubscriptionConnectionCloseHandler, onSubscriptionConnectionErrorHandler } = require('./handlers')
 const { kSubscriptionFactory, kLoaders } = require('./symbols')
-const { getProtocolByName } = require('./subscription-protocol')
+const { getProtocolByName, GRAPHQL_WS_PROTOCOL_SIGNALS } = require('./subscription-protocol')
 const { toGraphQLError } = require('./errors')
 const queryDepth = require('./queryDepth')
 
@@ -446,9 +446,10 @@ module.exports = class SubscriptionConnection {
   sendError (err, id) {
     const convertedError = toGraphQLError(err)
     // graphql-transport-ws requires errors wrapped in an array; the legacy
-    // subscriptions-transport-ws protocol (identified as 'graphql-ws') expects
-    // a single error object instead.
-    const payload = this.socket.protocol === 'graphql-ws' ? convertedError : [convertedError]
+    // subscriptions-transport-ws protocol expects a single error object.
+    // Use the resolved protocolMessageTypes (not socket.protocol) so that the
+    // default subprotocol applies correctly when the client omits the header.
+    const payload = this.protocolMessageTypes === GRAPHQL_WS_PROTOCOL_SIGNALS ? convertedError : [convertedError]
     this.sendMessage(this.protocolMessageTypes.GQL_ERROR, id, payload)
   }
 
--- a/test/query-depth.test.js
+++ b/test/query-depth.test.js
@@ -516,5 +516,6 @@ test('queryDepth - enforce depth limit for subscriptions over websocket', async 
   const errorMessage = await waitForMessageType('error')
 
   t.assert.strictEqual(errorMessage.id, '1')
-  t.assert.match(errorMessage.payload[0].message, /Graphql validation error/)
+  // graphql-ws (subscriptions-transport-ws) sends a single error object, not an array
+  t.assert.match(errorMessage.payload.message, /Graphql validation error/)
 })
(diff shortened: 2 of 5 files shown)


mcollina (changes requested):

tsushanth:
Added two unit tests in `test/subscription-connection.test.js` that verify `sendError` payload format per each protocol:

- **graphql-ws** (subscriptions-transport-ws): expects a plain error object
- **graphql-transport-ws**: expects errors wrapped in an array

Running the full test suite also surfaced three existing tests that were asserting the old (broken) array format for `graphql-ws` connections — those expectations have been corrected to match the spec as well (`subscription.test.js`, `subscription-hooks.test.js`, `query-depth.test.js`).

tsushanth:
Hi @mcollina — the unit tests you asked for are in [`test/subscription-connection.test.js`](https://github.com/tsushanth/mercurius/blob/fix/sendError-protocol-aware-payload/test/subscription-connection.test.js), added in the latest commit. Two tests cover `sendError` payload format per protocol:

- `sendError sends a plain object for graphql-ws protocol` (subscriptions-transport-ws)
- `sendError wraps error in array for graphql-transport-ws protocol`

Also updated three existing tests in `subscription.test.js`, `subscription-hooks.test.js`, and `query-depth.test.js` that were asserting the old (incorrect) array-wrapped format for graphql-ws, so the full test suite passes. Let me know if you'd like anything adjusted!

mcollina (approved):
lgtm

leorossi:
The spec direction here looks right to me — subscriptions-transport-ws does want a single error object, and graphql-transport-ws wants the array. But I think the protocol check needs to use the *resolved* protocol rather than the raw one, or the change lands inconsistently.

`socket.protocol` is the subprotocol the client actually negotiated. Mercurius elsewhere resolves that against `wsDefaultSubprotocol`:

- `subscription-connection.js:51` — `this.protocolMessageTypes = getProtocolByName(socket.protocol, this.wsDefaultSubprotocol)`
- `subscription-protocol.js:39` — `isValidClientProtocol()` explicitly accepts an empty client protocol when a default is configured

So on a server configured with `wsDefaultSubprotocol: 'graphql-ws'`, a client that omits `Sec-WebSocket-Protocol` is handled as legacy graphql-ws for every message type — `start`, `data`, `ka` — but with this patch still receives the *array* error shape, while a client that sends the header receives the object. Same server, same protocol semantics, two payload shapes.

I applied the diff to 16.9.0 and drove two clients against one server (`wsDefaultSubprotocol: 'graphql-ws'`), triggering an error via an unknown message type:

- client sending `Sec-WebSocket-Protocol: graphql-ws` → `payload: { ... }`
- client sending no subprotocol → `payload: [{ ... }]`

Minimal fix, since the signals object is already exported:

```js
const { GRAPHQL_WS_PROTOCOL_SIGNALS } = require('./subscription-protocol')

sendError (err, id) {
  const convertedError = toGraphQLError(err)
  const payload = this.protocolMessageTypes === GRAPHQL_WS_PROTOCOL_SIGNALS
    ? convertedError
    : [convertedError]
  this.sendMessage(this.protocolMessageTypes.GQL_ERROR, id, payload)
}
```

That reuses the resolution already done in the constructor, so it stays correct however `wsDefaultSubprotocol` is configured. Might be worth a test case for the header-less client too, since the existing ones all set `protocol` explicitly.

Separately on versioning: this changes an observable wire format for clients that have been parsing `payload[0]` against every released version so far, so it reads as breaking to me. Would you consider landing it behind an option in 16.x with the current behaviour as the default, and flipping the default in 17? That would let downstreams move client and server independently rather than atomically.

leorossi:
@tsushanth this is a test that shows the edge case

```
test('sendError uses one payload shape regardless of how the protocol was resolved', async (t) => {
  const app = fastify()
  t.after(() => app.close())

  app.register(mercurius, {
    schema: `
      type Query {
        _placeholder: String
      }

      type Subscription {
        onMessage: String!
      }
    `,
    resolvers: {
      Query: {},
      Subscription: {
        onMessage: {
          async * subscribe () {
            yield { onMessage: 'hello' }
          }
        }
      }
    },
    subscription: {
      wsDefaultSubprotocol: GRAPHQL_WS
    }
  })

  await app.listen({ port: 0 })
  const url = 'ws://localhost:' + (app.server.address()).port + '/graphql'

  // An unrecognised message type reaches the default branch of handleMessage(),
  // which calls sendError() directly. Every other sendError call site behaves
  // the same way.
  async function errorPayload (subprotocol) {
    const ws = subprotocol ? new WebSocket(url, subprotocol) : new WebSocket(url)

    ws.on('error', (error) => {
      t.assert.fail('must not error ' + error)
    })

    await once(ws, 'open')
    ws.send(JSON.stringify({ type: 'connection_init' }))
    ws.send(JSON.stringify({ type: 'not-a-real-message-type', id: 1 }))

    let payload
    for await (const [message] of on(ws, 'message')) {
      const data = JSON.parse(message.toString())
      if (data.type === 'error') {
        payload = data.payload
        break
      }
    }

    ws.close()
    return payload
  }

  const negotiated = await errorPayload(GRAPHQL_WS)
  const defaulted = await errorPayload()

  t.assert.strictEqual(
    Array.isArray(negotiated),
    Array.isArray(defaulted),
    'a client that omits the subprotocol is resolvultSubprotocol, so it must receive the same error payload shape as one that negotiates it explicitly'
  )
})
```

Add it to`subscription-connection.test.js` and that would be green both on `main` and with the proposed fix

tsushanth:
Fixed in the latest push — switched from `this.socket.protocol === 'graphql-ws'` to `this.protocolMessageTypes === GRAPHQL_WS_PROTOCOL_SIGNALS` so the resolved default subprotocol is used consistently.

Also added your exact test case (the one you posted above) to `test/subscription-connection.test.js` — it was failing on `main` and passes with the fix.

mcollina (approved):
lgtm

tsushanth:
Friendly ping — this has been approved by @mcollina for a week now and CI is green. Happy to rebase or make any changes if needed before merge.

mcollina:
This is semver-major. I'll land it whenever we cut a new major.

PR #1256, opened 2026-07-23
fix: graphql@17.0.0 compatibility

Tony133:
Proposal:

- Fixes `mercurius` compatibility with `graphql@17.0.0`, while keeping support for `graphql@16` at the same time so users who can't upgrade yet aren't forced to. Without the graphql@17 fixes, 73/376 tests failed after bumping the dependency (see here for preview: https://github.com/mercurius-js/mercurius/actions/runs/29320645825/job/87044889698#step:5:14).

Changes:
- **`index.js`**: `buildExecutionContext` was removed from `graphql/execution/execute` in graphql 17 → feature-detected against its drop-in successor `validateExecutionArgs` (same signature/behavior), falling back to `buildExecutionContext` on graphql@16.
- **`lib/errors.js`**: `GraphQLError`'s positional constructor was fully removed in graphql 17 (it's supported since graphql@16.3, so this works on both). `toGraphQLError()` used the old positional args, which silently dropped `originalError` — breaking status code mapping (validation/CSRF/persisted-query errors returned 200 instead of 400/405). Switched to the new options-object signature.
- **`index.js`**: graphql 17 tags AST `loc.source` with a non-cloneable `Symbol`, so `structuredClone()` crashed on pre-parsed documents (e.g. `graphql-tag`). Replaced with `rfdc` (already in the fastify dep tree) for cloning — faster than `structuredClone`/JSON, and it naturally skips the Symbol-keyed brand without needing to strip `loc`.
- **`package.json`**: widened `peerDependencies.graphql` to `^16.8.0 || ^17.0.0`; added `rfdc` as an explicit dependency.
- **`test/directives.test.js`**: the scalar-coercion error message wording differs between graphql-js 16 and 17, so the expected string is now resolved at runtime based on the installed `graphql` version (via the `version` export — `require('graphql/package.json')` no longer works on @17, its `exports` map blocks arbitrary subpaths).
- **`test/app-decorator.test.js`**: fixed a wrong assertion — mercurius wraps validation `GraphQLError`s in `MER_ERR_GQL_VALIDATION.errors[]`, the test was checking the wrong object (`e` instead of `e.errors[0]`).
- **`.github/workflows/ci.yml`**: added a `graphql-version: [16, 17]` matrix dimension, with a step that installs the matrix's graphql version after the regular install (`--no-save`), so both majors are covered in CI. Add node 26 to test matrix.

Note:
- This PR should resolve https://github.com/mercurius-js/mercurius/pull/1249

### DIFF
--- a/index.js
+++ b/index.js
@@ -20,7 +20,20 @@ const {
   specifiedRules,
   execute
 } = require('graphql')
-const { buildExecutionContext } = require('graphql/execution/execute')
+const { validateExecutionArgs } = require('graphql/execution/execute')
+
+// graphql@17 tags AST `loc.source` (Source) instances with a non-cloneable
+// Symbol, so `structuredClone` throws on documents built with e.g. `graphql-tag`.
+// Location info is not required for execution, so on failure we strip `loc`
+// and retry via a plain JSON clone instead of failing the whole request.
+function cloneDocument (source) {
+  try {
+    return structuredClone(source)
+  } catch (err) {
+    return JSON.parse(JSON.stringify(source, (key, value) => key === 'loc' ? undefined : value))
+  }
+}
+
 const queryDepth = require('./lib/queryDepth')
 const mq = require('mqemitter')
 const { PubSub, withFilter } = require('./lib/subscriber')
@@ -523,7 +536,7 @@ const mercurius = fp(async function (app, opts) {
       try {
         document = typeof source === 'string'
           ? parse(source, gqlParseOpts)
-          : structuredClone(source)
+          : cloneDocument(source)
       } catch (syntaxError) {
         try {
           // Do not try to JSON.parse maxToken exceeded validation errors
@@ -599,7 +612,7 @@ const mercurius = fp(async function (app, opts) {
     const shouldCompileJit = !adaptiveJit && cached && cached.count++ === minJit
     // Validate variables
     if (variables !== undefined && !shouldCompileJit) {
-      const executionContext = buildExecutionContext({
+      const executionContext = validateExecutionArgs({
         schema: fastifyGraphQl.schema,
         document,
         rootValue: root,
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -21,12 +21,14 @@ function toGraphQLError (err) {
 
   const gqlError = new GraphQLError(
     err.message,
-    err.nodes,
-    err.source,
-    err.positions,
-    err.path,
-    err,
-    err.extensions
+    {
+      nodes: err.nodes,
+      source: err.source,
+      positions: err.positions,
+      path: err.path,
+      originalError: err,
+      extensions: err.extensions
+    }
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -20,7 +20,20 @@ const {
...
+
+// graphql@17 tags AST `loc.source` (Source) instances with a non-cloneable
+// Symbol, so `structuredClone` throws on documents built with e.g. `graphql-tag`.
+// Location info is not required for execution, so on failure we strip `loc`
+// and retry via a plain JSON clone instead of failing the whole request.
+function cloneDocument (source) {
+  try {
+    return structuredClone(source)
+  } catch (err) {
+    return JSON.parse(JSON.stringify(source, (key, value) => key === 'loc' ? undefined : value))
```
mcollina, line 33: This is really slow. Can we use one of the fast libraries for cloning in the fastify dep tree?

mcollina (changes requested):
I would _prefer_ if we could support both 16 and 17 at the same time, as a lot of people would not be able to upgrade graphql in time. This would require to fiddle with CI so that we run tests for both versions.

Tony133:
> I would _prefer_ if we could support both 16 and 17 at the same time, as a lot of people would not be able to upgrade graphql in time. This would require to fiddle with CI so that we run tests for both versions.

yes I agree, I just wanted to know what you thought first 😉

index.js
Tony133: > This is really slow. Can we use one of the fast libraries for cloning in the fastify dep tree?

yes!, mmmh maybe we can use this library: https://www.npmjs.com/package/rfdc ?

Tony133:
I updated the files so as to have compatibility for graphql 16 and 17 and I also updated the CI file so it separates the tests for graphql 16 and 17, plus I added nodejs v26 🔥

mcollina (changes requested):
index.js has a lot of pure formatting changes (prettier?). Please revert them

mcollina:
keeping as semver-major as safety measure.

Tony133:
> index.js has a lot of pure formatting changes (prettier?). Please revert them

I think it's my editor that formats everything to a maximum of 80 columns 😄

mcollina:
Please revert, this is very hard to review

Tony133:
> Please revert, this is very hard to review

Ok, I think I've fixed it, see commit here: 7a38023. Sorry about the issues with the code indentation 🙏🏻

mcollina (approved):
lgtm

mcollina:
there are conflicts

Tony133:
> there are conflicts

I don't see any conflicts to resolve 😄, the CI is all green 🙌 ✅

Tony133:
I accidentally closed the PR 😅, sorry

Tony133:
ok,  the CI is all green 🙌 ✅

PR #1268, opened 2026-09-06
fix(subscriptions): send single error object for legacy graphql-ws GQL_ERROR

### LINKED ISSUE #1132: Incorrect error message format for the graphql-ws protocol

kilisamemarisaaa:
Fixes #1132

## Problem

The [legacy graphql-ws protocol](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_error) (`subscriptions-transport-ws`) defines the `GQL_ERROR` payload as a **single Error object**:

> - `payload` (_Error!_) - The payload must be an Error object.

while the newer [graphql-transport-ws](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#error) protocol wraps errors in an **array**. Mercurius always sent the array shape on both protocols, so legacy clients following the spec (e.g. `subscriptions-transport-ws` client) could not read the error payload from `GQL_ERROR` messages.

## Fix

`SubscriptionConnection.sendError` now discriminates on the connection's protocol signals and only wraps the converted error in an array for `graphql-transport-ws`; the legacy `graphql-ws` protocol receives the single error object:

```js
const payload = this.protocolMessageTypes === GRAPHQL_WS_PROTOCOL_SIGNALS
  ? convertedError
  : [convertedError]
this.sendMessage(this.protocolMessageTypes.GQL_ERROR, id, payload)
```

## Tests

- New `SubscriptionConnection` test: legacy `graphql-ws` protocol receives `payload` as a single error object (red first: actual `[...]` vs expected `{...}`), `graphql-transport-ws` array-shape coverage unchanged.
- Updated 4 existing assertions that had pinned the non-compliant array shape **under the legacy protocol**: `test/subscription.test.js` (execution error), `test/subscription-hooks.test.js` (preSubscriptionParsing / preSubscriptionExecution), `test/query-depth.test.js` (depth-limit error over legacy ws).
- Full suite: **377/377 pass**; `npm run typescript` 77/77 assertions pass; eslint clean on all touched files.

## Backward compatibility

- For **graphql-transport-ws** clients: wire shape is unchanged (errors remain an array).
- For **legacy graphql-ws** clients using the official `subscriptions-transport-ws` client: the shape now matches the spec it implements (`payload` is a single error object), which is what that client's `onerror` expects.
- A client that had **worked around** mercurius's array shape under the legacy protocol would see a breaking change; that shape contradicts the legacy spec (PROTOCOL.md#gql_error), which this issue documents. I chose not to add an opt-in flag since the spec defines a single correct shape — happy to add one if the maintainers prefer a softer rollout.


simone-sanfratello:
Thanks for the contribution @kilisamemarisaaa !

However, looking at the updates on tests, this fix does not seem to be back compatible. Can you improve?

simone-sanfratello:
Also, duplicated of https://github.com/mercurius-js/mercurius/pull/1255 - I'm closing the PR

